PR #4, opened 2023-01-19
Fix git-lfs track command in Performance

albertvillanova:
There is a mistake in the `git-lfs track` command in the Performance: images are not properly tracked by Git LFS, thus resulting in a git remote error while pushing.

This PR fixes the `git-lfs track` command, so that all images are properly tracked by Git LFS.

### DIFF
--- a/Performance.md
+++ b/Performance.md
@@ -42,7 +42,7 @@ Executed in   38.87 mins
 Compare this to a system like [git lfs](https://git-lfs.github.com/) on the same dataset
 
 ```
-git lfs track *.jpg
+git lfs track "*.jpg"
 git add images # ~136 sec
 git commit -m "adding images" # ~44 sec
 ```


Performance.md
```diff
@@ -42,7 +42,7 @@ Executed in   38.87 mins
 Compare this to a system like [git lfs](https://git-lfs.github.com/) on the same dataset
 
 ```
-git lfs track images # ~17 sec
+git lfs track *.jpg
```
julien-c, line 45:
```suggestion
git lfs track "*.jpg"
```

albertvillanova: Thanks @julien-c

gschoeni:
@albertvillanova we were wondering if something was off, thank you! Let me try again and update the performance numbers.

gschoeni:
Verfied this works with `git lfs ls-files` 👍

albertvillanova:
Great!

@gschoeni please note that the fix works because all the files inside the "images" have ".jpg" extension. Alternatively, to track all the files inside the "images" directory, independently of their extension, you could have used:
```
git lfs track "images/*"
```

PR #12, opened 2023-05-15
Add remote functionality to python/oxen

benartuso:


oxen/src/py_remote_repo.rs
```diff
@@ -95,18 +108,118 @@ impl PyRemoteRepo {
...
+                &self.revision,
+                &user_id,
+                &path,
+                liboxen::constants::DEFAULT_PAGE_NUM,
+                liboxen::constants::DEFAULT_PAGE_SIZE,
+            )
+            .await
+        })?;
+
+        let mut status = StagedData::empty();
```
gschoeni, line 202: It feels like we should do this in a From trait on the PyStagedData struct

oxen/src/py_remote_repo.rs
```diff
@@ -123,4 +236,49 @@ impl PyRemoteRepo {
...
+        match branch {
+            Ok(branch) => Ok(PyBranch {
+                name: branch.name,
+                commit_id: branch.commit_id,
+            }),
+            _ => Err(PyValueError::new_err("Could not get or create branch")),
+        }
+    }
+    fn checkout(&mut self, revision: String) -> PyResult<()> {
+        // TODO: Flatten
```
gschoeni, line 263: Agreed, feels like we can avoid nesting here

gschoeni (changes requested):
Minor aesthetic changes, but otherwise excited to get this in!

oxen/src/py_remote_repo.rs
benartuso: Updated, lmk if you had something different in mind for the commit case!

oxen/src/py_remote_repo.rs
benartuso: done!

gschoeni (approved):

oxen/src/py_remote_repo.rs
gschoeni: beautiful

PR #22, opened 2023-06-14
update for generative examples: compatibility w/ pre 3.10 python + add remote restore df functionality

benartuso:
Was breaking on Lambda (3.8.x) bc the `|` operator is a 3.10+ thing. 

Also need remote df restore functionality to avoid label drift if add imagery is unsuccessful but add data is successful.


gschoeni (approved):
LGTM

PR #28, opened 2023-07-29
Fix documentation typo: "it's" (it is) to "its" (possessive)

welpo:
This PR fixes a few instances in the documentation where the contraction "it's" (it is) has been incorrectly used in place of the possessive pronoun "its".

The code that creates the commits in https://www.oxen.ai/ox/FlyingOxen would also need to be updated from "[user] gave an Ox **it's** wings" to "[user] gave an Ox **its** wings".

I haven't discussed these changes, but I saw a small opportunity to improve this promising project and decided to take it. If this isn't the correct process for submitting changes like this, please let me know.


gschoeni (approved):
Thank you so much!

PR #58, opened 2025-01-30
add list workspaces command

EloyMartinez:


oxen/pyproject.toml
```diff
@@ -4,6 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "oxenai"
+version = "0.1.0"
```
jcelliott, line 7: We're currently setting the version when the pypi release gets pushed. I'm not sure we want to set this here.

EloyMartinez: good catch, my bad, maturin was asking for a version on the pyproject file and i added it to not have to debug further

jcelliott (approved):

PR #97, opened 2025-08-01
Put should be able to overwrite a head revision5

paul-hammant:
PUT and DELETE now


Eric-Laurence:
There's no feasible way to merge this, it would have to be reimplemented.

paul-hammant:
I'll start over

Eric-Laurence:
Thanks for understanding. Apologies that this was never reviewed when it was first made! Smaller PRs might help us merge these in quicker so they don't fall behind the main branch.

PR #122, opened 2025-08-12
OXB214 OXB216 Refactor workspace non tabular file download 

Lilianxr:
1. Added `get_version_stream()` and `get_version_chunk_stream()` in the version store interface.
2. Refactored the endpoint `GET /workspaces/{workspace_id}/files/{path:.*}` to use version store
3. Also modified the versions download endpoint from 
`GET /versions/{version_id}`, `GET /versions/{version_id}/chunks/download`
to 
`GET /versions/{resource:.*}`, `GET /versions/chunks/{resource:.*}/download`
Because the client is not able to know the file version_id when making a download request. 
4. Modified the versions download controller function with similar logic as `GET /file/{resource:.*}`

## Summary by CodeRabbit

* **New Features**
  * Storage-backed streaming reads and on-demand server-side image resizing via width/height query.

* **Performance**
  * Reduced memory usage and improved throughput through chunked/streamed transfers and incremental disk writes.

* **API Changes**
  * Routes moved to resource-style paths; download handlers accept resize params. Client get_file now streams responses; client put_file adds a directory argument. Storage interfaces gain async streaming methods for full files and chunks.

* **Chores**
  * Updated server dependencies (actix-files, tokio-stream, tokio-util).

### DIFF
--- a/oxen-rust/src/server/src/services/versions.rs
+++ b/oxen-rust/src/server/src/services/versions.rs
@@ -25,11 +25,11 @@ pub fn versions() -> Scope {
             web::post().to(controllers::versions::chunks::complete),
         )
         .route(
-            "/{version_id}/chunks/download",
+            "/chunks/{resource:.*}/download",
             web::get().to(controllers::versions::chunks::download),
         )
         .route(
-            "/{version_id}",
+            "/{resource:.*}",
             web::get().to(controllers::versions::download),
         )
         .route(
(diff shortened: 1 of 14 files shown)


oxen-rust/src/server/src/services/versions.rs
```diff
@@ -25,11 +25,11 @@ pub fn versions() -> Scope {
             web::post().to(controllers::versions::chunks::complete),
         )
         .route(
-            "/{version_id}/chunks/download",
+            "/chunks/{resource:.*}/download",
```
jcelliott, line 28: Let's keep the `/versions` endpoints just using `version_id` and use the `/file` endpoints when we don't know the version ID already.

jcelliott (changes requested):

PR #138, opened 2025-09-07
Oxb 244/missing files

subygan:
## Summary by CodeRabbit

* **New Features**
  * New server endpoint to list missing files between two commits.
  * CLI: added --missing-files to push only missing file contents; push flow supports this option.

* **Performance**
  * Faster missing-file detection via parallelized checks.

* **Chores**
  * Updated concurrency dependency (rayon) to a newer minor version.

### DIFF
--- a/oxen-rust/src/lib/src/api/client/branches.rs
+++ b/oxen-rust/src/lib/src/api/client/branches.rs
@@ -541,8 +541,7 @@ mod tests {
             let opts = PushOpts {
                 remote: constants::DEFAULT_REMOTE_NAME.to_string(),
                 branch: new_branch_name.to_string(),
-                delete: false,
-                missing_files: false,
+                ..Default::default()
             };
 
             // Push new branch
@@ -573,8 +572,7 @@ mod tests {
             let opts = PushOpts {
                 remote: constants::DEFAULT_REMOTE_NAME.to_string(),
                 branch: "main".to_string(),
-                delete: false,
-                missing_files: false,
+                ..Default::default()
             };
 
             // Push main branch first
@@ -587,8 +585,7 @@ mod tests {
             let opts = PushOpts {
                 remote: constants::DEFAULT_REMOTE_NAME.to_string(),
                 branch: "branch-does-not-exist".to_string(),
-                delete: false,
-                missing_files: false,
+                ..Default::default()
             };
 
             // Then try to push branch that doesn't exist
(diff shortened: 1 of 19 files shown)


oxen-rust/src/lib/src/core/v_latest/push.rs
```diff
@@ -196,13 +209,47 @@ async fn push_commits(
...
     history: &[Commit],
+    opts: &PushOpts,
 ) -> Result<(), OxenError> {
     // We need to find all the commits that need to be pushed
     let node_hashes = history
         .iter()
         .map(|c| c.hash().unwrap())
         .collect::<HashSet<MerkleHash>>();
 
+    if opts.missing_files {
```
jcelliott, line 258: The `push_commits` function is already really long. Can we extract this missing files logic into a separate function?

oxen-rust/src/lib/src/api/client/branches.rs
```diff
@@ -537,13 +538,15 @@ mod tests {
             let new_branch_name = "my-branch";
             repositories::branches::create_checkout(&repo, new_branch_name)?;
 
+            let opts = PushOpts {
```
jcelliott, line 541: We've added many new instances of `PushOpts` that are mostly defaults. It would be cleaner to do:
```
let opts = PushOpts {
    branch: new_branch_name.to_string(),
    ..Default::default()
}
```

oxen-rust/src/lib/src/repositories/entries.rs
```diff
@@ -286,6 +287,30 @@ pub fn group_schemas_to_parent_dirs(
...
+        let entries = list_for_commit(repo, &commit)?;
+        all_entries.extend(entries);
+    }
+
+    all_entries.sort_by(|a, b| a.path.cmp(&b.path));
+    all_entries.dedup_by(|a, b| a.path == b.path);
+
+    let missing_files: Vec<CommitEntry> = all_entries
+        .into_par_iter()
+        .filter(|entry| !fs::version_path(repo, entry).exists())
```
jcelliott: We're trying to migrate everything away from directly accessing the `version_path`. You should use the version store instead:
```
// above the filter
let version_store = repo.version_store()?;
```
then:
```
.filter(|entry| !version_store.version_exists(entry.hash))
```

jcelliott (changes requested):

oxen-rust/src/lib/src/repositories/entries.rs
subygan: Okay, I'll change it. Should I make it such that, we use `version_store.version_exists(entry.hash)` under the hood for  `fs::version_path(repo,entry)` so that we it's applied everywhere?

jcelliott (approved):

PR #153, opened 2025-10-03
Push refactor

Lilianxr:
The performance benchmark result is tested against uploading large files(~500MB)
| file_dir | Main     | push_refactor | delta   |
|----------|----------|---------------|---------|
| 5_1      | 16s 700ms | 19s 400ms     | 16.17%  |
| 10_1     | 30s 700ms | 37s           | 20.52%  |
| 10_10    | 31s 700ms | 37s 700ms     | 18.93%  |

The refactor mainly changes the large file push logic to reuse the workspace add logic (both client and server side). At first it was a 2x slowdown when testing the above benchmark. After changing the max concurrency limit from 64 to the same as the original push (cpu_num), the performance is as listed above. The best concurrency setup needs further research.

Other changes:
1. Added async writer in the version store interface, `upload_chunk` code on the server side now support streaming. 
2. batch download endpoint now changed to 'POST /versions/fetch', the old endpoint 'GET /versions' is kept for backwards compatibility.

I suggest testing push large files(>10MB), fetch and download (clone)

## Summary by CodeRabbit

- **New Features**
  - Large file uploads now report progress during chunked transfers and use version-aware upload paths.

- **Performance**
  - Large file uploads are parallelized with improved chunk handling for faster, more reliable pushes.
  - Server streams incoming chunks directly to storage, reducing memory usage.

- **API Changes**
  - New POST endpoint for fetching versioned data: /versions/fetch. The older GET route remains deprecated but available.

- **Bug Fixes**
  - Clearer authentication errors when downloading versioned data.
  - Pull progress now properly finishes after downloads.

### DIFF
--- a/oxen-rust/src/server/src/services/versions.rs
+++ b/oxen-rust/src/server/src/services/versions.rs
@@ -7,8 +7,16 @@ pub mod chunks;
 
 pub fn versions() -> Scope {
     web::scope("/versions")
-        .route("", web::get().to(controllers::versions::batch_download))
+        .route(
+            // Deprecated. Only kept to support older clients before v0.37.2
+            "",
+            web::get().to(controllers::entries::download_data_from_version_paths),
+        )
         .route("", web::post().to(controllers::versions::batch_upload))
+        .route(
+            "/fetch",
+            web::post().to(controllers::versions::batch_download),
+        )
         .route(
             "/{version_id}/metadata",
             web::get().to(controllers::versions::metadata),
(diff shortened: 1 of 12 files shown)


oxen-rust/src/server/src/services/versions.rs
```diff
@@ -7,8 +7,16 @@ pub mod chunks;
...
-        .route("", web::get().to(controllers::versions::batch_download))
+        .route(
+            // Deprecated. Only kept to support older clients before v0.37.2
+            "",
+            web::get().to(controllers::entries::download_data_from_version_paths),
+        )
         .route("", web::post().to(controllers::versions::batch_upload))
+        .route(
+            "/fetch",
+            web::post().to(controllers::versions::batch_download),
```
subygan, line 18: Only nitpick is that, since this is a download API. GET is more appropriate.

subygan (changes requested):

oxen-rust/src/server/src/services/versions.rs
Lilianxr: Oh this is because there's a request body used here. And GET shouldn't have a body

subygan (approved):

PR #161, opened 2025-10-17
Fix issue of downloading files with same hash and different paths

Lilianxr:
Fixed the issue and added duplicate images within the test directories.

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for edge cases involving files with identical content but different names.
  * Updated test validations across clone, checkout, and pull operations to verify correct handling of larger file datasets.


gschoeni (approved):
Tested locally and looks good to me

PR #190, opened 2025-12-02
Api 49/delete file api

rpschoenburg:
Adds delete file endpoint analogous to put_file in the controller module. It stages a file for removal and commits it in one step. 

To test this, I recommend setting up a remote repo with files in it, pointing a python remote repo to it, and then using the new delete_file() method with that repo.

## Summary by CodeRabbit

* **New Features**
  * Remote file deletion with optional commit messages and branch selection
  * Uploads accept explicit commit messages with sensible defaults

* **API**
  * File service now exposes DELETE for resources

* **Tests**
  * Added tests covering remote file deletion workflows

### DIFF
--- a/oxen-python/python/oxen/remote_repo.py
+++ b/oxen-python/python/oxen/remote_repo.py
@@ -383,6 +383,8 @@ class RemoteRepo:
             branch = self.revision
         if commit_message is None:
             message = f"Removed '{file_path}'"
+        else:
+            message = commit_message
         user = oxen_user.current_user()
 
         self._repo.delete_file(branch, file_path, message, user)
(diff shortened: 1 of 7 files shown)


rpschoenburg:
With regards to the ': Potential panic on non-UTF8 paths.' CodeRabbit suggestion, those unwraps should never be None, because we return an error earlier in the function if there's not a valid path

oxen-python/python/oxen/remote_repo.py
```diff
@@ -363,6 +363,29 @@ def commit(self, message: str, branch: Optional[str] = None):
             self._workspace = None
         return commit
 
+    def delete_file(
+        self,
+        src: str,
```
jcelliott: `src` seems confusing. Maybe `target` instead?

jcelliott (commented):
Two things I think we should add:
1. Also handle deleting entire directories
2. We should make the commit message optional. If it's not specified just create a message for the user like: "Deleted foo/bar" and if there are many files something like: "Deleted foo/bar, something/else, and 17 other files"

jcelliott (approved):

PR #194, opened 2025-12-08
Api 85/oxen node panic

rpschoenburg:
This is a correction for a small issue with `oxen node`. Previously, running this command with no arguments would cause a panic, rather than being aborted by CLap or returning an error. That's now fixed, and you can also now use `oxen node --path` to lookup dir nodes by path, instead of just file nodes

To reproduce the error on main:

1. Run `oxen node` with no arguments. It will panic

## Summary by CodeRabbit

* **New Features**
  * CLI adds a dedicated --path (-p) option to locate nodes by repository path.

* **Bug Fixes**
  * Replaces the old file flag with --path and enforces mutual exclusivity with the node identifier.
  * Clearer, explicit errors when a revision, path, or node identifier cannot be found.
  * Command now requires either a path or node; verbose output still shows node children when requested.

### DIFF
--- a/oxen-rust/src/cli/src/cmd/node.rs
+++ b/oxen-rust/src/cli/src/cmd/node.rs
@@ -31,6 +31,8 @@ impl RunCmd for NodeCmd {
                 Arg::new("node")
                     .long("node")
                     .short('n')
+                    .conflicts_with("file")
+                    .required_unless_present("file")
                     .help("Node hash to inspect"),
             )
             // add --file flag
@@ -38,6 +40,8 @@ impl RunCmd for NodeCmd {
                 Arg::new("file")
                     .long("file")
                     .short('f')
+                    .conflicts_with("node")
+                    .required_unless_present("node")
                     .help("File path to inspect"),
             )
             // add --revision flag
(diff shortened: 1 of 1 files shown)


oxen-rust/src/cli/src/cmd/node.rs
```diff
@@ -53,32 +57,43 @@ impl RunCmd for NodeCmd {
...
             };
-            let node = repositories::entries::get_file(&repository, &commit, file)?;
+            let Some(node) = repositories::tree::get_node_by_path(&repository, &commit, path)?
+            else {
+                return Err(OxenError::basic_str(format!(
+                    "Error: path {:?} not found in commit {:?}",
+                    path, commit.id
+                )));
+            };
+
```
Lilianxr, line 78: It's a good suggestion

Lilianxr (changes requested):
The testing works. Fix the coderabbit suggestion and it's good to go!

oxen-rust/src/cli/src/cmd/node.rs
rpschoenburg: I agree -- it's implemented

jcelliott (approved):

PR #197, opened 2025-12-09
Api 23/workspace deadlock

rpschoenburg:
Fixes the workspace add deadlock issue.

For context, in the old iteration of parallel_batched_small_file_upload, I was using a semaphore to limit the number of outbound files on the server. This definitely was working at one point, but I'd since ran into a deadlock with consistently whereby the semaphore would never actually release its permits, getting the function stuck after the permits ran out.

This PR rebuilds that function from the ground up, simplifying the logic drastically. We now:
1. Split the small files into batches of ~AVG_CHUNK_SIZE
2. Spawn a producer tasks, which uses futures::stream with the file batches to move each batch into a blocking threadpool (tokio::task::spawn_blocking) and process them in parallel. When each batch is completed, it is sent into the mpsc pipeline
3. Spawn a consumer task, which uses tokio_stream::ReceiverStream to read from the mpsc pipeline, and creates an upload task for each read batch.
4. Collect the errors and err_files from this operation, and retry the err_files. If there are errors with the operations as a whole, or any files fail to upload after retry, return Err. Otherwise, return successfully.

There still isn't directly any retry functionality for the version file upload. That's definetly the next step here. However, this branch solves the deadlock issue by not using a semaphore while maintaining the performance benefits from the initial optimization.

To test:

1. Create a large dataset of small files. This PR doesn't relate to the large file upload at all
2. Create a remote, locally or otherwise, and create a workspace on that remote
3. Use `oxen workspace add` on the dataset as a whole. The expected behavior is for files to be processed and uploaded continuously, so long as there's still room in the upload pipeline. No files should fail after retry

Consider using different kinds of files, or otherwise different small file workloads. The directory structure shouldn't be relevant to this logic, as dirs aren't directly uploaded to workspaces 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace file upload reliability with enhanced per-batch error aggregation and retry handling
  * Fixed workspace validation to correctly return 404 for non-existent workspaces

* **Improvements**
  * Optimized uploads with streaming-based batching for better throughput and per-batch progress reporting
  * Minor logging enhancements for clearer upload diagnostics

* **Chores**
  * Removed legacy configurable global upload concurrency setting (behavior now uses streaming-based flow)

### DIFF
--- a/oxen-rust/src/lib/src/api/client/workspaces/files.rs
+++ b/oxen-rust/src/lib/src/api/client/workspaces/files.rs
@@ -489,11 +489,9 @@ async fn parallel_batched_small_file_upload(
 
                     let err_files_clone = Arc::clone(&consumer_err_files);
                     let errors = Arc::clone(&consumer_errors);
-                    
                     let bar = Arc::clone(&progress_clone);
-                    
+
                     async move {
-            
                         let result: Result<(), OxenError> = async move {
                             let (current_batch_parts, mut files_to_stage, current_batch_size) = processed_batch;
                             let num_entries = current_batch_parts.len();
@@ -565,7 +563,7 @@ async fn parallel_batched_small_file_upload(
                                             .map(|f| ErrorFileInfo {
                                                 hash: f.hash.clone(),
                                                 path: Some(f.path.clone()),
-                                                error: format!("{:?}", e),
+                                                error: format!("{e:?}"),
                                             })
                                             .collect::<Vec<ErrorFileInfo>>()
                                     );
@@ -604,7 +602,7 @@ async fn parallel_batched_small_file_upload(
     let err_file_paths: Vec<PathBuf> = err_files
         .clone()
         .into_iter()
-        .map(|f| f.path.unwrap())
+        .filter_map(|f| f.path)
         .collect();
 
     // TODO: Should we communicate to the user that we're retrying these files?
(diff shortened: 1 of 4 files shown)


Lilianxr (commented):
The testing looks good! Uploading small files via workspace add is not stuck according to debug logs. I'm thinking we should have a benchmark for this change. Maybe the baseline can be the version before the deadlock was introduced? Or just have one benchmarking against main

Lilianxr (approved):

oxen-rust/src/lib/src/api/client/workspaces/files.rs
```diff
@@ -272,381 +273,320 @@ async fn parallel_batched_small_file_upload(
...
+        repositories::commits::head_commit_maybe(local_repo)?
+    } else {
+        None
+    };
 
-    // Create a client for uploading batches
-    // let client = Arc::new(api::client::builder_for_remote_repo(&remote_repo_clone)?.build()?);
-    let progress = Arc::new(PushProgress::new_with_totals(small_files.len() as u64, 0));
+    // For individual files
     let err_files: Arc<Mutex<Vec<ErrorFileInfo>>> = Arc::new(Mutex::new(vec![]));
```
Lilianxr: tokio::sync::Mutex

PR #198, opened 2025-12-09
Default to unauthenticated client if auth_config.toml not found

rpschoenburg:
This solves an issue where you couldn't connect to any host, even unauthenticated ones, if auth_config.toml did not exist.

In current main, if auth_config.toml exists but there's no auth token for a host, the client code can still return a client with no authorization. This PR matches that behavior for if auth_config.toml doesn't exist.

To test:
1. Move or rename `~/.config/oxen/auth_config.toml`
2. Use `oxen create-remote --host localhost:3000 --name ox/no-auth --scheme http` to create a new remote. This won't work in current main, but does with this PR

You can also consider connecting to other unauthenticated hosts

## Summary by CodeRabbit

* **Bug Fixes**
  * Client now proceeds gracefully when no authentication config is found, allowing unauthenticated requests instead of failing.
  * Improved logging and clearer error messages for malformed or missing authorization headers.
  * 403 Forbidden responses now prompt supplying a valid API key, improving recovery guidance for auth failures.

### DIFF
--- a/oxen-rust/src/lib/src/api/client.rs
+++ b/oxen-rust/src/lib/src/api/client.rs
@@ -161,7 +161,7 @@ pub async fn parse_json_body(url: &str, res: reqwest::Response) -> Result<String
             Ok(config) => config,
             Err(err) => {
                 log::debug!("remote::client::new_for_host error getting config: {err}");
-                return Err(OxenError::auth_token_not_set());
+                return Err(OxenError::must_supply_valid_api_key());
             }
         };
     }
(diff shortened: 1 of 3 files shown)


jcelliott (changes requested):
Looks like this works for the unauthenticated server case, but in the case of an authenticated server (oxen.ai) it now prints a less friendly message. Let's update that fallback message (I assume it's how we're handling a `401` response from the server) to something closer to the other one.

rpschoenburg:
I believe this commit addresses the comment?

jcelliott (approved):

PR #199, opened 2025-12-09
Fix oxen restore --staged for non-glob paths

rpschoenburg:
It seems my restore refactor might have broken restore --staged for non-glob paths. This PR makes the glob module return   relative paths for `oxen restore --staged` by always feeding paths into `search_staged_db` when the stage flag is set on glob opts.

To test:
1. Do `oxen restore --staged` with different paths. All of them should now behave as expected.

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling and clearer control flow for path expansion (glob vs non-glob).
  * Safer UTF-8 handling for file paths with explicit error messages instead of panics.
  * Simplified debug output to report path counts.

* **Bug Fixes**
  * Restore targets are resolved relative to the repository root for correct staged matching.
  * Searches that require a local repository now return explicit errors if no repo is present.

### DIFF
--- a/oxen-rust/src/lib/src/util/glob.rs
+++ b/oxen-rust/src/lib/src/util/glob.rs
@@ -67,43 +67,44 @@ pub fn parse_glob_paths(
         if util::fs::is_glob_path(&glob_path) {
             if *staged_db {
                 // If staged flag set, only match against the staged db
-                let staged_paths = search_staged_db(
-                    path,
-                    repo.expect("Cannot parse staged_db for paths without a repo"),
-                )?;
+                let Some(repo) = repo else {
+                    return Err(OxenError::basic_str(
+                        "Cannot parse staged_db for paths without a repo",
+                    ));
+                };
+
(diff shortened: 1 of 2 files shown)


oxen-rust/src/lib/src/util/glob.rs
```diff
@@ -35,7 +35,7 @@ pub fn parse_glob_paths(
     };
 
     let paths = &opts.paths;
-    log::debug!("parse_glob_paths got {:?} paths", paths.len());
+    log::debug!("parse_glob_paths got {:?} paths", paths.len(),);
```
Lilianxr: extra comma

Lilianxr (changes requested):
Testing looks good! Just remove that extra comma and can be merged

oxen-rust/src/lib/src/util/glob.rs
rpschoenburg: Ah. It's removed!

Lilianxr (approved):

PR #210, opened 2025-12-11
Api 71/error messages

rpschoenburg:
I think we should look into the errors we're returning a bit more. We have an error module full of error types, but for the most part it's just OxenError::basic_str

This PR is a start, but there's definitely more to do. I've thought it's weird for a while now that the error message when you accidentally add a non-existent file is 'file must be committed to use `oxen rm`', so this clarifies that a little


jcelliott (changes requested):
I like the idea here, but I think we'll need to refine how we're handling those add/rm error messages. This is what I get when I try to add a non-existent file now:
```
> oxen add fake.txt
Error: `oxen add` could not remove files:
Error: "text2.txt" must be committed in order to use `oxen rm`
```

PR #225, opened 2025-12-19
Deduplication during pull

subygan:
## Summary by CodeRabbit

* **Refactor**
  * Fetch/deduplication now tracks file hashes across subtree traversal, reducing duplicate downloads and improving total-byte accuracy and fetch performance.
* **Chores**
  * Archive creation now uses a faster compression setting, producing archives quicker (with different compression trade-offs).

### DIFF
--- a/oxen-rust/src/lib/src/core/v_latest/fetch.rs
+++ b/oxen-rust/src/lib/src/core/v_latest/fetch.rs
@@ -8,8 +8,8 @@ use crate::core;
 use crate::core::refs::with_ref_manager;
 use crate::error::OxenError;
 use crate::model::entry::commit_entry::Entry;
-use crate::model::merkle_tree::node::{EMerkleTreeNode, FileNodeWithDir, MerkleTreeNode};
-use crate::model::{Branch, Commit, CommitEntry};
+use crate::model::merkle_tree::node::{EMerkleTreeNode, MerkleTreeNode};
+use crate::model::{Branch, Commit, CommitEntry, MerkleHash};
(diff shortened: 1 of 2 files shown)


oxen-rust/src/lib/src/core/v_latest/fetch.rs
```diff
@@ -259,13 +303,12 @@ fn collect_missing_entries(
...
             );
             for subtree_path in subtree_paths {
-                let mut unique_hashes = HashSet::new();
                 let Some(tree) = repositories::tree::get_subtree_by_depth_with_unique_children(
                     repo,
                     commit,
                     subtree_path,
-                    Some(&shared_hashes),
-                    Some(&mut unique_hashes),
+                    None,
```
jcelliott: This seems wrong to me. I thought the main purpose of this `get_subtree_by_depth_with_unique_children` function was to avoid walking the entire merkle tree for every commit. It seems like we're just disabling this optimization in order to do the aggregation at the top level instead. @rpschoenburg I know you worked on this as well, do you have any input?

rpschoenburg (approved):
With the latest commit, I think this is good. The prior optimization (skipping loading parts of the merkle tree that have already been loaded into memory) is restored, and this still does the de-dup. 

Considering the need for it in multiple places now, I'm thinking I'll go ahead and make a new tree loading method to collect all the file nodes when we're reading the tree into memory, which would eliminate the need for the tree traversal in collect_missing_entries. But, I think this is good for now

subygan:
@rpschoenburg  exactly what I was thinking, we now have it in pull and push. so would be nice to have that.

PR #228, opened 2025-12-23
Api/collect files

rpschoenburg:
Implements the new tree loading function to collect file nodes while loading the merkle tree. This resolves the TODOs in the push and fetch code. This should be functionally identical to main, except with simplified logic in these modules and a small optimization to avoid the extra tree traversals

To test:

Run `oxen push` and `oxen fetch` with the same test cases that resulted in duplicate files being pushed originally. They should not upload/download duplicate files

## Summary by CodeRabbit

* **Refactor**
  * Tree loading and depth-limited path resolution now collect file-level leaf entries during traversal.
  * Commit/push preparation aggregates directory-node hashes and derives missing file entries from collected file nodes instead of re-walking subtrees.
  * User impact: more efficient syncs with reduced redundant traversals and smaller transferred payloads.


oxen-rust/src/lib/src/core/v_latest/push.rs
subygan: @rpschoenburg  we can replace this as well with `get_subtree_by_depth_with_unique_file_nodes()`

oxen-rust/src/lib/src/core/v_latest/index/commit_merkle_tree.rs
```diff
@@ -1135,6 +1267,104 @@ impl CommitMerkleTree {
         Ok(())
     }
 
+    #[allow(clippy::too_many_arguments)]
+    fn load_children_and_collect_file_nodes(
```
subygan: I wonder if instead of duplicating this whole traversal we can have something like `include_files=true` argument to  `get_subtree_by_depth_with_unique_children()`.

having too many traversal code makes it difficult to add optimization etc.

oxen-rust/src/lib/src/core/v_latest/index/commit_merkle_tree.rs
```diff
@@ -1119,6 +1251,104 @@ impl CommitMerkleTree {
...
+                            requested_depth,
+                            traversed_depth,
+                        )?;
+                    }
+
+                    node.children.push(child);
+                }
+                // FileChunks and Files are leaf nodes
+                MerkleTreeNodeType::FileChunk | MerkleTreeNodeType::File => {
+                    if let EMerkleTreeNode::File(file_node) = &child.node {
```
Eric-Laurence, line 1339: Won't this drop FileChunks?

oxen-rust/src/lib/src/core/v_latest/push.rs
```diff
@@ -302,60 +302,50 @@ async fn get_commit_missing_hashes(
...
                 }
-            });
-        }
+            }
 
-        // After processing all paths for this commit, update base_hashes
-        base_hashes.extend(unique_hashes);
-        base_hashes.extend(file_hashes_seen);
+            base_hashes.extend(unique_hashes.clone());
+            dir_node_hashes.extend(unique_hashes);
```
Eric-Laurence, line 336: Is this clone necessary? Can't we just borrow it?

```suggestion
            base_hashes.extend(&unique_hashes);
            dir_node_hashes.extend(unique_hashes);
```

Eric-Laurence (commented):
I think it's worth looking at the CodeRabbit comments, I didn't go over them in depth but some of them look like valid concerns.

PR #231, opened 2025-12-30
Api 101/perf monitoring

subygan:
## Summary by CodeRabbit

* **New Features**
  * Optional performance logging: lightweight instrumentation across core and server paths, initialized at startup and controllable at runtime via OXEN_PERF_LOGGING; disabled by default with negligible overhead.

* **Chores**
  * Development and runtime images updated to build/run with the performance-logging option.
  * Downstream bindings updated to opt into performance logging.

### DIFF
--- a/oxen-rust/src/lib/src/core/v_latest/commits.rs
+++ b/oxen-rust/src/lib/src/core/v_latest/commits.rs
@@ -435,8 +435,11 @@ pub fn list_from(
     repo: &LocalRepository,
     revision: impl AsRef<str>,
 ) -> Result<Vec<Commit>, OxenError> {
+    let _perf = crate::perf_guard!("core::commits::list_from");
+
     let revision = revision.as_ref();
     if revision.contains("..") {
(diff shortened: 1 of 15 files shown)


oxen-rust/src/lib/src/core/v_latest/commits.rs
```diff
@@ -446,9 +449,14 @@ pub fn list_from(
             .ok_or(OxenError::revision_not_found(head.into()))?;
         return list_between(repo, &base_commit, &head_commit);
     }
-    let mut results = vec![];
+
+    let _perf_get = crate::perf_guard!("core::commits::get_revision");
     let commit = repositories::revisions::get(repo, revision)?;
+    drop(_perf_get);
```
subygan: Calling `drop()` specifically, records the end of that function we are recording.

In case we don't call `drop()`, at the end of the scope when the variable is collected, the end of the function is clculated.

jcelliott (approved):

PR #236, opened 2026-01-06
Update environment variable names

rpschoenburg:
Deprecates the 'MAX_RETRIES' environment variable. We had 2 different variables  to configure the number of HTTP Request retries, so this deprecates one of them. Also fixed the help message for the clone depth option (--subtree isn't a path anymore, it's called --filter now)

## Summary by CodeRabbit

* **Documentation**
  * Clarified the clone command help to better describe the --depth option for subtree cloning.

* **Chores / Configuration**
  * Renamed the environment variable that controls retry behavior and removed the old fixed retry constant; default retry count for non-test builds changed from 10 to 5, and client retry fallbacks now use the updated lookup.

### DIFF
--- a/oxen-rust/src/lib/src/constants.rs
+++ b/oxen-rust/src/lib/src/constants.rs
@@ -226,11 +226,11 @@ pub fn max_retries() -> usize {
             max_retries
         } else {
             // If parsing failed, fall back to default
-            NUM_HTTP_RETRIES
+            NUM_HTTP_RETRIES.try_into().unwrap()
         }
     } else {
         // Environment variable not set, use default
-        NUM_HTTP_RETRIES
+        NUM_HTTP_RETRIES.try_into().unwrap()
     }
 }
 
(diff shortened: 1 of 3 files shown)


oxen-rust/src/lib/src/constants.rs
```diff
@@ -178,9 +178,6 @@ pub const EVAL_DURATION_COL: &str = "_oxen_eval_duration";
 pub const AVG_CHUNK_SIZE: u64 = 1024 * 1024 * 10;
 // Retry and back off of upload tasks N times
 /// Retry and back off of upload tasks N times
-pub const MAX_RETRIES: usize = 5;
-// Allow up to N concurrent upload tasks
```
jcelliott, line 182: It looks like you deleted the wrong comments here. Also, 5 retries seems like plenty to me. Should that be the default instead of 10? Seems like 10 will just put more load on the server and it doesn't seem likely that a request will succeed after 10 retries but not 5.

rpschoenburg: Fixed the comments and went down to 5 on NUM_HTTP_REQUESTS. It previously was 10, so I just kept it at that, but I agree that seems excessive

jcelliott (approved):

PR #249, opened 2026-01-23
Allow creating workspace on an empty repo, Workspace create endpoint refactor

jcelliott:
## Summary by CodeRabbit

* **New Features**
  * Create an initial commit automatically for empty repositories.

* **Improvements**
  * Workspace creation during clone supports explicit workspace names and better status reporting.
  * CLI: commit and workspace commands accept an optional --branch flag and show clearer errors when no branch is specified.
  * Workspace creation endpoints are now idempotent (safe to retry).

* **Deprecations**
  * POST /workspaces deprecated; prefer PUT /workspaces (get-or-create).

### DIFF
--- a/oxen-rust/src/server/src/controllers/workspaces.rs
+++ b/oxen-rust/src/server/src/controllers/workspaces.rs
@@ -2,8 +2,9 @@ use crate::errors::{OxenHttpError, WorkspaceBranch};
 use crate::helpers::get_repo;
 use crate::params::{app_data, path_param, NameParam};
 
+use liboxen::constants::INITIAL_COMMIT_MSG;
 use liboxen::error::OxenError;
-use liboxen::model::NewCommitBody;
+use liboxen::model::{NewCommitBody, User};
 use liboxen::repositories;
 use liboxen::view::merge::MergeableResponse;
 use liboxen::view::workspaces::{ListWorkspaceResponseView, NewWorkspace, WorkspaceResponse};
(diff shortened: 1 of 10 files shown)


Eric-Laurence:
@jcelliott does this replace #246?

Eric-Laurence:
Oh, I just saw that you closed that.

oxen-rust/src/server/src/controllers/workspaces.rs
```diff
@@ -62,13 +63,41 @@ pub async fn get_or_create(
...
+                "get_or_create: empty repo, creating initial commit on branch {}",
                 data.branch_name
-            ))),
-        );
+            );
+            let user = User {
+                name: "Oxen".to_string(),
+                email: "oxen@oxen.ai".to_string(),
+            };
+            repositories::commits::create_initial_commit(
```
subygan, line 90: I wonder if we can move this to repo init flow directly. As it would just remove this class of problems completely.

because the file upload `/api/repos/{namespace}/{repo_name}/file/{resource}` also errors out with a 500 in a new repo.

subygan (approved):

oxen-rust/src/server/src/controllers/workspaces.rs
jcelliott: Earlier versions of Oxen actually did create an initial commit. But that causes other issues, for example the case where you have an existing local repo that you want to push to a new remote repo.

The file endpoint you mentioned (and other workspace endpoints) seem like they should work with a new repo though. I wonder if we can just centralize handling that on _workspace creation_ rather than trying to tackle all these individually.

subygan: ah! yes, I hadn't considered that scenario. makes sense.

PR #250, opened 2026-01-27
Tree API for recursive directory listing

subygan:
## Summary by CodeRabbit

* **New Features**
  * Depth-controlled directory listing so users can request traversal depth (0 = current, 1 = immediate children, -1 = unlimited).
  * Directory metadata now includes nested children for hierarchical representation.
  * API query parameters updated to accept an optional depth value for listings.

* **Tests**
  * Added tests validating depth-controlled listings across multiple nesting levels.

### DIFF
--- a/oxen-rust/src/lib/src/core/v_latest/entries.rs
+++ b/oxen-rust/src/lib/src/core/v_latest/entries.rs
@@ -394,6 +394,7 @@ fn file_node_to_metadata_entry(
     }))
 }
 
+#[allow(clippy::too_many_arguments)]
 fn p_dir_entries(
     repo: &LocalRepository,
     node: &MerkleTreeNode,
--- a/oxen-rust/src/lib/src/core/v_latest/entries.rs
+++ b/oxen-rust/src/lib/src/core/v_latest/entries.rs
@@ -45,6 +45,16 @@ pub fn list_directory(
     parsed_resource: &ParsedResource,
     paginate_opts: &PaginateOpts,
 ) -> Result<PaginatedDirEntries, OxenError> {
+    list_directory_with_depth(repo, directory, parsed_resource, paginate_opts, 0)
+}
+
+pub fn list_directory_with_depth(
+    repo: &LocalRepository,
+    directory: impl AsRef<Path>,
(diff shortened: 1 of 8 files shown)


oxen-rust/src/server/src/services/tree.rs
```diff
@@ -39,4 +39,5 @@ pub fn tree() -> Scope {
             web::get().to(controllers::tree::download_tree_nodes),
         )
         .route("/download", web::get().to(controllers::tree::download_tree))
+        .route("/{resource:.*}", web::get().to(controllers::dir_tree::get))
```
jcelliott: We can't use `/tree/{resource:.*}` because it's shadowed by the other tree endpoints. For example if someone has a branch called "download", it will be impossible to get the tree for that branch because it's shadowed by the `/tree/download` endpoint.

jcelliott: We have `/tree/nodes`, maybe this endpoint could be `/tree/files/{resource:.*}`?

jcelliott (changes requested):
Can you add a client module function to call this endpoint and add a unit test there?

oxen-rust/src/server/src/services/tree.rs
jcelliott: Or maybe it should just be the same as the `/dir` endpoint, but with a `depth` query? What do you think?

subygan: I had this implemented first. might be a good idea. but also the "tree" structure is similar to github's tree API.
and the dir can remain without any issue. but yes, tree api with `depth=1` returns the same data as the `dir` api

subygan: I've implemented it in the dir API for now with the test.
something for future decisions, gitlab uses a `/-/` to separate regular and user APIs, which I think is pretty smart. repo names as `-` don't make sense.

[gitlab doc](https://docs.gitlab.com/development/routing/#:~:text=To%20eliminate%20the%20ambiguity%20and%20to%20make%20the%20backend%20easier%20to%20maintain)

jcelliott (approved):
LGTM!

PR #255, opened 2026-02-03
oxen-rust Cargo Workspace Tweaks & Fixes

malcolmgreaves:
**Fix cargo workspace**
- **No more `Oxen` project**
  + It is now a virtual project as it has been fully converted to a workspace-only
- Binaries & lib crates are already handled by `--workspace`. No need to re-declare them in the workspace's `Cargo.toml`.
- Crates properly use workspace dependencies: remove redundant listings & ensure version reuse.

**`oxen-bench`**
- Moved all cross-crate benchmarking code to new crate: `oxen-bench`
- Kept as `src/benches`
- _Non-publishable crate_

**Bugfix**
- `oxen-server` was missing the `async-zip` dependency.

**Other**
- Updated to use `cargo ... --workspace` in:
  + CI
  + pre-commit hooks
  + documentation / README mentions
- Updated builds that enable features to use proper syntax (`crate/feature`)
- Updated GHA release workflows
- Make all experimental crates explicitly not publishable.
- Removed unused dependencies with `cargo machete`.
- Top-level "virtual"/workspace is marked as not publishable.
- Experiments are **not** included in workspace
  + each are independent projects


oxen-rust/src/benches/src/oxen/workspace_add.rs
```diff
@@ -196,7 +196,7 @@ pub fn workspace_add_benchmark(c: &mut Criterion, data: Option<String>, iters: O
                 let name = format!("{branch_name}: {workspace_id}");
 
                 let workspace = rt
-                    .block_on(api::client::workspaces::create_with_new_branch(
+                    .block_on(api::client::workspaces::create_with_path(
```
malcolmgreaves, line 199: bugfix -- `create_with_new_branch` doesn't exist

jcelliott: Nice catch! I removed that last week but didn't catch the bench usage.

malcolmgreaves:
A few other ideas:

(1) Rename all directories to match its respective rust crate name (_less cognitive overhead, easier to script actions between crate <--> filesystem, no need for name lookup table_)
  + `src/tests` -> `src/oxen-test`
  + `src/lib` -> `src/liboxen`
  + `src/server` -> `src/oxen-server`
  + `src/cli` -> `src/oxen-cli`
  + `src/bench` -> `src/oxen-bench`
  
(2) Make an `Oxen` crate that depends on `liboxen`, `oxen-cli`, and `oxen-server`. This crate just re-exports `liboxen` as a library and re-exports the `oxen` CLI and `oxen-server` binaries. This would re-create the existing `Oxen` package semantics.

(3) A modification of (2): only re-export the `oxen` and `oxen-server` binaries with a _dependency_ on `liboxen` (_no re-exporting of the library_). We'd be able to do `cargo install oxen` and let people get the binaries to use on their system. If someone wants to, they can depend on `liboxen` only or `liboxen` + `Oxen` in their Rust project.

oxen-rust/src/lib/src/config/auth_config.rs
```diff
@@ -1,11 +1,12 @@
 use crate::constants::{CONFIG_DIR, DEFAULT_HOST, OXEN};
 use crate::error::OxenError;
+use crate::test::REPO_ROOT;
```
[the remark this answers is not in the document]
malcolmgreaves: Agreed on this point. However, this requires some more significant changes. I am not changing existing functionality in this PR -- this runtime dependency on test-time only things was present _before_ this PR. This PR is just making the path resolution for test-time things explicit and correct (since it's now trying to look for things at `oxen-rust/src/lib/` instead of the repository root).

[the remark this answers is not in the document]
malcolmgreaves: In a follow-up PR, we should do the following:

- Deprecate all of the `get` that do default initialization and replace with `impl Default for ... `.
  + This is the idiomatic way in Rust to declare a type's default initialization.
  
- Remove all `if env("TEST").ok() { ... }` logic and replace with explicit initializers for test-only code.
  + Should be **impossible** to trigger test-only things from production: having it gated behind an undocumented env var is surprising.
  + There's no need to inject anything at runtime: we should simply have functions that let us pass-in paths. At test time, we just pass in the test-specific paths.

oxen-rust/src/oxen-test/src/test.rs
```diff
@@ -0,0 +1,61 @@
...
+        test::run_training_data_repo_test_fully_committed_async(|repo| async move {
+            // Add a file that we are going to ignore
+            let ignore_dir = "ignoreme/";
+            let ignore_path = repo.path.join(ignore_dir);
+            std::fs::create_dir(&ignore_path)?;
+            test::write_txt_file_to_path(ignore_path.join("0.txt"), "I should be ignored")?;
+            test::write_txt_file_to_path(ignore_path.join("1.txt"), "I should also be ignored")?;
+
+            let oxenignore_file = repo.path.join(".oxenignore");
+            test::write_txt_file_to_path(oxenignore_file, "ignoreme.txt")?;
```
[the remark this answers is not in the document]
malcolmgreaves: Fixed.

malcolmgreaves:
@coderabbitai please re-review this entire PR

malcolmgreaves:
NOTE: Wait to merge this until after these are merged:
-  https://github.com/Oxen-AI/Oxen/pull/260
- https://github.com/Oxen-AI/Oxen/pull/259
- https://github.com/Oxen-AI/Oxen/pull/257

oxen-rust/src/cli/README.md
```diff
@@ -0,0 +1 @@
+../../README.md
```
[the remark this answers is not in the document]
malcolmgreaves: This is how git handles symlinks. Suggested change is wrong.

oxen-rust/src/lib/src/config/user_config.rs
```diff
@@ -1,10 +1,11 @@
 use crate::constants::{CONFIG_DIR, OXEN};
 use crate::error::OxenError;
 use crate::model::User;
+use crate::test::REPO_ROOT;
```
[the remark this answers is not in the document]
malcolmgreaves: Agreed, but this requires changing the code to not rely on the `TEST` env var. The code needs to be changed so that the path can be provided by the caller. And the `get()` should change to have default behavior implemented with an `impl Default`.

This PR preserves existing functionality. These suggested changes are for another day.

oxen-rust/experiments/block-level-dedup/Cargo.toml
```diff
@@ -1,29 +1,21 @@
...
-thiserror = "2.0.12"
-xxhash-rust = { version = "0.8.7", features = ["xxh3"] }
-liboxen  = { path = "../../src/lib" }
-flate2 = { version = "1.1.0", default-features = false, features = ["zlib-ng"] }
-async-trait = "0.1.88"
-fastcdc = "3.2.1"
+edition = {workspace=true}
+publish = false
 
+[workspace]
```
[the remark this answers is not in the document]
malcolmgreaves: This is to prevent it from being included in the workspace.

oxen-rust/experiments/lmdb-oxen/Cargo.toml
```diff
@@ -2,9 +2,9 @@
 name = "lmdb-oxen"
 version = "0.1.0"
 edition = "2024"
```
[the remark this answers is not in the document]
malcolmgreaves: This is what the edition was before. It is not part of the workspace so this is ok.

oxen-rust/data/test/config/user_config.toml
malcolmgreaves, line 1: I keep running into this issue -- all tests fail because the repo isn't initialized. And the only thing missing for initialization is this `user_config.toml` file.

.github/workflows/release.yml
```diff
@@ -55,7 +55,7 @@ jobs:
     needs: set_env
     uses: ./.github/workflows/release_docker.yml
     with:
-      ARCHITECTURE: x86_64
+      ARCHITECTURE: amd64
```
malcolmgreaves, line 58: Pre-existing bug: there's no `x86_64` target. Docker decided to be confusing and name this `amd64`.

oxen-rust/data/test/config/user_config.toml
malcolmgreaves: cc @jcelliott @gschoeni should we have this in the test data? Or is there a reason why we need to do `add-user`?

oxen-rust/tests/s3_tests.rs
malcolmgreaves, line 1: Moved

oxen-rust/tests/test.rs
malcolmgreaves, line 1: Moved

oxen-rust/tests/s3_tests.rs
malcolmgreaves, line 1: Not deleted, moved.

oxen-rust/tests/test.rs
malcolmgreaves, line 1: Not deleted, moved.

oxen-rust/Cargo.toml
```diff
@@ -82,13 +58,16 @@ clap = { version = "4.4.2", features = ["cargo", "derive"] }
...
+criterion = { version = "2.10.1", package = "codspeed-criterion-compat", features = [
+    "async_tokio",
+] }
 deadqueue = "0.2.4"
 derive_more = { version = "1.0.0", features = ["full"] }
 dialoguer = "0.11.0"
 difference = "2.0.0"
 dirs = "5.0.1"
 dotenv = "0.15.0"
-duckdb = { package = "duckdb", version = "=1.1.1", default-features = false, optional = true, features = [
```
malcolmgreaves, line 91: Workspace dependencies aren't optional. Crates determine if they need them and whether or not they're optional.

oxen-rust/data/test/config/user_config.toml
jcelliott: I think it would be fine to just add that directly to the repo.

CleanCut (changes requested):
@malcolmgreaves I love the energy and focus on improvement! I'm going to jump in without having done a full review, since it sounds like you're pushing to merge this quickly and I have some concerns about merging as-is.

- This would be better to split this into multiple PRs. For anyone coming fresh to this PR, it's a bit overwhelming. I know you've expressed that you feel it is mostly just changes to lockfiles, but there is a ton of comments and changes to wade through beyond.
- There's way too much going on for a single PR. There's at least a workspace refactor, multiple instance of unrelated code moves, and multiple kinds of fixes, some related to the code moves and some not. And perhaps other things going on because I haven't gone all the way through the PR.
- Moving the tests into separate crates is an anti-pattern. Let's please not do that. I'm having trouble figuring out the concern that prompted that. Could you explain the concern so we can figure out another solution?
- Regarding the workspace refactor, I think this merits a discussion on the best way to do this.
  - For example, if this is primarily a Rust project, best practice is to have a top-level `crates/` directory, instead of having most Rust code rooted under `oxen-rust/`.  If that doesn't make sense in this project for `$REASONS`, then `oxen-rust/src` ought to become `oxen-rust/crates`.

I'm happy to put my money where my mouth is and help out with this. I'd love to get us on a clean, nice workspace design. Would you like to pair on this? Or if you're feeling pressured with the other stuff you're working on, would it help if I put up some PR(s)?

jcelliott (approved):

oxen-rust/src/server/Cargo.toml
```diff
@@ -1,7 +1,14 @@
...
-edition = "2021"
+version = { workspace = true }
+edition = { workspace = true }
+license-file = { workspace = true }
+description = { workspace = true }
+homepage = { workspace = true }
+repository = { workspace = true }
+readme = { workspace = true }
+keywords = { workspace = true }
+categories = { workspace = true }
```
malcolmgreaves: Fix categories and look over keywords again.

oxen-rust/tests/s3_tests.rs
malcolmgreaves, line 1: Move into `src/liboxen/`

oxen-rust/tests/test.rs
malcolmgreaves, line 1: Move into `src/liboxen/`

oxen-rust/Cargo.toml
```diff
@@ -1,9 +1,20 @@
...
+    "src/server",
+]
+exclude = ["experiments/*"]
+
+[workspace.package]
+version = "0.43.1"
 edition = "2021"
 license-file = "LICENSE"
-description = "Oxen is a fast, unstructured data version control, to help version large machine learning datasets written in Rust."
+description = "Oxen is a fast, unstructured data version control, to help version datasets, written in Rust."
```
malcolmgreaves: Revert to `"Oxen is a fast, unstructured data version control, to help version large machine learning datasets written in Rust."`

oxen-rust/Cargo.toml
```diff
@@ -1,9 +1,20 @@
...
+    "src/cli",
+    "src/lib",
+    "src/oxen-test",
+    # ^^ test-only crate: never publish!
+    "src/server",
+]
+exclude = ["experiments/*"]
+
+[workspace.package]
+version = "0.43.1"
```
malcolmgreaves: Fix -- version regression.

CleanCut (approved):
Approving after our slack meeting. Items we discussed:

- breaking: version number regressed

Items that should be done soon, but could potentially be in a different PR

- revert oxen-test crate creation / code move
- revert benchmark crate creation / code move
- specify individual `categories` entries for published crates

PR #260, opened 2026-02-06
Fix `workspaces::files::exist`: `Err` on empty workspace & false-positives

malcolmgreaves:
**Bug: `exists()` `Err` on Empty Workspace**
The `repositories::workspace::files::exists` function currently fails with `Err` if
there are no staged changes in the workspace. It fails because the RocksDB instance
in the workspace is not initialized when the workspace is created.

Other workspace options (e.g. `add`) use `with_staged_db_manager` to ensure that
the DB is present. `exists` has been changed to use this function.

Now `exists` will return `Ok(false)` for any path in a newly created workspace.

**Bug: `exists` false-positives**
Previously, `exists` could return a false-positive for some path existsing in the
workspace. This is because it used `key_may_exist`, which is a bloom filter
checking for path existence. It now uses a `get` call, via `read_from_staged_db`,
 which is exact.

## Summary by CodeRabbit

* **Bug Fixes**
  * File existence and delete checks now consistently reflect the workspace's staged state using an optimized staged-data lookup, reducing unnecessary I/O and improving reliability.

* **New Features**
  * Staged-data manager provides a fast existence check that avoids I/O when a file is clearly absent.

* **Tests**
  * Added tests verifying existence is false when nothing is staged and flips true after staging.

### DIFF
--- a/oxen-rust/src/lib/src/core/v_latest/workspaces/files.rs
+++ b/oxen-rust/src/lib/src/core/v_latest/workspaces/files.rs
@@ -1,5 +1,3 @@
-use rocksdb::{DBWithThreadMode, MultiThreaded};
-
 use actix_web::{web, Error};
 use futures::StreamExt;
 use parking_lot::Mutex;
@@ -12,13 +10,12 @@ use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 use zip::ZipArchive;
 
-use crate::constants::STAGED_DIR;
+use crate::core;
 use crate::core::staged::staged_db_manager::with_staged_db_manager;
 use crate::core::v_latest::add::{
     add_file_node_to_staged_db, get_file_node, get_status_and_add_file,
     process_add_file_with_staged_db_manager, stage_file_with_hash,
 };
-use crate::core::{self, db};
 use crate::error::OxenError;
 use crate::model::file::TempFilePathNew;
 use crate::model::merkle_tree::node::EMerkleTreeNode;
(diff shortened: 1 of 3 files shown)


oxen-rust/src/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -216,13 +216,33 @@ pub fn exists(workspace: &Workspace, path: impl AsRef<Path>) -> Result<bool, Oxe
...
+    let staged_db: DBWithThreadMode<MultiThreaded> = DBWithThreadMode::open_for_read_only(
+        &db::key_val::opts::default(),
+        dunce::simplified(&db_path),
+        false,
+    )?;
 
     let path = util::fs::path_relative_to_dir(path, &workspace_repo.path)?;
-    let relative_path_str = path.to_str().unwrap();
+
+    let relative_path_str = path.to_string_lossy().to_string();
```
subygan: this is a nice change.

subygan (approved):

oxen-rust/src/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -203,28 +200,19 @@ pub async fn remove_files_from_staged_db(
...
-    let staged_db: DBWithThreadMode<MultiThreaded> =
-        DBWithThreadMode::open_for_read_only(&opts, dunce::simplified(&db_path), false)?;
-
-    let path = util::fs::path_relative_to_dir(path, &workspace_repo.path)?;
-    let relative_path_str = path.to_str().unwrap();
-    let result = staged_db.key_may_exist(relative_path_str);
-    Ok(result)
+    let path = util::fs::path_relative_to_dir(path.as_ref(), &workspace_repo.path)?;
+    with_staged_db_manager(workspace_repo, |staged_db_manager| {
+        staged_db_manager.exists(&path)
```
malcolmgreaves, line 214: @subygan I ended up wanting to make sure that there was a more fool-proof way of making sure we `Ok(false)` when the DB doesn't exist. I found this pattern of using `with_staged_db_manager` , which ensures the DB exists before running the function. 

I also got claude to point out that `key_may_exist` uses a bloom filter, which can produce false positives. It suggested using `.get` which is an exact lookup. I opted to make a new method `exists` on `StagedDBManager` to do an optimal exact existence check: it uses the bloom filter first (avoiding I/O) to determine if the key is definitely not present. If it may be present, then it does I/O to get the value. But it uses `get_pinned` to avoid a copy into the `Optional`.

PR #264, opened 2026-02-11
Add repository and namespace name validation

jcelliott:
I saw some flaky unit tests caused by trying to create a repository with a space in the name. This adds validation for repo and namespace names to match what we currently require on OxenHub (`^[[:alnum:]][[:alnum:]_.-]+$`)

## Summary by CodeRabbit

* **New Features**
  * Repository creation now validates names and namespaces (start with alphanumeric, only alphanumeric/underscore/dot/hyphen, min length 2).
  * Invalid names are rejected with descriptive error messages during creation.

* **Bug Fixes**
  * Creation endpoints now return a clear 400 Bad Request with a standardized "invalid_repo_name" error when names are invalid.

### DIFF
--- a/oxen-rust/src/lib/src/repositories.rs
+++ b/oxen-rust/src/lib/src/repositories.rs
@@ -207,6 +207,22 @@ pub async fn create(
     root_dir: &Path,
     new_repo: RepoNew,
 ) -> Result<LocalRepositoryWithEntries, OxenError> {
+    // Validate repo name does not contain spaces
+    if new_repo.name.contains(' ') {
+        return Err(OxenError::invalid_repo_name(format!(
+            "Repository name '{}' cannot contain spaces",
+            new_repo.name
+        )));
+    }
+
+    // Validate namespace does not contain spaces
+    if new_repo.namespace.contains(' ') {
+        return Err(OxenError::invalid_repo_name(format!(
+            "Namespace '{}' cannot contain spaces",
+            new_repo.namespace
+        )));
+    }
+
     let repo_dir = root_dir
         .join(&new_repo.namespace)
         .join(Path::new(&new_repo.name));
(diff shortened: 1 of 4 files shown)


Eric-Laurence (approved):
Nice catch. I love improving unit tests!

oxen-rust/src/lib/src/repositories.rs
```diff
@@ -203,10 +203,42 @@ pub fn transfer_namespace(
...
+    if name.len() < 2 {
+        return false;
+    }
+    let mut chars = name.chars();
+    let first = chars.next().unwrap();
+    if !first.is_ascii_alphanumeric() {
+        return false;
+    }
+    chars.all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '.' || c == '-')
+}
```
malcolmgreaves: This is fine as-is. Since you already have the pattern you want as a regular expression, have you considered doing regular expression matching instead?

If not, you could use the following to do it:
```
use std::sync::LazyLock;
use regex::Regex;

static VALID_REPO_NAME_RE: LazyLock<Regex> = LazyLock::new(|| {
    Regex::new(r"^[[:alnum:]][[:alnum:]_.\-]+$").unwrap()
});

fn is_valid_repo_name(name: &str) -> bool {
  VALID_REPO_NAME_RE.is_match(name)
}
```

The regular expression will be compiled once on first use and reused on subsequent `is_valid_repo_name` calls.

oxen-rust/src/lib/src/repositories.rs
```diff
@@ -203,10 +203,42 @@ pub fn transfer_namespace(
...
 pub async fn create(
     root_dir: &Path,
     new_repo: RepoNew,
 ) -> Result<LocalRepositoryWithEntries, OxenError> {
+    // Validate repo name matches ^[[:alnum:]][[:alnum:]_.-]+$
+    if !is_valid_repo_name(&new_repo.name) {
+        return Err(OxenError::invalid_repo_name(format!(
+            "Invalid repository name '{}'. Names must be at least 2 characters and match [a-zA-Z0-9][a-zA-Z0-9_.-]+",
+            new_repo.name
+        )));
```
malcolmgreaves: Suggestion -- it can be easier to keep the actual invalid repo name in the new OxenError variant directly, then construct the appropriate error message when calling `std::fmt::Display`.

Doing this lets one programmatically determine what was the offending invalid name versus only having the human-readable error message.

Changing that would be:
- the enum variant is `InvalidRepoName(String)`
- the doc comment states to put the invalid repository name as the argument

Then having this addition to the existing `impl std::fmt::Display for OxenError`

```
...
   OxenError::InvalidRepoName(n) => write!(f, "Invalid namespace '{}'. Namespaces must be at least 2 characters and match [a-zA-Z0-9][a-zA-Z0-9_.-]+", n),
...
```

oxen-rust/src/lib/src/repositories.rs
```diff
@@ -203,10 +203,42 @@ pub fn transfer_namespace(
...
+    }
+
+    // Validate namespace matches the same pattern
+    if !is_valid_repo_name(&new_repo.namespace) {
+        return Err(OxenError::invalid_repo_name(format!(
+            "Invalid namespace '{}'. Namespaces must be at least 2 characters and match [a-zA-Z0-9][a-zA-Z0-9_.-]+",
+            new_repo.namespace
+        )));
```
malcolmgreaves: As a side benefit, this would become shorter:
```
return Err(OxenError::invalid_repo_name(new_repo.namespace))
```

oxen-rust/src/lib/src/repositories.rs
```diff
@@ -448,6 +480,75 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_local_repository_api_create_rejects_invalid_name() -> Result<(), OxenError> {
+        test::run_empty_dir_test_async(|sync_dir| async move {
+            let namespace = "test-namespace";
+            let name = "repo with spaces";
+            let repo_new = RepoNew::from_namespace_name(namespace, name, None);
+            let result = repositories::create(&sync_dir, repo_new).await;
+
+            assert!(result.is_err());
```
malcolmgreaves:
```suggestion
            assert!(result.is_err(), "{:?}", result);
```
Minor suggestion -- if this assert fails, we won't really know what `result` was (just that it was some `Ok`). It can be helpful to also print out what the `result` is when the test fails.

oxen-rust/src/lib/src/repositories.rs
```diff
@@ -448,6 +480,75 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_local_repository_api_create_rejects_invalid_namespace() -> Result<(), OxenError> {
+        test::run_empty_dir_test_async(|sync_dir| async move {
+            let namespace = "-invalid-namespace";
+            let name = "valid-repo";
+            let repo_new = RepoNew::from_namespace_name(namespace, name, None);
+            let result = repositories::create(&sync_dir, repo_new).await;
+
+            assert!(result.is_err());
```
malcolmgreaves: Same minor suggestion here.

malcolmgreaves (approved):
LGTM! Nice job finding and fixing this bug! I have a few non-blocking suggestions for consideration.

oxen-rust/src/lib/src/repositories.rs
jcelliott: I was trying to avoid adding a dependency, but I see we actually already have the regex library, so I agree your suggestion is nicer.

malcolmgreaves: Ah that makes sense! It's fine as-is IMO -- if it works it works :)

jcelliott:
@malcolmgreaves great suggestions! Updated with those changes.

PR #265, opened 2026-02-12
Allow Python Workspace.add to accept multiple files & dirs

malcolmgreaves:
**Python `Workspace.add`**
Changes `class Workspace`'s `add` to accept multiple file or directory inputs.
For any directory encountered, it still recursively walks its contents and adds all accessible files.

Additionally, `add` now accepts `pathlib.Path` instances too. 
New tests cover the new add multiple behavior.

**Small, unrelated changes**:
- `api::client::workspaces::files::add_many` didn't need to take ownership of `paths`, so it is now a slice
- tweaks to the ignore file
- updated type annotations in `oxen-python/python/oxen/workspace.py`
- using `Path` in the Python workspace tests

### DIFF
--- a/oxen-python/python/oxen/workspace.py
+++ b/oxen-python/python/oxen/workspace.py
@@ -1,8 +1,7 @@
-import os
+from pathlib import Path
+from typing import TYPE_CHECKING, Iterable, Iterator, Optional
 
-from typing import Optional, TYPE_CHECKING
-
-from .oxen import PyWorkspace, PyCommit
+from .oxen import PyCommit, PyWorkspace
 
 # Use TYPE_CHECKING for type hints to avoid runtime circular imports
 if TYPE_CHECKING:
@@ -55,7 +54,7 @@ class Workspace:
         workspace_id: Optional[str] = None,
         workspace_name: Optional[str] = None,
         path: Optional[str] = None,
-    ):
+    ) -> None:
         """
         Create a new Workspace.
 
@@ -86,9 +85,9 @@ class Workspace:
             print(
                 f"\033[91mMake sure that you have write access to `{repo.namespace}/{repo.name}`\033[0m\n"
             )
-            raise e
+            raise
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"Workspace(id={self._workspace.id()}, branch={self._workspace.branch()}, commit_id={self._workspace.commit_id()})"
 
     @property
@@ -113,7 +112,7 @@ class Workspace:
         return self._workspace.branch()
 
     @property
-    def commit_id(self):
+    def commit_id(self) -> str:
         """
         Get the commit id of the workspace.
         """
@@ -126,7 +125,7 @@ class Workspace:
         """
         return self._repo
 
-    def status(self, path: str = ""):
+    def status(self, path: str = "") -> None:
         """
         Get the status of the workspace.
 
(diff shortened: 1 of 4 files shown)


malcolmgreaves:
@coderabbitai review

oxen-python/python/oxen/workspace.py
```diff
@@ -136,30 +136,49 @@ def status(self, path: str = ""):
...
+            if isinstance(src, str) or isinstance(src, Path):
+                raise ValueError(f"{src} is not a valid filepath nor directory.")
+            else:
+                raise ValueError(
+                    "No valid filepaths provided: adding nothing to a workspace is invalid."
+                )
+        elif len(paths) == 1:
+            self._workspace.add(paths[0], dst)
+        else:
+            self._workspace.add_many(paths, dst)
```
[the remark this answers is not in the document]
malcolmgreaves: Fixed this one.

malcolmgreaves:
Also addressed:
- `232-238: _filepaths_from can follow symlink loops indefinitely.` (fixed)
- `51-89: Minor: raise e → raise to preserve original traceback.` (implemented)

oxen-python/python/oxen/workspace.py
```diff
@@ -136,30 +136,62 @@ def status(self, path: str = ""):
...
+                for fpath in _filepaths_from(Path(path))
+            ]
+
+        if len(paths) == 0:
+            if isinstance(src, str) or isinstance(src, Path):
+                raise ValueError(f"{src} is not a valid filepath nor directory.")
+            else:
+                raise ValueError(
+                    "No valid filepaths provided: adding nothing to a workspace is invalid."
+                )
```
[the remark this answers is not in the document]
malcolmgreaves: Reverting the reassignment.

oxen-python/python/oxen/workspace.py
```diff
@@ -136,30 +135,62 @@ def status(self, path: str = ""):
...
+        elif len(paths) == 1:
+            self._workspace.add(paths[0], dst)
+        else:
+            self._workspace.add_many(paths, dst)
```
jcelliott: It might be cleaner to just remove the `PyWorkspace::add_many` method and just use the `add` method for both cases. Those implementations are identical except for wrapping the single file in a Vec. I believe the separation is just a holdover from when we had a separate implementation.

oxen-rust/src/lib/src/api/client/workspaces/files.rs
```diff
@@ -810,19 +810,37 @@ async fn p_upload_bytes_as_file(
...
 
+fn filename_from_path(p: &Path) -> Result<String, OxenError> {
+    if let Some(filename) = p.file_name() {
+        Ok(filename.to_string_lossy().to_string())
+    } else {
+        Err(OxenError::file_has_no_name(p))
+    }
+}
+
 pub async fn add_many(
```
jcelliott, line 813: We should just remove this `add_many` function. It uses the older multipart request which is much slower than the parallel chunked/batched transfer we have in `add`.

jcelliott (changes requested):
Two small things to address. I like the improved type specs and using pathlib though.

oxen-python/python/oxen/workspace.py
malcolmgreaves: Got it, will do. Do you want to have the remaining method be called just "add"?

oxen-rust/src/lib/src/api/client/workspaces/files.rs
malcolmgreaves: Will remove 🫡

malcolmgreaves: Put this into its own PR: https://github.com/Oxen-AI/Oxen/pull/266

oxen-python/python/oxen/workspace.py
malcolmgreaves: Updated!

jcelliott (approved):

PR #266, opened 2026-02-12
Remove client workspace add_many

malcolmgreaves:
Fix: `add_version_file` ignoring hashes & rm `add_many`, `get_status_and_add_file

**Replace `add_many` with `add`**
Remove client workspace's `add_many` function: replaced all uses of
`api::client::workspaces::files::add_many` with
`api::client::workspaces::files::add`, unifying upload behavior.

**Bugfix: workspace add ignoring file hashes**
The workspace add function was ignoring file hashes. This is because staged
files were being completed (`/complete` endpoint) with `add_version_file(s)`,
which was using `get_status_and_add_file` incorrectly. It was hardcoding the
merkle tree node input in its `get_status_and_add_file` call, which
and this function was ignoring the file contents when determining add status.

Instead, it was hardcoding the merkle tree node in the `get_status_and_add_file`
call to `None` (_which is supposed to be used when there's not a pervious version_).
This meant that it never got a file hash to compare against and thus incorrectly
thought it was a new file.

`get_status_and_add_file` and `add_version_file` have been removed. They are
not needed, because they were only used in one server route:
`oxen_server::controllers::versions::chunks::complete`. And this route already
has the hash of the file contents (parameter `version_id`), which means it
can use `add_version_file_with_hash`.

And, since there's only one way to add a file (with a hash), this function
has been renamed to replace `add_version_file`. This is now complimentary
with `add_version_files` (which also already expects a file hash).

**Breaking Change**
- Deleted `liboxen::api::client::workspaces::files::add_many`
- Deleted `liboxen::core::v_latest::add::get_status_and_add_file`
- Deleted `liboxen::core::v_latest::workspaces::files::add_version_file`
- Renamed `liboxen::core::v_latest::workspaces::files::add_version_file_with_hash` to `add_version_file`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ logs/
 **/.claude/settings.local.json
 *.pyc
 .superset
+*.swp
--- a/oxen-rust/src/lib/src/api/client/entries.rs
+++ b/oxen-rust/src/lib/src/api/client/entries.rs
@@ -99,11 +99,12 @@ pub async fn upload_entries(
         api::client::workspaces::create(remote_repo, &branch_name, &workspace_id).await?;
     assert_eq!(workspace.id, workspace_id);
 
-    api::client::workspaces::files::add_many(
+    api::client::workspaces::files::add(
         remote_repo,
         &workspace_id,
         &opts.dst.to_string_lossy(),
         file_paths,
+        &None,
     )
     .await?;
 
--- a/oxen-rust/src/lib/src/api/client/workspaces/commits.rs
+++ b/oxen-rust/src/lib/src/api/client/workspaces/commits.rs
@@ -92,11 +92,12 @@ mod tests {
                 test::test_img_file_with_name("cole_anthony.jpeg"),
             ];
             api::client::workspaces::create(&remote_repo, &branch_name, &workspace_id).await?;
-            let result = api::client::workspaces::files::add_many(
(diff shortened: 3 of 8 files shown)


malcolmgreaves:
I like the suggestion of "41-47: Consider Option<&LocalRepository> instead of &Option<LocalRepository>." ~I'll add it!~ I'll move this to its own PR.

.superset/config.json
```diff
@@ -0,0 +1,6 @@
+{
```
jcelliott: Did you mean to commit this file?

malcolmgreaves: whoops! I did not -- let me add this to the .gitignore 😅

jcelliott (approved):

malcolmgreaves:
@jcelliott significant changes since your last review -- I tracked down the root cause of the failing test. It was `add_version_file` using `get_status_and_add_file`, which ignored the file hash and would always treat the file's status as new, leading to it being considered `Added` when it came to doing a workspace commit.

I saw that `add_version_file` was only used in `fn complete`. And `complete` already has the file hash as `version_id`. So I deleted `add_version_file` and `get_status_and_add_file` and instead use `add_version_file_with_hash`.

And since there's only one way to add a versioned file, I renamed `add_version_file_with_hash` to `add_version_file` & updated uses of it.

PR #270, opened 2026-02-13
Rust dependency caching in all GH action builds

malcolmgreaves:
Adds Rust dependency caching to publish / release workflows on all platforms (windows, linux, mac os).
Ensures that all uses have the same configuration.


malcolmgreaves:
I think the release builds should use a different cache key than the CI builds. They're in `--release` mode so the compiled binaries won't be the same.

PR #271, opened 2026-02-13
Fix: Windows: server staging keys + oxen tests

malcolmgreaves:
**Fix: Path separator mismatch in staged DB keys**
PathBuf::join uses `\` on Windows, so uploaded file keys get stored as `path\to\file` in
RocksDB. Downloads resolve via URL paths (always `/`), causing lookup misses. Fixed by
normalizing all keys, replacing `\` with `/`: implemented as `normalize_key()` and updated
all key accesses. Fixes `test_download_uploaded_file_from_workspace`.

**Test fix: CRLF line endings in test CSV**
Git's `core.autocrlf` converts `\n` to `\r\n` on Windows checkout. The test in 
`test_csv_video_captions_quoting` was comparing over lines, causing a mismatch. Fix is
to use `cfg!(windows)` to swap-out the newline sequence.

**Test fix: File locks from unclosed DB connections**
On Windows, open file handles prevent the file from being removed. The test cleanup 
function  was still holding on to the DuckDB and RocksDB connections, which was causing
the code  32 error ("...cannot access the file because it is being used by another process.").

Added `fn remove_from_cache_with_children` to `dir_hashes_db.rs` and updated both 
`fn cleanup_sync_dir` (server) and `fn maybe_cleanup_repo` (lib) to flush all 4 caches.
Fixes 7 tests in `controllers::workspaces::data_frames::tests::*`.

### DIFF
--- a/oxen-rust/src/lib/src/core/db/dir_hashes/dir_hashes_db.rs
+++ b/oxen-rust/src/lib/src/core/db/dir_hashes/dir_hashes_db.rs
@@ -35,6 +35,28 @@ pub fn dir_hash_db_path_from_commit_id(
         .join(DIR_HASHES_DIR)
 }
 
+/// Removes all dir_hashes DB instances from cache whose path starts with the given prefix.
+/// Used in test cleanup to release file handles before directory deletion.
+pub fn remove_from_cache_with_children(db_path_prefix: impl AsRef<Path>) -> Result<(), OxenError> {
+    let db_path_prefix = db_path_prefix.as_ref();
+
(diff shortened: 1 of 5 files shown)


malcolmgreaves:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform line-ending and path handling for consistent behavior across operating systems.

* **Tests**
  * Enhanced cache cleanup during test execution to ensure proper resource management and prevent test interference.

* **Chores**
  * Optimized internal cache management and eviction strategies for better stability.

jcelliott (approved):

PR #282, opened 2026-02-18
New Test PyPI publish workflow for any commit

malcolmgreaves:
Adds a new Test PyPI publishing workflow that allows building and publishing
`oxen-python` wheels to Test PyPI (https://test.pypi.org/) for any single commit.

The published version is a _development_ version. If the version is `0.5.0` then
the dev version is `0.5.0.devN` where `N` is the iteration number. `N` starts at 0
and is incremented for each change on a version. The last published version dev
version on Test PyPI is used to calculate the incremented `N`.

This workflow makes use of two new scripts:
- `patch_dev_version.sh`: updates `oxen-python`'s versions to use the `.devN` suffix
- `publish_test_pypi.sh`: builds `oxen-python` and publishes it to Test PyPI

PR #283, opened 2026-02-18
Workspace add files preserving relative paths

malcolmgreaves:
Adds a new API client function for workspaces, `add_files`, which adds files without
modifying their relative paths within an Oxen repository. The existing `add` function
will move files into a directory on add.

This new function is appropriate when one is trying to import an existing set of files
into an oxen repo. E.g. migrating some existing data repository into oxen.

Adds a new PyWorkspace method, `add_files_in_repository`, that calls this new
`add_files` workspace client function.

New tests show that `add_files` preserves relative paths in the repository.

### DIFF
--- a/oxen-python/python/oxen/workspace.py
+++ b/oxen-python/python/oxen/workspace.py
@@ -125,7 +125,7 @@ class Workspace:
         """
         return self._repo
 
-    def status(self, path: str = "") -> None:
+    def status(self, path: str = ""):
         """
         Get the status of the workspace.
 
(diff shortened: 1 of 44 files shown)


malcolmgreaves:
@jcelliott hold off on merging -- I'm getting this out here and I want to try https://github.com/Oxen-AI/Oxen/pull/282 on it before merging.

oxen-python/python/oxen/workspace.py
```diff
@@ -187,6 +187,59 @@ def stringify_ok(fpath: str | Path) -> str:
                 )
         self._workspace.add(paths, dst)
 
+    def add_files_in_repository(
+        self,
+        repo_dir: str | Path,
```
jcelliott: I think we should call this `base_dir` because the import location won't necessarily be a repository.

oxen-python/python/oxen/workspace.py
```diff
@@ -187,6 +187,59 @@ def stringify_ok(fpath: str | Path) -> str:
...
+        if not (repo_dir / ".oxen").is_dir():
+            raise ValueError(f"repo_dir is not a valid Oxen repository: {repo_dir}")
+
+        resolved: list[str] = []
+        for p in paths:
+            p = Path(p)
+            if not p.is_absolute():
+                p = repo_dir / p
+            else:
+                resolved_repo = repo_dir.resolve()
```
jcelliott: We should just resolve this once rather than on each iteration.

jcelliott (approved):

malcolmgreaves:
@coderabbitai review

PR #292, opened 2026-02-20
Build liboxen once in test PyPI workflow

malcolmgreaves:
Refactors the Test PyPI publish workflow so that `liboxen` is
only built once per-platform (OS + architecture). The crate is
reused for each individual pyO3 + Python version build.


malcolmgreaves:
Inspecting changes here: https://github.com/Oxen-AI/Oxen/actions/runs/22242233517

PR #304, opened 2026-02-27
Re-enable Coderabbit AI summary in PR description

malcolmgreaves:
Removed comments and adjusted settings for PR summary generation.


jcelliott:
@malcolmgreaves I'm actually liking the way you already configured it so I'm just going to close this one. I think we should update the other repos to match this instead of the other way around. And I can update my changelog script to grab the summary from the comment instead of the description.

PR #306, opened 2026-03-02
add codeowners file

EloyMartinez:
## Release Notes

* **Chores**
  * Updated repository maintenance configuration.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @Oxen-AI/rust-learners


CleanCut:
What's the intent, here?

The change here will add rust-learners as a reviewer for any change to the entire repo, whether it's about Rust or not. Did you mean to specify the Rust codebase in the pattern?

I don't have access to the Settings panel. Have we enabled the setting that will make this require a review from someone in that group? That's typically what CODEOWNERS is used for.

If folks just want to get a notification for any PR that happens, it would be better to have them watch the repository like this:

image

malcolmgreaves:
@EloyMartinez I agree with @CleanCut here. CODEOWNERS is usually paired with a repository setting to require reviews from the mentioned individual (or team) whenever a PR includes a change to a file that fits a listed pattern.

We don't have that setting enabled now, so there wouldn't be a change to the PR process.

But I think the better fit here is to watch for changes and view them, not have rust-learners be an owner of code :)

PR #307, opened 2026-03-03
Some onboarding fixes

CleanCut:
I ran through the readme and tried onboarding to the project that way. This PR fixes a bunch of sharp edges I encountered:

- The readme needed some significant updating
- The setup, running, and teardown of local Rust tests was a tedius manual cycle, so I wrote a `test-rust` script to handle it.

Fixes ENG-696


oxen-rust/README.md
```diff
@@ -303,13 +349,13 @@ curl -H "Authorization: Bearer $TOKEN" -X POST -d '{"name": "MyRepo"}' "http://$
 Create the docker image
 
 ```bash
-docker build -t oxen/server:0.6.0 .
+docker build -t oxen/server:0.44.1 .
```
jcelliott, line 314: Probably makes sense to just leave a `{version}` placeholder here, unless we also update the bump version script to update this automatically.

oxen-rust/README.md
```diff
@@ -91,23 +97,49 @@ Then create `.cargo/config.toml` in your Oxen repo root with the following:
 
 ```toml
 [target.aarch64-apple-darwin]
-rustflags = [ "-C", "link-arg=-fuse-ld=/opt/homebrew/opt/llvm/bin/ld64.lld"]
+rustflags = [ "-C", "link-arg=-fuse-ld=/opt/homebrew/opt/llvm/bin/ld64.lld", ]
+```
+
+### macOS with Intel
+
+On macOS Intel you can use the MIT-licensed version of mold, [sold](https://github.com/bluewhalesystems/sold)).
```
malcolmgreaves: I believe this documentation is no longer applicable. If you're changing it around, it would be good to remove it.

oxen-rust/README.md
```diff
@@ -20,13 +20,29 @@ The documentation for the Oxen.ai tool chain can be found [here](https://docs.ox
...
+
 # 🔨 Build & Run
 
 ## Install Dependencies
 
-Oxen is purely written in Rust 🦀. Refer to the [shared Rust toolchain installation instructions](../README.md#build-).
+### Rust
 
-If you are a developer and want to learn more about adding code or the overall architecture [start here](docs/dev/AddLibraryCode.md). Otherwise, a quick start to make sure everything is working follows.
+Oxen is purely written in Rust 🦀. You should install the Rust toolchain with rustup: https://www.rust-lang.org/tools/install.
```
malcolmgreaves: Please revert this change. The point is to direct everyone to a single set of instructions for set up. There's shared rust setup between `oxen-rust` and `oxen-python` and this was recently intentionally consolidated.

oxen-rust/README.md
```diff
@@ -20,13 +20,29 @@ The documentation for the Oxen.ai tool chain can be found [here](https://docs.ox
   - [ ] Backblaze
 - [ ] Block level deduplication
 
+# Library Architecture
+
+If you are a developer and want to learn more about adding code or the overall architecture [start here](docs/dev/AddLibraryCode.md). Otherwise, a quick start to make sure everything is working follows.
```
malcolmgreaves: `docs/dev/AddLibraryCode.md` was recently removed. Since you're touching this in this PR, please remove the mention to this file.

oxen-rust/README.md
```diff
@@ -212,65 +250,73 @@ docker load -i result
...
-cargo test -- --test-threads=$(nproc)
+scripts/test-rust
 ```
 
 It can be faster (in terms of compilation and runtime) to run a specific test. To run a specific library test:
 
 ```bash
-cargo nextest run --lib test_get_metadata_text_readme
+scripts/test-rust --lib test_get_metadata_text_readme
 ```
```
malcolmgreaves: Let's please keep the instructions focused on using `cargo nextest` for all of the tests. I believe there was an intentional switch some time ago to prefer this test runner because there are a few flakey tests and `nextest` will automatically re-run them.

oxen-rust/README.md
```diff
@@ -212,65 +250,73 @@ docker load -i result
...
+## Pre-Commit Hook
+
+We use [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) to check for commit consistency.
+make sure to install [`pre-commit`](https://pre-commit.com/) library
+and then install the pre-commit-hooks locally using:
 
 ```bash
-env OXEN_TEST_HOST=0.0.0.0:4000 cargo nextest run
+pre-commit install
 ```
```
malcolmgreaves: Please revert. This is redundant as the instructions for installing the pre-commit hooks already exists and we'd like to not have multiple repeated instructions: https://github.com/Oxen-AI/Oxen/blob/main/README.md#pre-commit-hooks

oxen-rust/src/server/.env.template
malcolmgreaves, line 1: Could you please provide some more context around this change? I'm just curious as to why this file is being renamed.

oxen-rust/README.md
```diff
@@ -212,65 +250,73 @@ docker load -i result
...
 ```bash
 ulimit -n 10240
 ```
 
-Run tests with nextest (preferred):
+Then you can run the tests with the `cargo test` or `cargo nextest` (preferred) directly. To run all tests with the default number of threads:
+
 ```bash
-cargo nextest run
+cargo test -- --test-threads=$(getconf _NPROCESSORS_ONLN)
```
malcolmgreaves: I'm unfamiliar with `getconf` and `nproc`. Is there a particular reason to switch from `nproc` to `getconf` here?

oxen-rust/README.md
malcolmgreaves: I believe that this works to get the right version number:
```
$(cargo pkgid | cut -f2 -d'@')
```

oxen-rust/scripts/test-rust
```diff
@@ -0,0 +1,72 @@
...
+
+# ---------- 4. Configure test user ----------
+echo "==> Configuring test user..."
+mkdir -p ./data/test/runs
+mkdir -p ./data/test/config
+
+./target/debug/oxen-server add-user \
+    --email ox@oxen.ai \
+    --name Ox \
+    --output data/test/config/user_config.toml
```
malcolmgreaves: This only needs to run once.

malcolmgreaves (commented):
Thanks for taking a crack at updating the onboarding documentation. I have a few things I'd like to see changed, particularly around having redundant documentation that's already listed in the top-level README and linked in this README.

oxen-rust/README.md
CleanCut: ✅ I looked into `scripts/bump-version.sh` and discovered that the workspace update PR had broken it, so since it needed to be fixed anyway I also made it update the readme.

oxen-rust/README.md
CleanCut: ✅ Removed

oxen-rust/README.md
CleanCut: ✅ Done

oxen-rust/README.md
CleanCut: `nproc` isn't on macOS, so the existing command only worked on Linux. `getconf` is available on both. I tested the `getconf` approach on Ubuntu and macOS, and it appears to work well.

Do we really need to provide the vanilla, threaded alternative, though? We said to use nextest which does this by default already. We could just not provide this example at all. Thoughts?

oxen-rust/scripts/test-rust
CleanCut: ✅ Updated to only run if one of the data/test/{runs,config} directories are not present.

oxen-rust/src/server/.env.template
CleanCut: `.env.template` was not helpful. It's not used and it didn't actually contain a helpful template. In fact, the only reason I discovered the file was that my attempt to run things crashed because there was no `env.local` file in the same directory, so when I went to the directory to add an `.env.local` I saw it. Running stuff locally requires a `.env.local` file to exist. So I put it in place and commented-out (and provided a comment for) the one item you can customize. Now you don't have to know to copy the file, it's just there and works, and you can uncomment `SYNC_DIR` and customize it if you want to.

oxen-rust/README.md
CleanCut: ✅ 🧹

oxen-rust/README.md
CleanCut: ✅ I added a link to the prerequisites section of the main readme, compared the content to the main readme and removed all of the sections that seemed redundant, and moved the cmake section to the main readme since it was missing there.

oxen-rust/src/server/.env.template
jcelliott: Honestly, I didn't even know this existed. I'm not sure why we're reading from `.env.local` in the first place. It doesn't make sense for it to be hard coded to look in `src/server` (https://github.com/Oxen-AI/Oxen/blob/main/oxen-rust/src/server/src/main.rs#L375). IMO this is much better handled by environment managers like `mise`. I would be fine leaving the default `dotenv` handler to optionally read a `.env` file in the cwd.

CleanCut:
Cherry-picked the CI fix from #308

oxen-rust/README.md
malcolmgreaves: Got it, thanks! Makes sense to use `getconf` then.

malcolmgreaves: > Do we really need to provide the vanilla, threaded alternative, though? 

Ah, that's a great question. I can see an argument for being thorough and documenting how to do things with "vanilla" tools. But, arguably, if that's not how we want to do it as a project, we can make that decision.

Maybe to be a bit on the conservative side with changing the core documentation, we can do it in phases. This PR can de-emphasize using `cargo test` and emphasize using `cargo nextest` and the new script you've made.

Ultimately though, after thinking about this a bit more, I think it's just important that we document the commands we use to run tests and point them to the helper script you wrote. Sometimes e.g. I just keep the server going and iterate on client code, but it is nice to have a single command to run to ensure the server is up and running. So feel free to edit and change this if you feel strongly about just having the docs talk about `nextest` instead of `cargo test` w/ `test-threads`.

.pre-commit-config.yaml
```diff
@@ -34,13 +34,6 @@ repos:
         files: ^oxen-python/.*\.rs$
         pass_filenames: false
 
-      - id: cargo-sort
```
malcolmgreaves, line 37: Could you actually please add this back? 

I apologize, because I added this and `cargo-sort` just had a version update that changed its sorting format. I didn't have versions pinned, so this causing your PR to be marked as failing by CI is in part my fault 😬 

I have since fixed that here https://github.com/Oxen-AI/Oxen/commit/077af58a9fbd91566eca1dd1af82d8dc792431e7, so you can rebase and you won't have this problem.

And I'm fixing it more thoroughly with https://github.com/Oxen-AI/Oxen/pull/312

oxen-rust/src/server/.env.template
malcolmgreaves: Ah ok, that makes a lot of sense as to why you changed this! Thanks for the context.

> I'm not sure why we're reading from .env.local in the first place.
> I would be fine leaving the default dotenv handler to optionally read a .env file in the cwd.

I agree, I can't find a good reason why it should fail if this file isn't present. (The env could already be set up!)

oxen-rust/README.md
malcolmgreaves: Another thought -- we have talked about merging `oxen-{rust,python}` into the same workspace, and at that point I wonder if we will need to adjust the documentation and strategy here too. We don't use `cargo nextest` on the `oxen-python` tests. We may want to opt for something like a `justfile` and have a set of commands we can run so something like "run me all the tests" is `just test` and we control the implementation. (And e.g. "run specific tests" could be `just test <testname1> <testname2>`).

oxen-rust/scripts/test-rust
```diff
@@ -0,0 +1,76 @@
...
+echo "==> Starting oxen-server..."
+./target/debug/oxen-server start &
+SERVER_PID=$!
+
+# Give the server a moment to start
+sleep 2
+if ! kill -0 "$SERVER_PID" 2>/dev/null; then
+    echo "ERROR: oxen-server failed to start."
+    exit 1
+fi
```
malcolmgreaves: I had to poke around for it, but we actually have a health check on `oxen-server`!

This should let you be able to make sure the server is up and running:
```suggestion
wait_for_server() {
  for i in $(seq 1 30); do
    if curl -s -o /dev/null --fail http://localhost:3001/api/health 2>/dev/null; then
      return 0
    fi
    sleep 0.5
  done
  return 1
}

if ! wait_for_server || ! kill -0 "$SERVER_PID" 2>/dev/null; then
  echo "ERROR: oxen-server failed to start."
fi
```

malcolmgreaves (approved):
LGTM with one change -- please keep the cargo-sort pre-commit hook. Two more suggestions but they're non-blocking.

oxen-rust/src/server/.env.template
CleanCut: Sounds like a a good plan. I created [a linear issue](https://linear.app/oxenai/issue/ENG-710/clean-up-env-file-behavior-in-oxen-server) to track following up on this.

.pre-commit-config.yaml
CleanCut: What if we didn't re-add it to a PR's CI as a blocker? What if it was just a local check?

I'm concerned that if cargo-sort is still releasing breaking changes then we are going to be forced to frequently update it locally or risk periodically being blocked in CI by changes. 

I'd really like to move forward and deploy. My two PRs have been blocked on this since yesterday afternoon. I've lost four CI cycles on this (discovery & fix on two PRs), and reverting requires another CI run _and_ feels like it introduces risk to hitting this again.

Maybe this won't feel so risky if I get CI times down. Currently some of my CI cycles take over an hour(!!!).
I do like the sorted Cargo.toml files...

oxen-rust/scripts/test-rust
CleanCut: 😎

.pre-commit-config.yaml
CleanCut: In the meantime, I've cherry-picked the formatting fix and reverted as requested.

malcolmgreaves: #312 pins the version of cargo-sort so this problem won't occur again @CleanCut

PR #308, opened 2026-03-03
Rust 1.93.1 and 2024 Edition

CleanCut:
This updates to Rust 1.93.1 and the 2024 edition for all Rust code in the repository.

I don't think it's necessary to review every single change in this update, it's an extensive, mechanical update.

 - The first commit is the actual version updates.
 - The second commit has a ton of no-op fixes.
   - `rustfmt` was updated significantly between 2021 and 2024. The whitespace changes (more vertical chains) and the module import name sorting change (upper-case now comes first) are the big changes that affected tons of files.
   - `clippy` - Most of the clippy changes were related to introducing "[let chains](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains)".
   - There are a few changes where we were using `ref varname` in a pattern that was matching a reference, which is no longer allowed.

Fixes ENG-695


malcolmgreaves (approved):

CleanCut:
The `cargo sort` check has failed on [this PR](https://github.com/Oxen-AI/Oxen/actions/runs/22683836261/job/65761295532) and on [one of Surya's PRs](https://github.com/Oxen-AI/Oxen/actions/runs/22684166984/job/65762526402?pr=311), despite it passing locally. Apparently it gets different results on Linux vs macOS. 😕

PR #311, opened 2026-03-04
help message in case of push failure.

subygan:


oxen-rust/src/lib/src/core/v_latest/push.rs
```diff
@@ -321,6 +321,11 @@ async fn get_commit_missing_hashes(
             )?
             else {
                 log::error!("push_commits commit node not found for commit: {commit}");
+                eprintln!(
```
CleanCut, line 324: For my own education: Where do the messages for `log::error!` vs `eprintln!` end up being routed to?

malcolmgreaves: stderr! :) (AFAIK we don't have file loggers or the like configured at the moment)

oxen-rust/src/lib/src/core/v_latest/push.rs
```diff
@@ -499,6 +508,20 @@ async fn push_commits(
     Ok(())
 }
 
+fn wrap_push_entry_error(entry_type: &str, err: OxenError) -> OxenError {
+    let err_str = err.to_string();
+    if err_str.contains("No such file or directory") || err_str.contains("NotFound") {
```
malcolmgreaves, line 513: Hmmmm, this strikes me as a bit odd. It's looking at the stringified representation of the error to try and guess what it's root cause is? This seems a bit indirect. I'd prefer it if we can be direct here. 

Can we structure this so that we are using discrete error variants that represent the actual root cause instead?

malcolmgreaves: For instance, I believe what this is doing is trying to take a few different error conditions, all of which crop up during a push, and tries to give them a better human-readable error message. Is that right?

If that is the intent, then a clearer way to indicate it is to push up the structured error message to the CLI and then _there_ explicitly `match` on these variants and convert that into a string message.

oxen-rust/src/lib/src/api/client/tree.rs
```diff
@@ -81,6 +81,13 @@ pub async fn create_nodes(
         progress.set_message(format!("Packing {}/{} nodes", i + 1, nodes.len()));
 
         log::debug!("create_nodes appending dir to tar");
+        if !node_dir.exists() {
+            return Err(OxenError::basic_str(format!(
+                "Node {node_hash} not found locally. \
+                Not all nodes are present locally. \
+                Run `oxen pull <remote> <branch> --all` to fetch the complete history before pushing."
+            )));
```
malcolmgreaves, line 89: I notice that this string is repeated almost verbatim 3 times in this PR. There's also a lot of structured information in this string that'd be better served if it was a `struct` instead. This looks like it'd be a good candidate to turn into a new error variant.

Perhaps something like:
```rust
pub enum OxenError {
  ...
  NodesNotPresent{node_hash: &str}
  ...
}
```

The `impl std::fmt::Display` is where we can turn this into a human-readable error message:
```rust
match error {
   ...
  NodesNotPresent{node_hash} => write!(f, "Node {node_hash} not found locally. \ ..."),
  ...
}
```

When the CLI runs, it can:
```
match <expr> {
  Ok(x) => // do what it wanted to do with x
  Err(e) => {
    eprintln!("{e}") // this will be the message
    exit
}
```

oxen-rust/src/lib/src/core/v_latest/push.rs
malcolmgreaves: If, looking into this, it becomes hard to do because it turns out that there's just too many uses of `OxenError::basic_str`, then I'd actually say that we should press pause and go and fix those up.

The overuse of `basic_str` has been on my radar for a little while now. We can do a better job about being specific and structured with our error responses. If we take a moment to address the underlying issue, things like having helpful error messages to users will be actually easy to do. Because there will be one and only one place where we make those error messages.

malcolmgreaves (commented):
I like the new error message! I think we can improve the mechanism by which we create these errors and surface the user-facing message though. More details in my comments.

oxen-rust/src/lib/src/core/v_latest/push.rs
CleanCut: if these messages are both going to the stderr, we should standardize on one way to emit them.

oxen-rust/src/lib/src/api/client/tree.rs
subygan: yes. we don't have a specific mechanism to capture. errors. i'll create a different PR with a better error handling

PR #312, opened 2026-03-04
Dev pre-req install script

malcolmgreaves:
Adds a new script that installs all required development tools that are 
listed in the `Build` section of `README.md`. This script installs:
- the rust toolchain at the version used in `oxen-rust`
- `uv` for `oxen-python`
- installs all system packages in linux and checks that xcode is installed in macos
- installs all cargo tools
- installs pre-commit and installs the project's pre-commit hooks

All script actions are idempotent, so it's safe to run multiple times.

PR #314, opened 2026-03-05
Re-format all Cargo.tomls with updated cargo-sort version

malcolmgreaves:


malcolmgreaves:
Hotfix to unblock PRs. This https://github.com/Oxen-AI/Oxen/pull/312 will unify dev tool installation locally & in CI and enforce versions so we don't have this problem again.

PR #315, opened 2026-03-05
Update to Rust 1.94.0

CleanCut:
Super straightforward.

Resolves https://linear.app/oxenai/issue/ENG-711/update-to-rust-194


oxen-rust/scripts/test-rust
```diff
@@ -27,8 +27,7 @@ cargo build
 # ---------- 2. Install/Update nextest ----------
```
CleanCut, line 27: We can change this step to just call `scripts/install-pre-reqs.sh` in a later PR once #312 is merged.

malcolmgreaves (approved):

PR #319, opened 2026-03-06
Adds shellcheck pre-commit hook and fixes lint errors

malcolmgreaves:
- Adds `shellcheck` as a pre-commit hook. 
- Integrates it into the pre-req install script and pins the version.
- Ran `shellcheck` on all bash scripts and fixed all true issues.
    + Suppress false-positives that were manually verified.
- Ensures that all scripts use `env` to locate the correct `bash` interpreter.
- Appends `.sh` to `oxen-rust/scripts/test-rust` so that it is picked up by the pre-commit hook.


oxen-rust/scripts/backup_and_migrate_repo.sh
```diff
@@ -23,35 +23,25 @@ else
...
   exit 1
 fi
 
 # Step 3: Verify that the tarball has been uploaded to s3
-aws s3 ls "s3://$BUCKET_NAME/$FILEPATH.tar.gz"
-if [ $? -ne 0 ]; then
+if ! aws s3 ls "s3://$BUCKET_NAME/$FILEPATH.tar.gz"; then
   echo "Verification failed, tarball not found in S3"
   exit 1
 fi
```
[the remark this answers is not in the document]
malcolmgreaves: OOOOOOF! Great catch!

oxen-rust/scripts/index_repos.sh
```diff
@@ -15,17 +17,18 @@ then
...
-    cd $repo
-    oxen config --set-remote origin https://$SYNC_SERVER/ox/$repo
-    oxen push origin main
-    cd ..
+    oxen clone "https://staging.hub.oxen.ai/ox/$repo" --all
+    (
+      cd "$repo" || exit
+      oxen config --set-remote origin "https://$SYNC_SERVER/ox/$repo"
+      oxen push origin main
+    )
```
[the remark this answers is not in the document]
malcolmgreaves: Added `set -euo pipefail`

scripts/install-pre-reqs.sh
```diff
@@ -226,7 +226,57 @@ else
...
+        *)       echo "Unsupported architecture: $arch"; exit 1 ;;
+    esac
+
+    local url="https://github.com/koalaman/shellcheck/releases/download/v${version}/shellcheck-v${version}.${sc_os}.${sc_arch}.tar.xz"
+    local tmp_dir
+    tmp_dir="$(mktemp -d)"
+    info "Downloading shellcheck v${version} from ${url}..."
+    curl -fsSL "$url" | tar -xJ -C "$tmp_dir"
+    sudo install -m 755 "$tmp_dir/shellcheck-v${version}/shellcheck" /usr/local/bin/shellcheck
+    rm -rf "$tmp_dir"
```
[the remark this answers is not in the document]
malcolmgreaves: Incorrect

scripts/bump-version.sh
```diff
@@ -95,26 +95,22 @@ echo "" >&2
...
-  else
-    sed -i 's/server:.* /server:'"$VERSION"' /' "$file"
-  fi
-  echo "  ✓ $file" >&2
-done
+if [[ "$(uname)" == "Darwin" ]]; then
+  sed -i '' 's/server:.* /server:'"$VERSION"' /' oxen-rust/README.md
+else
+  sed -i 's/server:.* /server:'"$VERSION"' /' oxen-rust/README.md
+fi
```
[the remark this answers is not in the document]
malcolmgreaves: Fixed

oxen-rust/scripts/test-rust.sh
```diff
@@ -75,7 +75,7 @@ fi
 echo "==> oxen-server running (pid $SERVER_PID)."
 
 # ---------- 6. Run tests ----------
-echo "==> Running tests with 'cargo nextest run $@' ..."
+echo "==> Running tests with 'cargo nextest run $*' ..."
```
CleanCut, line 106: That's incorrect.

```suggestion
echo "==> Running tests with 'cargo nextest run $@' ..."
```

oxen-rust/scripts/backup_and_migrate_all_repos.sh
```diff
@@ -29,9 +28,7 @@ for namespace in "$ABSOLUTE_ROOT_PATH"/*; do
 
           # Make the script exectuable
           chmod +x "$DIR/backup_and_migrate_repo.sh"
-          "$DIR/backup_and_migrate_repo.sh" "$repository" "$namespace_name/$repository_name" "$MIGRATION_NAME"
-
-          if [ $? -ne 0 ]; then
+          if ! "$DIR/backup_and_migrate_repo.sh" "$repository" "$namespace_name/$repository_name" "$MIGRATION_NAME"; then
             echo "Backup and migration failed for $repository"
             # exit 1
```
CleanCut: We can delete this dead code, and AI will stop worrying about it. It makes sense to try migrating the rest of the repos, even if one fails.

```suggestion
```

oxen-rust/scripts/index_repos.sh
```diff
@@ -15,17 +18,18 @@ then
...
```
CleanCut: Consider reverting this. Adding two parens and indenting the code to execute a subshell seems significantly more complicated than just doing `cd ..`. I can't find any advantage to this approach. If I missed something, I'd love to learn what it is.

oxen-rust/install-ubuntu.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
CleanCut: ~For this line to be useful, we need to also make the script executable and commit the executable bit.~

⚠️ I can't find mention of `install-ubuntu.sh` _anywhere_ in the `Oxen-AI` GitHub org, across all repositories. Can this file be deleted? Is it used somewhere else that I'm missing? cc @jcelliott @gschoeni

oxen-rust/scripts/test-rust.sh
```diff
@@ -60,7 +60,7 @@ SERVER_PID=$!
```
CleanCut, line 87: **Discussion:** Let's not add `.sh` to scripts.

- It's not necessary to use the filename to indicate the language an executable is implemented in. (We don't rename our other executables `oxen.rust` after they're created!)
- This is already in a directory literally named `scripts`
- IDEs and shellcheck ignore the file extension except as a last resort, relying on the shebang line instead
- None of these scripts are written in actual `sh`.
- If we're going to change extensions, I'd rather remove the `.sh` extensions from the rest of the scripts in the repo and make sure they're all located in a `scripts` directory, or similar.
- The pre-commit shellcheck config supports scanning the `scripts` directories. I tested it, and shellcheck does the right thing. It found and checked all the shell scripts and ignored python scripts. See comment on `‎.pre-commit-config.yaml‎` for more details.

Strong opinion, weakly held. Just a pet peeve. If y'all just really like the extensions, let me know and I'll deal with it. Just wanted to have the conversation once, in case y'all are amenable to omitting the extensions on executables. cc @jcelliott

scripts/install-pre-reqs.sh
```diff
@@ -226,7 +226,57 @@ else
 fi
 
 ###############################################################################
-# 5. pre-commit (installed as a uv tool)
+# 5. shellcheck (shell script linter)
+###############################################################################
+
+install_shellcheck() {
```
CleanCut, line 232: We should use `brew` to manage shellcheck on macOS.

.pre-commit-config.yaml
```diff
@@ -68,3 +68,10 @@ repos:
         language: system
         files: ^scripts/.*\.py$
         pass_filenames: false
+
+      - id: shellcheck
+        name: shellcheck
+        entry: shellcheck -x
+        language: system
+        types: [shell]
+        files: \.sh$
```
CleanCut, line 77: I tested checking all the files in the `scripts` directories, and it works great. `shellcheck` ignores any files that aren't shell scripts (like our python scripts), and checks all the scripts based on their shebang line.

This does require either moving `oxen-rust/install-ubuntu.sh` into one of the `scripts` directories or deleting it.

```suggestion
        files: scripts/.*$
```

CleanCut (commented):
👍🏻 `set -euo pipefail` is so nice.

oxen-rust/scripts/test-rust.sh
malcolmgreaves: The pre-commit hook in this PR is set up so that it runs `shellcheck` on every script that exists in the repository. The rest of the scripts in the repository have the convention of ending in `.sh`. The new script you made last week does not. I'd prefer to keep the existing convention and not introduce changes to the project at this time.

oxen-rust/scripts/test-rust.sh
malcolmgreaves: This is a suggestion from shellcheck.

Could you please provide more explanation -- what makes you think this is incorrect?

I believe `shellcheck`'s default rule is set up to encourage using the single-string variant of the script's args when using it in a string (or it may be when using it with `echo` so that one doesn't accidentally get into a situation where string arguments are accidentally split when it isn't desired.

oxen-rust/scripts/index_repos.sh
malcolmgreaves: I'll revert. This was a pattern shellcheck suggested, but it only made sense when there wasn't `set -e`.  IIRC what it was doing was making the `cd` temporary because it was executing it in a subshell. If the `oxen ...` commands failed, then it would not cause the script's working directory to change.

scripts/install-pre-reqs.sh
malcolmgreaves: Changed to use `brew` and `apt`

malcolmgreaves: @CleanCut reverting back to installing binaries from github releases directly. Unfortunately we can't install a specific version using brew and apt: https://github.com/Oxen-AI/Oxen/pull/319#discussion_r2913759111

oxen-rust/scripts/index_repos.sh
malcolmgreaves: @CleanCut coderabbit figured out that `cd $repo; ...; cd ..` can fail if `$repo` is nested: https://github.com/Oxen-AI/Oxen/pull/319#discussion_r2913759101

A benefit of doing this in `( ... )` is that it avoids this pitfall.

However, I'll remove the `(..)` and use `pushd / popd` to correctly move into and move back.

scripts/install-pre-reqs.sh
```diff
@@ -226,7 +226,48 @@ else
...
+        brew unlink shellcheck &>/dev/null || true
+        brew install "shellcheck@${version}"
+        brew link --overwrite "shellcheck@${version}"
+    elif [ "$PLATFORM" = "linux" ]; then
+        info "Installing shellcheck ${version} via apt..."
+        sudo apt-get update -y
+        # Use --allow-downgrades to force the pinned version even if a newer one is installed
+        sudo apt-get install -y --allow-downgrades "shellcheck=${version}*"
+    fi
+}
```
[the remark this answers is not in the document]
malcolmgreaves: reverting

oxen-rust/scripts/test-rust.sh
CleanCut: Oh, I intended for a literal `$@` to be output, not for it to be substituted. A single string substitution with `$*` is more useful than a literal `$@`, so let's go ahead and leave this.

oxen-rust/scripts/index_repos.sh
CleanCut: AFAIK, we don't support nested repos in the repo-with-owner identifier, so that shouldn't be possible. `pushd / popd` is always good. 👍🏻

scripts/install-pre-reqs.sh
CleanCut: Well, stink.

oxen-rust/install-ubuntu.sh
jcelliott: Yes, let's just delete this. I'm pretty sure it was from way back when we were deploying oxen-server directly on an EC2 instance.

oxen-rust/scripts/test-rust.sh
jcelliott: Conceptually I agree with executables not needing an extension, and we should just consistently add shebang lines. But I believe at least some of these scripts are used elsewhere and don't want to break CI/tests. So for consistency let's keep the extension for now. If we want to convert them all later let's do that as a separate PR.

CleanCut: Acknowledged!

CleanCut (approved):

PR #322, opened 2026-03-09
update PUT file to allow full destination path in URL

gschoeni:
# Description

The API for multipart PUT was not quite working as expected when you had nested target path, it was just taking the file name and not the actual path.

# Repro setup: create a real temp file for upload
echo "# Hero Content" > /tmp/hero.md

# 1) Single-file full-path upload (this is what was broken before, now works)
curl -X PUT "http://localhost:3000/api/repos/ox/test-put-upload/file/main/pages/home/hero.md" \
  -F "file=@/tmp/hero.md;filename=hero.md" \
  -F "message=add hero" \
  -F "name=ox" \
  -F "email=ox@test.local"

# 2) Verify exact same path
curl "http://localhost:3000/api/repos/ox/test-put-upload/file/main/pages/home/hero.md"

# 3) Directory upload semantics still work via files[]
curl -X PUT "http://localhost:3000/api/repos/ox/test-put-upload/file/main/pages/home" \
  -F "files[]=@/tmp/hero.md;filename=hero.md" \
  -F "message=add hero in dir mode" \
  -F "name=ox" \
  -F "email=ox@test.local"

# 4) Verify file landed under directory resource + filename
curl "http://localhost:3000/api/repos/ox/test-put-upload/file/main/pages/home/hero.md"

### DIFF
--- a/oxen-rust/src/lib/src/api/client/file.rs
+++ b/oxen-rust/src/lib/src/api/client/file.rs
@@ -69,7 +69,7 @@ async fn put_multipart_file(
     let form = apply_commit_body(Form::new().part(field_name, file_part), commit_body);
     let res = client.put(&url).multipart(form).send().await?;
     let body = client::parse_json_body(&url, res).await?;
-    serde_json::from_str(&body).map_err(Into::into)
+    Ok(serde_json::from_str(&body)?)
 }
 
 async fn make_file_part(
--- a/oxen-rust/src/lib/src/api/client/file.rs
+++ b/oxen-rust/src/lib/src/api/client/file.rs
@@ -20,32 +20,15 @@ pub async fn put_file(
 ) -> Result<CommitResponse, OxenError> {
     let branch = branch.as_ref();
     let directory = directory.as_ref();
-    let file_path = file_path.as_ref();
-    let uri = format!("/file/{branch}/{directory}");
-    log::debug!("put_file {uri:?}, file_path {file_path:?}");
-    let url = api::endpoint::url_from_repo(remote_repo, &uri)?;
-
-    let client = client::new_for_url(&url)?;
-    let file_part = Part::file(file_path).await?;
-    let file_part = if let Some(file_name) = file_name {
-        file_part.file_name(file_name.as_ref().to_string())
-    } else {
-        file_part
-    };
-    let mut form = Form::new().part("files[]", file_part);
-
-    if let Some(body) = commit_body {
-        form = form.text("name", body.author);
-        form = form.text("email", body.email);
-        form = form.text("message", body.message);
-    }
-
-    let req = client.put(&url).multipart(form);
-
-    let res = req.send().await?;
-    let body = client::parse_json_body(&url, res).await?;
-    let response: CommitResponse = serde_json::from_str(&body)?;
-    Ok(response)
+    put_multipart_file(
+        remote_repo,
+        format!("/file/{branch}/{directory}"),
+        "files[]",
+        file_path,
+        file_name,
+        commit_body,
+    )
+    .await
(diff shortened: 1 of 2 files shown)


oxen-rust/src/lib/src/api/client/file.rs
```diff
@@ -20,32 +20,76 @@ pub async fn put_file(
 ) -> Result<CommitResponse, OxenError> {
     let branch = branch.as_ref();
     let directory = directory.as_ref();
+    put_multipart_file(
+        remote_repo,
+        format!("/file/{branch}/{directory}"),
+        "files[]",
```
CleanCut, line 26: **Question:** Why do we use an array of files here in put_file (`files[]`), but a single file below in `put_file_to_path` (`file`)?

oxen-rust/src/lib/src/api/client/file.rs
```diff
@@ -20,32 +20,76 @@ pub async fn put_file(
...
+    log::debug!("put_multipart_file {uri:?}, file_path {file_path:?}");
     let url = api::endpoint::url_from_repo(remote_repo, &uri)?;
-
     let client = client::new_for_url(&url)?;
+
+    let file_part = make_file_part(file_path, file_name).await?;
+    let form = apply_commit_body(Form::new().part(field_name, file_part), commit_body);
+    let res = client.put(&url).multipart(form).send().await?;
+    let body = client::parse_json_body(&url, res).await?;
+    serde_json::from_str(&body).map_err(Into::into)
```
CleanCut: **Nit:** It's a bit more idiomatic to manage the error conversion with `?`.

```suggestion
    Ok(serde_json::from_str(&body)?)
```

oxen-rust/src/server/src/controllers/file.rs
```diff
@@ -48,6 +48,14 @@ pub struct FileUploadBody {
     pub email: Option<String>,
 }
 
+type ParsedMultipartRepoFields = (
+    Option<String>,
+    Option<String>,
+    Option<String>,
+    Vec<TempFileNew>,
+    Vec<TempFileNew>,
+);
```
CleanCut: Manually parsing the multipart form data is leading to some awkward duplication between the `FileUploadBody` struct that is only used to generate documentation and the `parse_multipart_fields_for_repo` function which actually does the parsing and returns it via the `ParsedMultipartRepoFields` type alias containing the information that the above-mentioned struct is supposed to have.

I originally wrote an essay about how we could fix this, but it got ridiculously long so [I pushed a stacked PR with an implementation](https://github.com/Oxen-AI/Oxen/pull/330) instead. Take a look and let me know what you think. 😄

oxen-rust/src/server/src/controllers/file.rs
```diff
@@ -251,39 +259,51 @@ pub async fn put(
...
 
-    let mut files: Vec<FileNew> = vec![];
-    for temp_file in temp_files {
-        files.push(FileNew {
-            path: temp_file.path,
-            contents: temp_file.contents,
-            user: user.clone(), // Clone the user for each file
-        });
+    for file in &files {
+        ensure_no_file_ancestors_in_tree(&repo, &commit, &file.path, &resource.path)?;
```
CleanCut, line 347: **Question:** Do we need to check _all_ the ancestors? This is `O(n^2)`, as we need to iterate over every ancestor for every file. Could we get away with just checking the immediate ancestor? That would only be `O(n)`.

CleanCut (approved):
This can be merged as-is, though I left some questions and suggestions to consider.

I also authored a PR as one of my suggestions. 😄 (see comment below)

PR #324, opened 2026-03-09
Switch to env-logger so that oxen-python respects RUST_LOG

malcolmgreaves:
`oxen-python` now uses `env-logger` so that the `RUST_LOG` env var can now properly configure logging in `liboxen`. This also avoids the GIL deadlock issue that can arise when using `pyo3-log` incorrectly. Removes the `pyo3-log` dependency.


crates/oxen-py/src/lib.rs
```diff
@@ -41,20 +41,9 @@ fn oxen(m: Bound<'_, PyModule>) -> PyResult<()> {
...
-    // shouldn't be calling back to python to need to grab the GIL again,
-    // and can just perform all the heavy work in rust.
-
-    // println! also works to debug, make sure to remove it before committing
-
-    // https://docs.rs/pyo3-log/latest/pyo3_log/#interaction-with-python-gil
-    // pyo3_log::init();
+    // Initialize env_logger so that RUST_LOG is respected.
+    // We use env_logger instead of pyo3_log to avoid GIL deadlocks.
+    let _ = env_logger::try_init();
```
CleanCut, line 47: **Question:** Is there any reasonable action we can take if the logger fails to initialize? (I can't think of any)

CleanCut (approved):
🤞🏻

crates/oxen-py/src/lib.rs
malcolmgreaves: Not that I can think of off the top of my head. But FWIW I'm pretty sure the `try_init` is there so it doesn't panic if something else has already set up the logger. At this point, there should be nothing else, but it's a defensive init.

PR #326, opened 2026-03-09
Workspace::{add,add_files} return list of failed-to-upload paths

malcolmgreaves:
When `api::client::workspaces::files::{add,add_files}` fails to upload the contents at some path due to a server-side error, that path is now returned in an `Ok(vec[..])` instead of causing the function to return `Err(...)`. Callers will now be responsible for checking that the returned vec of paths is empty.

This change is made so that large file uploads to `oxen-server` do not fail when a single path fails in a large batch.

The Python `Workspace` class's `add` and `add_files` methods have been updated to return a list of the failed-to-upload paths now.

Additionally, now both `add` and `add_files` will return an `Err` if attempting to upload a file or directory that does not exist locally. This shouldn't be an upload failure, because there was nothing to upload when something was requested to be uploaded.


malcolmgreaves:
@coderabbitai re-review

jcelliott (approved):

PR #336, opened 2026-03-13
Remove AsRef generics to reduce compilation time & binary size

malcolmgreaves:
Replaces all generic `AsRef<X>` uses with `&X` direct references.

Includes all instances of a generic parameter of the form `Y: AsRef<X>` and
`impl AsRef<X>`. Notably, this included many `AsRef<str>` and `AsRef<Path>`,
which are now replaced with `&str` and `&Path`, respectively.

This cuts down on compilation time and binary size considerably. Since
Rust monomorphizes generics: it must compile new variants of each function
for every different concrete type use. Unlike other traits, the codebase's
use of `AsRef` doesn't provide much utility as we nearly never swap-out
different things that could be turned into the reference type. The code is
not only smaller and faster to compile, it is clearer in what, exactly,
it requires and what it exactly operates on.

### DIFF
--- a/crates/cli/src/cmd/branch.rs
+++ b/crates/cli/src/cmd/branch.rs
@@ -196,8 +196,8 @@ impl BranchCmd {
     ) -> Result<(), OxenError> {
         let (scheme, host) = get_scheme_and_host_from_repo(repo)?;
 
-        check_remote_version_blocking(scheme.clone(), host.clone()).await?;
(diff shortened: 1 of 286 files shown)


malcolmgreaves:
NOTE: This is not ready for review. I'm not marking it as a draft so that I can see the effect on compile time in CI.

PR #343, opened 2026-03-16
Bump timeout and retry requests

CleanCut:
Rework #325 from scratch to include only essential changes.

- I made it explicit in `Cargo.toml` that we are using reqwest 0.13.2, since this PR uses a retry API introduced in 0.12.23.
- Bump the default reqwest timeout that we set on some clients from `2m` to `10m`. On the rest of the clients we don't set any timeout.
- Bump reqwest's retries from the default `2` to `3`. Applies to all reqwest clients built in `crates/lib/src/api/client.rs`
- Configure reqwest to retry on all application-level errors, not just low-level network errors. We may need to tighten this up to only retry specific application-level errors if this causes problems. Applies to all reqwest clients built in `crates/lib/src/api/client.rs`
- Exponential backoff is used by default in all reqwest retries, so we didn't need to configure that

### DIFF
--- a/oxen-rust/crates/lib/src/constants.rs
+++ b/oxen-rust/crates/lib/src/constants.rs
@@ -193,7 +193,7 @@ pub const NUM_HTTP_RETRIES: u64 = 5;
 /// Number of workers
 pub const DEFAULT_NUM_WORKERS: usize = 8;
 /// Default timeout for HTTP requests
-pub const DEFAULT_TIMEOUT_SECS: u64 = 120;
+pub const DEFAULT_TIMEOUT_SECS: u64 = 600;
 /// Default vnode size
 pub const DEFAULT_VNODE_SIZE: u64 = 10_000;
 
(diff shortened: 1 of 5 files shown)


oxen-rust/crates/lib/src/constants.rs
```diff
@@ -193,7 +193,7 @@ pub const NUM_HTTP_RETRIES: u64 = 5;
 /// Number of workers
 pub const DEFAULT_NUM_WORKERS: usize = 8;
 /// Default timeout for HTTP requests
-pub const DEFAULT_TIMEOUT_SECS: u64 = 120;
+pub const DEFAULT_TIMEOUT_SECS: u64 = 600;
```
CleanCut, line 196: This is the timeout that is used by liboxen::api::client::new_for_host(), which is called by `new_for_url`, `new_for_url_no_user_agent`, and `new_for_remote_repo`.

The other client helper methods don't add a timeout.

jcelliott (approved):

PR #345, opened 2026-03-17
HTTP Client TCP Keep Alives & Explicit Hostname Type

malcolmgreaves:
Second attempt at PR https://github.com/Oxen-AI/Oxen/pull/325

Fixes critical bug in #325 -- the root cause of that was improperly setting the auth token to the
right hostname. When using a remote repository to create the HTTP client, #325 was incorrectly
getting the full URL and using that. Instead, it should have been getting only the hostname.

This PR removes the confusion that led to the bug. The hostname is no longer typed as a String.
Instead, it is represented by a new `struct Hostname`, which has explicit `host`, `port`, and 
`scheme` fields as well as a `.hostname()` method that gets the correct `host` and, if non-none,
appends the `port` to it.


malcolmgreaves:
This works with auth:

```
$ oxen-dev push
Warning: 🐂 Oxen remote version mismatch.

CLI Version: 0.46.3
Server Version: 0.46.2

Please visit https://docs.oxen.ai/getting-started/install for installation instructions.

🐂 oxen push origin main -> adb01f32ce28f4dec76125baf26818a8
🐂 push complete 🎉 took 324ms
```

This is for a remote repository in the hub:
```
$ cat .oxen/config.toml
remote_name = "origin"
min_version = "0.36.0"

[[remotes]]
name = "origin"
url = "https://hub.oxen.ai/malcolm-oxen/deleteme"

[storage]
type = "local"

[storage.settings]
path = ".oxen/versions/files"
```

PR #346, opened 2026-03-17
Install the prereqs for ffmpeg on macOS

CleanCut:
This installs the prereqs you need to build oxen server with the `-F ffmpeg` feature on macOS.

Note that (so far) actually enabling the above feature is left to an exercise for the reader.


gschoeni (approved):
LGTM

PR #348, opened 2026-03-17
Remove unhelpful default server settings

CleanCut:
### `keep_alive`

Actix's [keep_alive](https://github.com/Oxen-AI/Oxen/compare/cleancut/fix-server-settings?expand=1#diff-d37c11ba1bce3d86b20c038fb390a9b538c5df7aa92362f0e3c1a8e668cb09ccL512) setting sets the frequency that keep-alive packets are sent to keep a connection open if it doesn't have any traffic actively transiting it. We were decreasing the frequency from the default every 5 seconds to every 10 _minutes_. A setting this infrequent effectively disables server-side keep-alives, which can lead to unnecessarily dropped connections in transactions.

When I tested uploading 10 GB of files, I noticed that the upload took ~2 minutes, and then the server took ~1 minute to process the uploads and send a response.  I suspect there was no traffic happening in that last ~1 minute, which is where this setting could have caused problems if the client were also not sending keep-alives. Many corporate firewalls and other network infrastructure will terminate traffic-less connections. The most aggressive I've had to deal with in the past was set to 15 seconds, but 1-5 minutes is more common. `<cough>`_Hotels_`</cough>` 😂 

### `client_request_timeout`

Actix's [cient_request_timeout](https://github.com/Oxen-AI/Oxen/compare/cleancut/fix-server-settings?expand=1#diff-d37c11ba1bce3d86b20c038fb390a9b538c5df7aa92362f0e3c1a8e668cb09ccL513) setting configures the timeout for how long we will wait for a client to send their _headers_. This timeout is to prevent [DOS attacks](https://en.wikipedia.org/wiki/Denial-of-service_attack) from attackers who open a connection, start sending headers, but never finish sending headers, consuming a connection. The default is 5 seconds, which is a sufficiently small window to prevent most attackers from being able to exhaust resources before connections timeout. We set it to 10 _minutes_, which makes us trivially easy to DOS.

### DIFF
--- a/oxen-rust/crates/server/src/main.rs
+++ b/oxen-rust/crates/server/src/main.rs
@@ -24,7 +24,6 @@ pub(crate) mod test;
 extern crate log;
 extern crate lru;
 
-use actix_web::http::KeepAlive;
 use actix_web::middleware::{Condition, DefaultHeaders, Logger};
 use actix_web::{App, HttpServer, web};
 use actix_web_httpauth::middleware::HttpAuthentication;
@@ -74,7 +73,6 @@ use clap::{Parser, Subcommand};
 
 use std::env;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 
 const VERSION: &str = liboxen::constants::OXEN_VERSION;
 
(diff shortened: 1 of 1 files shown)


jcelliott (approved):
This looks correct to me. We shouldn't have set those values before.

https://docs.rs/actix-web/4.9.0/actix_web/struct.HttpServer.html#method.keep_alive
https://docs.rs/actix-web/4.9.0/actix_web/struct.HttpServer.html#method.client_request_timeout

PR #350, opened 2026-03-18
fix tabular functions that panic'd in async tasks

gschoeni:
I noticed that `oxen df data.jsonl --randomize` was causing the tool to panic

Screenshot 2026-03-18 at 2 51 47 PM

PR #351, opened 2026-03-18
Fix df {randomize, rename-col, vstack}

CleanCut:
Intended as an alternative approach to https://github.com/Oxen-AI/Oxen/pull/350

- Remove some task blocking (rather than doubling down on mixing sync/async code with blocking)
- fix randomize: Enable the polars "random" feature and use it directly
- fix rename-col: Use the upstream "rename" feature directly
- fix vstack: Use the upstream "concat" feature directly
- tests for the three fixes (randomize and rename-col tests copy-n-pasted from #350, added the vstack test which that PR didn't address)
- tests for add-col and add-row, even though they weren't broken, copy-n-pasted from #350, since they look like decent tests and there was no existing coverage I saw in the file

Closes #350

### DIFF
--- a/oxen-python/Cargo.lock
+++ b/oxen-python/Cargo.lock
@@ -4841,6 +4841,8 @@ dependencies = [
  "polars-json",
  "polars-schema",
  "polars-utils",
+ "rand 0.8.5",
+ "rand_distr",
  "rayon",
  "regex",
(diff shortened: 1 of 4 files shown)


gschoeni (approved):
LGTM!

PR #352, opened 2026-03-19
Improve pre-commit CI output

CleanCut:
- Enable colors in pre-commit in CI
- Output `git status` in color if pre-commit fails in CI


jcelliott (approved):

CleanCut:
Much better!

image

PR #357, opened 2026-03-19
Try enabling sccache

CleanCut:
- Try to get our CI working much more quickly by using sccache, which compiles individual compilation artifacts.
- Still use `Swatinem/cache-rust` to the cargo registry and any Rust tools installed with `cargo install`
- Now that we're caching individual build artifacts and the cargo registry separately, I'm disabling only writing to the cache on `main`, since I suspect we'll have enough space. If cache space still ends up being a problem, we should be able to switch at least sccache to use S3 (or maybe everything)


.github/workflows/ci_test.yml
```diff
@@ -76,18 +78,24 @@ jobs:
...
+      - name: Setup sccache
+        uses: mozilla-actions/sccache-action@v0.0.7
+
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v2
         with:
           # Only cache builds from main. Otherwise our caches will be too big and start to churn. (10 GB limit in GitHub Actions)
           # Other branches will still use the cache but won't save it.
           save-if: ${{ github.ref == 'refs/heads/main' }}
+          cache-on-failure: true
```
jcelliott: What's the reasoning behind caching on failure?

CleanCut: I'm glad you asked! I forgot to add the `cache-targets: false` line, which helps this make more sense.

The idea is to use `Swatinem/cache-rust` to cache just the cargo registry downloads, and use sccache for the compilation artifacts (basically, the `target` directory). So under that setup, even if compilation fails, the registry downloads are still correct.

I pushed up a commit to fix that, to clarify better in comments, and to update the version of the sccache action.

CleanCut: And also enable the `cache-all-crates` option that it claims is good to do if you install tooling, and we certainly do install tooling.

CleanCut:
Closing the sccache experiment. I can't find any way to use sccache (with or without Swatinem/cache-rust) that speeds up any of our test suite builds.

PR #358, opened 2026-03-19
Fix: error propagation on server error

subygan:
Download failures during pull/clone/checkout were silently ignored. 
Workers would log::error! and continue, the caller would get Ok(()), and the user would see a successful pull with missing files.

  - Client: Use handle_non_json_response for consistent HTTP error handling in entries.rs and versions.rs; collect and surface errors from chunked downloads instead of discarding them or storing them as the file data.
  - Workers: Propagate errors from tokio::spawn worker loops in fetch.rs. previously these logged errors and returned Ok(()) which


oxen-rust/crates/lib/src/api/client/entries.rs
```diff
@@ -612,41 +625,35 @@ pub async fn download_large_entry(
...
         })
+        .buffer_unordered(num_workers)
+        .collect()
         .await;
 
+    // Propagate the first error before recombining chunks
+    for result in &results {
+        if let Err(err) = result {
+            // Clean up temp dir on failure
+            util::fs::remove_dir_all(&tmp_dir)?;
```
CleanCut: The rabbit's correct on this one. We can ignore any potential cleanup errors.

```suggestion
            let _ = util::fs::remove_dir_all(&tmp_dir);
```

crates/server/src/controllers/entries.rs
```diff
@@ -118,11 +119,24 @@ pub async fn download_chunk(
...
     let chunk_size: u64 = query.chunk_size.unwrap_or(AVG_CHUNK_SIZE);
 
-    let file_node = repositories::entries::get_file(&repo, &commit, &path)?
-        .ok_or(OxenError::path_does_not_exist(path.clone()))?;
-    let chunk = version_store
+    let file_node = match repositories::entries::get_file(&repo, &commit, &path)? {
+        Some(node) => node,
+        None => {
+            let err: OxenHttpError = OxenError::path_does_not_exist(path).into();
+            return Ok(err.error_response());
```
CleanCut: **Question:** So...we want to return an HTTP OK, saying that the path doesn't exist? Shouldn't we return a 404?

crates/server/src/controllers/entries.rs
```diff
@@ -118,11 +120,24 @@ pub async fn download_chunk(
...
```
CleanCut: I dug into this. This change is a no-op. In both cases we start with an `OxenError` and converting it to an `OxenHttpError` loses the `OxenError` variant about the path not existing and just converts it to an `OxenHttpError::InternalOxenError` which then gets converted to a 500 Internal Server http response.

What we should be doing here is returning a 404 Not Found http response. This is a `/chunk/{resource}` endpoint call to retrieve a specific chunk. If we can't find the chunk that they ask for, we should tell the client that we can't find it. The pragmatic fix is to just construct the HttpResponse here directly instead of relying on two error conversions.

```suggestion
            return Ok(HttpResponse::NotFound().finish());
```

In the future I'd like to clean up the whole `OxenError -> OxenHttpError -> HttpResponse` conversion flow. It looks like we do a lot of work to specify specific `OxenError` variants, which are all are indiscriminantly converted to 500 Server Errors.

crates/lib/src/api/client/entries.rs
```diff
@@ -413,22 +413,20 @@ pub async fn pull_large_entry(
...
         .await;
 
+    // Propagate the first error before combining chunks
+    for result in &results {
+        if let Err(err) = result {
+            return Err(OxenError::basic_str(format!(
+                "Failed to download chunk for {remote_path:?}: {err}"
+            )));
+        }
+    }
```
CleanCut: I think this specific spot looks good, but if you look at the only caller of this "pull_large_entry" function, the error is _ignored_. So if the goal is to propagate the error back outside of the oxen code, we'll need to go fix that layer as well.

crates/lib/src/api/client/entries.rs
```diff
@@ -612,41 +625,35 @@ pub async fn download_large_entry(
...
+    // Propagate the first error before recombining chunks
+    for result in &results {
+        if let Err(err) = result {
+            // Clean up temp dir on failure
+            let _ = util::fs::remove_dir_all(&tmp_dir);
+            return Err(OxenError::basic_str(format!(
+                "Failed to download chunk for {remote_path:?}: {err}"
+            )));
+        }
+    }
```
CleanCut: This code also looks fine, but _one_ of the two places that calls the `download_large_entry` function ignores the error, so that layer needs to be fixed as well if you want the errors to propagate outward.

crates/server/src/controllers/entries.rs
CleanCut: I investigated deeper--see my other comments.

CleanCut:
In general, it would be super helpful if, in the top comment of the PR, you would describe the problem you are addressing and why/how you expect this change resolves it. I'm not sure I'm guessing correctly! 😄

crates/server/src/controllers/entries.rs
```diff
@@ -118,8 +118,13 @@ pub async fn download_chunk(
...
     let chunk_size: u64 = query.chunk_size.unwrap_or(AVG_CHUNK_SIZE);
 
-    let file_node = repositories::entries::get_file(&repo, &commit, &path)?
-        .ok_or(OxenError::path_does_not_exist(path.clone()))?;
+    let file_node = match repositories::entries::get_file(&repo, &commit, &path)? {
+        Some(node) => node,
+        None => {
+            return Err(OxenHttpError::NotFound);
+        }
+    };
```
CleanCut, line 126: (Non-blocking) The match statement here is a manual (re)implementation of `Option::ok_or`:

```suggestion
    let file_node = repositories::entries::get_file(&repo, &commit, &path)?.ok_or(OxenHttpError::NotFound)?;
```

CleanCut (approved):
Excellent improvement to propagation. Good streamlining as well. It's always a good sign when there is a net-reduction in code.

PR #363, opened 2026-03-23
Rename scripts -> bin

CleanCut:
This PR is purely renaming. There should be no changed functionality.

- Remove unnecessary extensions from bash script files
- Switch to the unix-style (singular) ~`script`~ `bin` directory name


jcelliott:
If we're already changing this, would it make more sense to just call the directory `bin` rather than `script`?

CleanCut:
> If we're already changing this, would it make more sense to just call the directory `bin` rather than `script`?

😄 Great idea! 

✅ Made the change and pushed it up.

jcelliott (approved):

PR #364, opened 2026-03-24
Import endpoint accepts arbitrary URLs

jcelliott:
The core change here is removing the restriction that the import endpoint can only fetch data from 3 whitelisted domains. Now we can fetch from any arbitrary domain. The bulk of the changes are just dealing with the security implications of that (mainly SSRF). 

There is an [API change](https://github.com/Oxen-AI/Oxen/pull/364/changes#diff-92cff25551dee549465e18daac0ed7d584ae4d2c6e00d6aadb58ebce3895c2deR84) for the import endpoint (not currently used anywhere AFAIK) to make it more consistent with the other endpoints (moving some parameters from headers to JSON body fields). It also replaces the generic `headers` field with an `auth` field. It was misleading before because `Authorization` was the only header we were actually sending.


CHaiz15 (approved):
Nice! Code looks good to my eyes, although i'm probably not the most authorized reviewer here 😅

crates/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -190,32 +193,173 @@ pub fn exists(workspace: &Workspace, path: impl AsRef<Path>) -> Result<bool, Oxe
...
+    // 192.0.0.0/24 — IETF protocol assignments (RFC 6890)
+    if octets[0] == 192 && octets[1] == 0 && octets[2] == 0 {
+        return true;
+    }
+    // 198.18.0.0/15 — Benchmarking (RFC 2544)
+    if octets[0] == 198 && (octets[1] & 0xFE) == 18 {
+        return true;
+    }
+    false
+}
```
CleanCut, line 265: Concern: This is a less complete (inverse) version of the standard library's [IpAddr::is_global()](https://doc.rust-lang.org/beta/std/net/enum.IpAddr.html#method.is_global) method. We should use the standard library's implementation.
  

```suggestion
```

crates/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -190,32 +193,173 @@ pub fn exists(workspace: &Workspace, path: impl AsRef<Path>) -> Result<bool, Oxe
...
+        .ok_or_else(|| OxenError::file_import_error("URL has no host"))?;
+    let port = url.port_or_known_default().unwrap_or(443);
+    let addr = format!("{host}:{port}");
+
+    let resolved = tokio::net::lookup_host(&addr).await.map_err(|e| {
+        OxenError::file_import_error(format!("DNS resolution failed for {host}: {e}"))
+    })?;
+
+    for socket_addr in resolved {
+        if is_private_ip(&socket_addr.ip()) {
```
CleanCut, line 282:
```suggestion
        if !socket_addr.ip().is_global()) {
```

crates/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -271,26 +415,90 @@ pub async fn upload_zip(
...
+                "HTTP request failed with status {status}"
+            )));
+        }
+
+        response = Some(resp);
+        break;
+    }
+
+    let response = response.ok_or_else(|| {
+        OxenError::file_import_error("Failed to get a successful response after redirects")
```
CleanCut: **Nit:** We don't know whether or not there have been redirects. We could store that info and customize the message if we care about it.

```suggestion
        OxenError::file_import_error("Failed to get a successful response")
```

crates/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -300,11 +508,33 @@ async fn fetch_file(
...
+    } else if let Some(cd) = resp_headers
+        .get("content-disposition")
+        .and_then(|h| h.to_str().ok())
+    {
+        parse_content_disposition_filename(cd)
+            .or_else(|| filename_from_url(&current_url))
+            .unwrap_or_else(|| uuid::Uuid::new_v4().to_string())
+    } else {
+        filename_from_url(&current_url).unwrap_or_else(|| uuid::Uuid::new_v4().to_string())
+    };
```
CleanCut: **Nit:** This can be simplified:

```suggestion
    let raw_filename = caller_filename.clone().unwrap_or_else(|| { // caller specified
        resp_headers
            .get("content-disposition")
            .and_then(|h| h.to_str().ok())
            .and_then(parse_content_disposition_filename) // Content-Disposition
            .or_else(|| filename_from_url(&current_url)) // URL path
            .unwrap_or_else(|| uuid::Uuid::new_v4().to_string()) // UUID
    });
```

crates/lib/src/repositories/workspaces/files.rs
```diff
@@ -238,4 +238,74 @@ mod tests {
...
+            // Non-HTTP scheme
+            let result = workspaces::files::import(
+                "file:///etc/passwd",
+                "",
+                std::path::PathBuf::from("data"),
+                None,
+                &workspace,
+            )
+            .await;
+            assert!(result.is_err(), "should reject non-HTTP scheme");
```
CleanCut: **Nit:** This can be simplified to a test-case list + loop:

```suggestion
            let cases = [
                ("http://127.0.0.1/secret", "loopback address"),
                ("http://10.0.0.1/internal", "private address"),
                ("http://169.254.169.254/latest/meta-data/", "link-local/metadata address"),
                ("http://[::1]/secret", "IPv6 loopback"),
                ("file:///etc/passwd", "non-HTTP scheme"),
            ];

            for (url, label) in cases {
                let result = workspaces::files::import(
                    url,
                    "",
                    std::path::PathBuf::from("data"),
                    None,
                    &workspace,
                )
                .await;
                assert!(result.is_err(), "should reject {label}");
            }
```

crates/server/src/controllers/import.rs
```diff
@@ -486,15 +492,17 @@ mod tests {
         let uri = format!("/oxen/{namespace}/{repo_name}/file/import/main/data");
 
         // import a file from oxen for testing
-        let body = serde_json::json!({"download_url": "https://hub.oxen.ai/api/repos/datasets/GettingStarted/file/main/tables/cats_vs_dogs.tsv"});
+        let body = serde_json::json!({
+            "download_url": "https://hub.oxen.ai/api/repos/datasets/GettingStarted/file/main/tables/cats_vs_dogs.tsv",
+            "name": author,
+            "email": email,
+        });
```
CleanCut, line 499: **Concern:** (Existing situation--this could be addressed in a later PR) 🤔 This URL looks like it causes requests to oxen-server in production. If that's true, then if the hub or oxen-server in production are down, we fail CI for anything we're doing...including trying to get production back up.

We should probably stand up a simple test helper that listens on a port and returns the file content we want for the test, and point the import at that, and then tear that down at the end of the test.

crates/server/src/controllers/import.rs
```diff
@@ -544,15 +552,17 @@ mod tests {
         let uri = format!("/oxen/{namespace}/{repo_name}/file/import/main/notebooks");
 
         // import a file from oxen for testing
-        let body = serde_json::json!({"download_url": "https://hub.oxen.ai/api/repos/datasets/GettingStarted/file/main/notebooks/chat.py"});
+        let body = serde_json::json!({
+            "download_url": "https://hub.oxen.ai/api/repos/datasets/GettingStarted/file/main/notebooks/chat.py",
```
CleanCut, line 556: Ditto the concern about depending on production to be up to pass tests.

CleanCut (approved):
A couple concerns, and some optional suggestions.

All of the above could be addressed in a later PR if we want to go ahead and ship this one now.

crates/lib/src/core/v_latest/workspaces/files.rs
jcelliott: I looked at using `is_global`. Unfortunately, it's nightly-only and the [tracking issue](https://github.com/rust-lang/rust/issues/27709) has been open since 2015. Do you think we're missing any important cases here?

jcelliott:
@CleanCut applied some of your suggestions. Thanks!

crates/lib/src/core/v_latest/workspaces/files.rs
CleanCut: Dang, I forgot to check if it was nightly. That's lame this has stalled, because it looks great. No, as I skimmed it I only saw a few other edge cases. If we want to be thorough, we could copy over the nightly code and use it directly, but otherwise I'd just leave this as-is.

PR #370, opened 2026-03-24
Fixup the `mv` function

CleanCut:
While working on #366 I found that `mv` was returning a path that no one used, and behaving inconsistently when trying to move a file to itself. I fixed it so it returns `()` and always returns an error (early) if you try to move a file to itself.


jcelliott (commented):
This does change the semantics of the no-op move from Ok to an Error. That seems reasonable to me but there's a chance something downstream relies on that behavior.

jcelliott (approved):

PR #371, opened 2026-03-24
Don't unstage all parent trees

CleanCut:
In #366 I noticed that if we `restore` a dataframe row and that is the last change, instead of just unstaging the file the dataframe corresponds to we were unstaging...basically everything. I fixed it.

### DIFF
--- a/crates/lib/src/core/v_latest/workspaces/data_frames/rows.rs
+++ b/crates/lib/src/core/v_latest/workspaces/data_frames/rows.rs
@@ -80,16 +80,6 @@ pub async fn restore(
             &workspace.workspace_repo,
             &HashSet::from([path.as_ref().to_path_buf()]),
         )?;
-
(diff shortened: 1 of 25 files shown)


crates/lib/src/core/v_latest/workspaces/data_frames/rows.rs
```diff
@@ -80,16 +80,6 @@ pub async fn restore(
             &workspace.workspace_repo,
             &HashSet::from([path.as_ref().to_path_buf()]),
         )?;
-
-        // loop over parents and delete from staged db
```
jcelliott, line 84: We do need to traverse up the tree and clear any leftover parent nodes, but it looks like `remove_staged_recursively` already handles that, although that function is pretty hard to read...

jcelliott (approved):

PR #374, opened 2026-03-24
Return error when workspace file is staged to be removed

CleanCut:
Found during #366 

Return an error when trying to download a workspace file that is staged to be removed.


jcelliott:
@CleanCut When you found this, was it actually causing an error? I understand the motivation here, but I actually think the current behavior is more correct. When the node is staged to be removed in a workspace it's not actually removed yet. So while it's kind of weird to then try to get that file, I think it would be weirder to not be able to get it.

CleanCut:
Nope, wasn't causing an error. Just wasn't sure of the semantics.

PR #380, opened 2026-03-25
Api 159/better error handling in the client

subygan:
example,

Screenshot 2026-03-25 at 12 19 57 AM

this adds a unified cli_error.rs handler to convert OxenError into user readable error. so far we've been abusing oxenError::basic_str() and bespoke eprintln!() at various layers. this should allow us to bubble up the actual error message and handle it at a higher level.


crates/cli/src/cmd/tree.rs
```diff
@@ -115,12 +114,10 @@ impl TreeCmd {
...
-                repositories::tree::print_tree_depth_subtree(
-                    repo,
-                    commit,
-                    depth,
-                    subtrees.first().unwrap(),
-                )?;
+                let first = subtrees
+                    .first()
+                    .ok_or_else(|| OxenError::basic_str("No subtree paths configured"))?;
+                repositories::tree::print_tree_depth_subtree(repo, commit, depth, first)?;
```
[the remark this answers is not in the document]
subygan: not an issue right now.

crates/cli/src/cmd/download.rs
```diff
@@ -105,8 +105,6 @@ impl RunCmd for DownloadCmd {
             for path in paths {
                 repositories::download(&remote_repo, &path, &dst, &revision).await?;
             }
-        } else {
-            eprintln!("Repository does not exist {id}");
```
CleanCut, line 109: (Pre-existing problem) We should return an error here, not ignore the error case.

crates/cli/src/cmd/save.rs
```diff
@@ -41,8 +40,8 @@ impl RunCmd for SaveCmd {
 
         let output_path = Path::new(output_str);
         let repo_path = Path::new(repo_str);
-        let repo_dir =
-            util::fs::get_repo_root(repo_path).ok_or(OxenError::basic_str(error::NO_REPO_FOUND))?;
```
CleanCut, line 45: The constant `NO_REPO_FOUND` is no longer used and should be deleted.

crates/cli/src/cmd/upload.rs
```diff
@@ -130,8 +130,6 @@ impl RunCmd for UploadCmd {
             };
 
             repositories::workspaces::upload(&remote_repo, &opts).await?;
-        } else {
-            eprintln!("Repository does not exist {name}");
```
CleanCut, line 134: This should return an error as well.

crates/lib/src/api/client/oxen_version.rs
```diff
@@ -1,48 +1,29 @@
...
-    } else {
-        let err = format!("api::version::get_remote_version Err request failed: {url}");
-        Err(OxenError::basic_str(err))
-    }
+    let res = client.get(&url).send().await?;
+    log::debug!("get_remote_version got status: {}", res.status());
+    let body = client::parse_json_body(&url, res).await?;
+    log::debug!("get_remote_version got body: {body}");
+    let response: OxenVersionResponse = serde_json::from_str(&body)?;
+    Ok(response.version)
```
CleanCut: The [version endpoint](https://hub.oxen.ai/api/version) returns what appears to be a `StatusMessage`, not an `OxenVersionResponse`. The JSON doesn't have a `version` field, and that's not an optional field on `OxenVersionResponse`.

crates/lib/src/api/client/oxen_version.rs
```diff
@@ -1,48 +1,29 @@
...
-    } else {
-        let err = format!("api::version::get_min_oxen_version Err request failed: {url}");
-        Err(OxenError::basic_str(err))
-    }
+    let res = client.get(&url).send().await?;
+    log::debug!("get_min_oxen_version got status: {}", res.status());
+    let body = client::parse_json_body(&url, res).await?;
+    log::debug!("get_min_oxen_version got body: {body}");
+    let response: OxenVersionResponse = serde_json::from_str(&body)?;
+    Ok(response.version)
```
CleanCut, line 31: Note: Unlike the `version endpoint` in the previous comment, the [`min_version`](https://hub.oxen.ai/api/min_version) endpoint does seem to be an `OxenVersionResponse`. So this should be okay, but it's weirdly inconsistent. The `version` field seems to correctly be the min version, since it's a lower number than the `oxen_version`.

crates/lib/src/api/client/version.rs
```diff
@@ -1,49 +1,29 @@
...
-    } else {
-        let err = format!("api::version::get_remote_version Err request failed: {url}");
-        Err(OxenError::basic_str(err))
-    }
+    let res = client.get(&url).send().await?;
+    log::debug!("get_remote_version got status: {}", res.status());
+    let body = client::parse_json_body(&url, res).await?;
+    log::debug!("get_remote_version got body: {}", body);
+    let response: VersionResponse = serde_json::from_str(&body)?;
+    Ok(response.version)
```
CleanCut: 1. The version endpoint returns a `StatusMessage`, so this needs to be reverted back to that type.
2. `VersionResponse` is identical to `OxenVersionResponse` in all but name. `crates/lib/src/view/version.rs` should be deleted. (It's not even tied into the module trie in `crates/lib/src/view.rs`, so I don't know how this is even compiling! rust-analyzer complained about it.)

crates/lib/src/api/client/version.rs
```diff
@@ -1,49 +1,29 @@
...
-        }
-    } else {
-        let err = format!("api::version::get_min_oxen_version Err request failed: {url}");
-        Err(OxenError::basic_str(err))
-    }
+    let res = client.get(&url).send().await?;
+    log::debug!("get_min_oxen_version got status: {}", res.status());
+    let body = client::parse_json_body(&url, res).await?;
+    log::debug!("get_min_oxen_version got body: {}", body);
+    let response: VersionResponse = serde_json::from_str(&body)?;
```
CleanCut:
```suggestion
    let response: OxenVersionResponse = serde_json::from_str(&body)?;
```

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -51,7 +50,8 @@ impl LocalRepository {
     pub fn from_dir(path: impl AsRef<Path>) -> Result<Self, OxenError> {
         let path = path.as_ref().to_path_buf();
         let config_path = util::fs::config_filepath(&path);
-        let config = RepositoryConfig::from_file(&config_path)?;
+        let config = RepositoryConfig::from_file(&config_path)
+            .map_err(|_| OxenError::local_repo_not_found(&path))?;
```
CleanCut: The rabbit is right on this one, let the various possible errors through without converting them.


```suggestion
        let config = RepositoryConfig::from_file(&config_path)?;
```

crates/lib/src/error.rs
CleanCut, line 1: (Can be done later) The entire display implementation for `OxenError` should be transformed into thiserror `#[error]` annotations on the enum variants. The current code works, so unless you want to take this on now, I can catch this later when I'm breaking up liboxen into smaller crates.

Just to illustrate what I'm saying:

```rust
use thiserror::Error;

#[derive(Error, Debug)
pub enum OxenError {
    ...
    // this variant has a customized display
    #[error("Repository '{0}' not found")]
    RepoNotFound(Box<RepoNew>),

    // this variant doesn't have a customized display
    ForkStatusNotFound(StringError),
    ...
```

crates/cli/src/cli_error.rs
```diff
@@ -0,0 +1,140 @@
+use colored::Colorize;
+use liboxen::error::OxenError;
+
+/// Format an OxenError for CLI output.
+///
+/// This adds CLI-specific context on top of OxenError's Display impl:
+/// colored "Error:" prefix, actionable hints telling the user what command
+/// to run, and suppression of internal details for truly internal errors.
+pub fn format_error(err: &OxenError) -> String {
```
CleanCut: (Can be done later) Since this is only called in exactly this way:

```rust
eprintln!("{}", cli_error::format_error(&err));
```

We can refactor `format_error` to be `print_error` to print out each thing as we see it which will reduce the amount of macros to compile and also simplify the code. Essentially:

```rust
pub fn print_error(err: &OxenError) {
    eprintln!("{} {}", "Error:".red().bold(), err);
    match err {
        // for all the variants with hints
        OxenError::LocalRepoNotFound(_) => print_hint("Run `oxen init` ...");
        ...
        // and ignore the rest
        _ => {}
    }
}
```

Not necessary for this PR--I can take it on later if you want.

CleanCut (commented):
- ❤️ I like the focus on more `?`, less manual matching on the result!
- ❤️ I like the reduction in panics via `expect`!
- ⚠️ Some of my comments are optional, but I found a few actual bugs that should be fixed for sure or we'll break stuff.

crates/cli/src/cmd/download.rs
CleanCut: ✅ Double-checked

crates/cli/src/cmd/save.rs
CleanCut: ✅ Double-checked

crates/cli/src/cmd/upload.rs
CleanCut: ✅ Double-checked

crates/lib/src/api/client/oxen_version.rs
CleanCut: ✅ Double-checked

crates/lib/src/api/client/version.rs
CleanCut: ✅ Double-checked

crates/lib/src/api/client/version.rs
CleanCut: ✅ Double-checked

crates/lib/src/model/repository/local_repository.rs
CleanCut: ✅ Double-checked

crates/lib/src/error.rs
CleanCut: ✅ Nice! Thanks for making the change!

CleanCut (approved):
Nice! I verified you addressed all the feedback (and more!)

There were a few improvements that could be made to hints, but since it was faster to make the change than describe the change, I [stacked this PR for you to review](https://github.com/Oxen-AI/Oxen/pull/386). 😄 😉

PR #383, opened 2026-03-26
Api 159/remove option

subygan:
in [crates/lib/src/api/client/repositories.rs](https://github.com/Oxen-AI/Oxen/pull/383/changes#diff-1db2b179526adaa5c75446cdc149c18e76aa2e107bd6d3f779744d3f9ad7b1f6L114) file, 
we have a function that had
`get_by_remote(remote: &Remote) -> Result<Option<RemoteRepository>, OxenError> `

now we return,
`get_by_remote(remote: &Remote) -> Result<RemoteRepository, OxenError>`

because in most callers where this was being used, we were just doing `.ok_or(OxenError::remote_not_found(remote.clone()))?;` or returning a `OxenError::basic_str()` after checking if the Option is None

now we correctly return a `OxenError::RemoteRepoNotFound` if the repository is not found. this simplifies a lot of error handling and bubbles up the error up correctly.


crates/lib/src/api/client/branches.rs
```diff
@@ -162,22 +162,15 @@ pub async fn delete_remote(
...
-            } else {
-                Err(OxenError::remote_branch_not_found(branch_name))
-            }
-        } else {
-            Err(OxenError::remote_repo_not_found(&remote.url))
-        }
-    } else {
-        Err(OxenError::remote_not_set(remote))
-    }
+    let remote = repo
```
CleanCut, line 165: Nice simplification!

We can see here that `get_by_name` has the same problem that we're fixing in `get_by_remote`. It returns an awkward `Result<Option<Branch>, OxenError>`.

In some later PR we should change it to return `Result<Branch, OxenError>` with an error for the missing case.

crates/lib/src/api/client/repositories.rs
```diff
@@ -797,9 +791,7 @@ mod tests {
     async fn test_get_by_name() -> Result<(), OxenError> {
         test::run_empty_local_repo_test_async(|local_repo| async move {
             let remote_repo = test::create_remote_repo(&local_repo).await?;
-            let url_repo = api::client::repositories::get_by_remote_repo(&remote_repo)
-                .await?
-                .unwrap();
```
CleanCut, line 802: Nice elimination of a panic!

crates/lib/src/command/fetch.rs
CleanCut, line 1: 🧹 Nice dead code cleanup!

CleanCut (approved):
❤️ This looks great! And thank you for the excellent PR description!

PR #387, opened 2026-03-26
Remove clone cleanup on failure

subygan:
was facing an issue where because of an issue with clone the cloned repository was being cleaned up, for a 350GB repo, which takes ~45 minutes to clone, this will be a frustrating experience.

while git does this delete on clone failure. it doesn't make sense for us, where the clone time can be long. we also have added a lot of repository healing commands lik, `fsck`, `--missing-files`, `--revalidate`. so it is better to have a broken repo and then fix it as opposed to having to start a new clone

### DIFF
--- a/crates/lib/src/repositories/clone.rs
+++ b/crates/lib/src/repositories/clone.rs
@@ -5,6 +5,7 @@
 
 use std::path::Path;
 
+use crate::api;
 use crate::constants::DEFAULT_REMOTE_NAME;
 use crate::core;
 use crate::core::versions::MinOxenVersion;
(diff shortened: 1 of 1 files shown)


CleanCut (approved):

CleanCut:
The test that expects the cleanup needs to be updated.

malcolmgreaves (commented):
@subygan if a clone fails, does one need a specific`oxen fsck` command in order to fix it? Or can another `oxen clone` fix it?

subygan:
I don't think we have a feature that automatically detects if the folder exists and fix it.

subygan:
for now, we don't do that. user will have to consciously fix the repo. might be a good idea to add a flag in clone to fix the repo directly.

PR #388, opened 2026-03-26
Add oxen-py dependencies to workspace

malcolmgreaves:
Adds all of oxen-py's dependencies to the workspace and reuses existing 
workspace versions for the overlapping dependencies. Ensures the version, 
and other Cargo.toml attributes, are aligned with the workspace too.

Since `maturin` requires the LICENSE file to exist at the `oxen-python` root,
a symlink has been made from `oxen-python/LICENSE` to `LICENSE`.
The `oxen-py` crate uses this symlink to appease `maturin`.


crates/oxen-py/Cargo.toml
```diff
@@ -1,22 +1,28 @@
...
-version = "0.46.8"
-edition.workspace = true
+version = { workspace = true }
+edition = { workspace = true }
+license-file = { workspace = true }
+homepage = { workspace = true }
+repository = { workspace = true }
+readme = { workspace = true }
+keywords = { workspace = true }
+publish = false
```
malcolmgreaves, line 11: IIUC we want to publish the python package (`oxen-python`), but we do not want to publish these bindings.

CleanCut: I think you're correct. I don't believe we were publishing these as a Rust crate before, and since they're only consumed by our publish Python package, it wouldn't be helpful to anyone to publish them as a Rust crate.

CleanCut (approved):
❤️ Nice! I forgot to look at having the Cargo.toml dependencies inherit the workspace when I was rearranging things. Thanks for catching that!

PR #389, opened 2026-03-26
Remove unused async-std-runtime feature from oxen-py

malcolmgreaves:
The `oxen-py` crate only uses the Tokio runtime. Thus, there's no need for
the `async-std-runtime` feature for the `pyo3-async-runtimes` dependency.


CleanCut (approved):
❤️

PR #393, opened 2026-03-27
Doc & simplify OxenError; using .ok_or_else(|| OxenError); not-found + conflict HTTP

malcolmgreaves:
**`OxenError` Variants**:
Adding docstrings to all `OxenError` variants. Docstring explains what the variant is
and, if applicable, the situations and code paths where it is encountered. Also,
formatted and re-structured the `impl` block to focus on user-facing elements (hints),
introspection-based functions (`is_...)`, and constructor functions. To organize the
constructors, they're now separated by ones that produce variants and ones that are
`OxenError::Basic` string. The latter has TODO documentation put in describing the
next step of migrating these to their own structured `OxenError` variant.

Audited all `OxenError` variants and identified variants that are never used in code.
These variants have been removed.

Additionally, some variants only ever use statically allocated strings as their inner
value. These have been replaced with non-value bearing enum variants.

A new `RemoteNotSet(String)` enum has replaced the `basic_str` use of `remote_not_set`.

**`.ok_or_else(|| OxenError...)`**:
Audit uncovered unnecessary allocation of many `OxenError` variants in `.ok_or(...)`
constructs on `Result` values. This allocates space for the error even if the `Result`
itself is `Ok`: this value is created and immediately discarded.

All locations of `.ok_or(..)` that were creating an `OxenError` have now been changed
to use the lazily-called variant, `.ok_or_else(|| ...)`, which only constructs
the `OxenError` variant when it is necessary.

**Not-Found & OxenHttpError**
Adds new `OxenHttpError::error_response` match arms for the newly promoted
`OxenError::Basic`-str to structured variants that are considered "not found" to treat
these as `404 Not Found` instead of `500 Internal Server Error`.

**Workspace Behind is HTTP Conflict**
Adds a new match arm in `OxenHttpError::error_response` to treat the `WorkspaceBehind`
error as a HTTP conflict (`409 Conflict`) instead of a 500 error.

### DIFF
--- a/crates/lib/src/core/v_latest/entries.rs
+++ b/crates/lib/src/core/v_latest/entries.rs
@@ -72,19 +72,19 @@ pub fn list_directory_with_depth(
     let commit = parsed_resource
         .commit
         .clone()
-        .ok_or(OxenError::revision_not_found(revision.into()))?;
+        .ok_or_else(|| OxenError::RevisionNotFound(revision.into()))?;
 
     log::debug!("list_directory commit {commit}");
 
     let _perf_get_dir = crate::perf_guard!("core::entries::get_dir_with_children");
     let dir = repositories::tree::get_dir_with_children(repo, &commit, directory, None)?
-        .ok_or(OxenError::resource_not_found(directory.to_str().unwrap()))?;
+        .ok_or_else(|| OxenError::resource_not_found(directory.to_string_lossy()))?;
     drop(_perf_get_dir);
 
     log::debug!("list_directory dir {dir}");
 
     let EMerkleTreeNode::Directory(dir_node) = &dir.node else {
-        return Err(OxenError::resource_not_found(directory.to_str().unwrap()));
+        return Err(OxenError::resource_not_found(directory.to_string_lossy()));
     };
 
     log::debug!("list_directory dir_node {dir_node}");
@@ -155,9 +155,7 @@ pub fn get_meta_entry(
     let commit = parsed_resource
         .commit
         .clone()
-        .ok_or(OxenError::parsed_resource_not_found(
-            parsed_resource.clone(),
-        ))?;
+        .ok_or_else(|| OxenError::parsed_resource_not_found(parsed_resource.clone()))?;
     log::debug!("get_meta_entry path: {path:?} commit: {commit}");
     let node = repositories::tree::get_dir_without_children(repo, &commit, path, None)?;
     log::debug!("get_meta_entry node: {node:?}");
@@ -301,9 +299,10 @@ fn dir_node_to_metadata_entry(
         found_commits.entry(*dir_node.last_commit_id())
     {
         let _perf_commit = crate::perf_guard!("core::entries::get_commit_by_hash");
-        let commit = repositories::commits::get_by_hash(repo, dir_node.last_commit_id())?.ok_or(
-            OxenError::commit_id_does_not_exist(dir_node.last_commit_id().to_string()),
-        )?;
+        let commit = repositories::commits::get_by_hash(repo, dir_node.last_commit_id())?
+            .ok_or_else(|| {
+                OxenError::commit_id_does_not_exist(dir_node.last_commit_id().to_string())
+            })?;
         e.insert(commit);
     }
 
@@ -344,9 +343,10 @@ fn file_node_to_metadata_entry(
         found_commits.entry(*file_node.last_commit_id())
     {
         let _perf_commit = crate::perf_guard!("core::entries::get_commit_by_hash");
-        let commit = repositories::commits::get_by_hash(repo, file_node.last_commit_id())?.ok_or(
-            OxenError::commit_id_does_not_exist(file_node.last_commit_id().to_string()),
-        )?;
+        let commit = repositories::commits::get_by_hash(repo, file_node.last_commit_id())?
+            .ok_or_else(|| {
+                OxenError::commit_id_does_not_exist(file_node.last_commit_id().to_string())
+            })?;
         e.insert(commit);
     }
 
@@ -537,7 +537,7 @@ pub fn list_for_commit(
 
 pub fn update_metadata(repo: &LocalRepository, revision: impl AsRef<str>) -> Result<(), OxenError> {
     let commit = repositories::revisions::get(repo, revision.as_ref())?
-        .ok_or_else(|| OxenError::revision_not_found(revision.as_ref().to_string().into()))?;
+        .ok_or_else(|| OxenError::RevisionNotFound(revision.as_ref().to_string().into()))?;
     let tree: CommitMerkleTree = CommitMerkleTree::from_commit(repo, &commit)?;
     let mut node = tree.root;
 
--- a/crates/lib/src/core/v_latest/workspaces/commit.rs
+++ b/crates/lib/src/core/v_latest/workspaces/commit.rs
@@ -61,7 +61,7 @@ pub async fn commit(
 
         let conflicts = list_conflicts(workspace, &dir_entries, &branch)?;
         if !conflicts.is_empty() {
-            return Err(OxenError::workspace_behind(workspace));
+            return Err(OxenError::WorkspaceBehind(Box::new(workspace.clone())));
         }
 
         let dir_entries = export_tabular_data_frames(workspace, dir_entries).await?;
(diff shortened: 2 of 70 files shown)


crates/lib/src/error.rs
```diff
@@ -8,193 +8,306 @@ use std::io;
...
     // Repo
+    //
+    /// When an operation assumes a repository exists but it cannot be found.
     #[error("Repository '{0}' not found")]
     RepoNotFound(Box<RepoNew>),
+
+    /// When a local repository cannot be found at the given path.
     #[error("No oxen repository found at {0}")]
-    LocalRepoNotFound(Box<PathBufError>),
+    LocalRepoNotFound(PathBufError),
```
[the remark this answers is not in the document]
malcolmgreaves: @subygan do you agree with this AI review comment?

crates/lib/src/error.rs
```diff
@@ -8,193 +8,306 @@ use std::io;
...
-    NothingToCommit(StringError),
-    #[error("{0}")]
-    NoCommitsFound(StringError),
-    #[error("{0}")]
-    HeadNotFound(StringError),
+    RevisionNotFound(StringError),
 
+    /// The repository is empty: it has no commits.
+    #[error("\n No commits found.\n")]
+    NoCommitsFound,
```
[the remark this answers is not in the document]
malcolmgreaves: I agree with this one in-spirit. But I wonder if it's actually straightforward to make the CLI add the spacing to this error only? Or, perhaps we don't think it's worth the maintenance effort and instead should just remove the surrounding newlines?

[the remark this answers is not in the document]
malcolmgreaves: Removing newlines 🫡

crates/lib/src/core/v_latest/workspaces/commit.rs
```diff
@@ -110,16 +110,16 @@ pub fn mergeability(
     let branch_name = branch_name.as_ref();
     let Some(branch) = repositories::branches::get_by_name(&workspace.base_repo, branch_name)?
     else {
-        return Err(OxenError::revision_not_found(
-            branch_name.to_string().into(),
-        ));
+        // TODO: should this instead be BranchNotFound ??
+        //       Or should we eliminate "branch not found" and always use "revision not found"?
```
jcelliott: We should keep both. In Oxen terminology a "revision" is either a commit or branch reference. Many places take either, so "revision not found" is the correct error message. But some APIs *only* deal with branches, so "branch not found" is a better error message for those.

malcolmgreaves: Thanks for the clarification! I'll change this to be `BranchNotFound`.

crates/lib/src/core/v_latest/workspaces/commit.rs
```diff
@@ -174,8 +172,8 @@ fn list_conflicts(
     let Some(branch_commit) =
         repositories::commits::get_by_id(&workspace.base_repo, &branch.commit_id)?
```
subygan, line 173: At some point we'll have to remove this boiler plate of converting a None into an error and handle it directly at a higher layer and only return `Result<Commit, OxenError>` instead of an `Result<Option<Commit, OxenError>>`

crates/lib/src/error.rs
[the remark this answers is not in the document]
subygan: this is correct, we might end up with slightly different error. is worth fixing.

crates/lib/src/error.rs
```diff
@@ -8,193 +8,306 @@ use std::io;
...
 
+    //
     // Remotes
+    //
+    /// A remote repository with a given name was not located on the server.
     #[error("Remote repository not found: {0}")]
-    RemoteRepoNotFound(Box<StringError>),
-    #[error("{0}")]
-    RemoteAheadOfLocal(StringError),
-    #[error("{0}")]
```
subygan, line 58: nice cleanup of these errors!!

crates/lib/src/core/v_latest/entries.rs
```diff
@@ -72,7 +72,7 @@ pub fn list_directory_with_depth(
     let commit = parsed_resource
         .commit
         .clone()
-        .ok_or(OxenError::revision_not_found(revision.into()))?;
+        .ok_or(OxenError::RevisionNotFound(revision.into()))?;
```
subygan, line 76: might be worth standardizing to either `ok_or` or `ok_or_else`.

eg. in line 540. we're using `ok_or_else()`

crates/lib/src/error.rs
```diff
@@ -8,193 +8,306 @@ use std::io;
...
+    // Wrappers
+    //
+    //
+    /// Wraps the error from std::path::strip_prefix.
+    #[error("Error stripping prefix: {0}")]
+    StripPrefixError(#[from] std::path::StripPrefixError),
 
-    // External Library Errors
+    /// Wraps errors encotunered from file reading & writing operations.
     #[error("{0}")]
```
subygan, line 212: typo: encountered

subygan (approved):

subygan:
incredible cleanup!! just some tiny suggestions.

crates/lib/src/core/v_latest/entries.rs
malcolmgreaves: Great catch! I'dd do `ok_or_else` since this is creating a value. I try to use `ok_or` only when the argument is a constant, a reference to something, or is something small and fast to make, like an int or float.

malcolmgreaves:
@subygan I liked your feedback to use `.ok_or_else(|| `, so I did some find-and-replaces to update all places where we were doing `.ok_or(OxenError...)` into `.ok_or_else(|| OxenError...)`. 

I manually went through and preserved the few locations where it's appropriate to do `.ok_or(OxenError::<...>)`: there's a few variants that are now constructor-less, so they're just references to constant values now :)

(https://github.com/Oxen-AI/Oxen/pull/393/changes#r3012224198)

crates/lib/src/error.rs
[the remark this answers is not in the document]
malcolmgreaves: fixed

crates/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -407,12 +423,12 @@ pub async fn upload_zip(
...
             Ok(commit)
         }
-        Err(OxenError::WorkspaceBehind(workspace)) => {
+        workspace_behind_error @ Err(OxenError::WorkspaceBehind(_)) => {
             log::error!(
                 "unable to commit branch {:?}. Workspace behind",
                 branch.name
             );
-            Err(OxenError::WorkspaceBehind(workspace))
+            workspace_behind_error
```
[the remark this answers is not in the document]
malcolmgreaves: Updated so this is now a `409 Conflict` instead of a `500 Internal Server Error`.

crates/lib/src/core/v_latest/workspaces/files.rs
```diff
@@ -270,26 +270,40 @@ fn is_cgn_or_reserved_v4(octets: [u8; 4]) -> bool {
...
-    let resolved = tokio::net::lookup_host(&addr).await.map_err(|e| {
-        OxenError::file_import_error(format!("DNS resolution failed for {host}: {e}"))
-    })?;
+    let resolved =
+        tokio::net::lookup_host(&addr)
+            .await
+            .map_err(|e| InvalidUrlError::DnsResolutionFailed {
+                host: host.to_string(),
+                source: e,
+            })?;
```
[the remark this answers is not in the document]
malcolmgreaves: Seems unrelated to this PR. Will investigate further in another PR, if applciable.

crates/server/src/errors.rs
```diff
@@ -536,19 +509,84 @@ impl error::ResponseError for OxenHttpError {
...
+                            "error": {
+                                "type": MSG_RESOURCE_NOT_FOUND,
+                                "title": "Queryable workspace not found",
+                                "detail": "Queryable workspace not found."
+                            },
+                            "status": STATUS_ERROR,
+                            "status_message": MSG_RESOURCE_NOT_FOUND,
+                        });
+                        HttpResponse::NotFound().json(error_json)
+                    }
```
malcolmgreaves, line 576: These were all previously being treated as `500 Internal Server Error`. They're now `404 Not Found`. This treats them the same as the other `OxenError` variants that are considered `OxenError::is_not_found`.

cc @jcelliott @CleanCut @subygan -- any idea if this change would be a problem in the hub? It's still an error, so I would think no?

crates/server/src/errors.rs
```diff
@@ -536,19 +509,84 @@ impl error::ResponseError for OxenHttpError {
...
+                            "error": {
+                                "type": MSG_CONFLICT,
+                                "title": "Workspace is behind",
+                                "detail": format!("Workspace '{}' is behind at commit {}", workspace.id, workspace.commit.id)
+                            },
+                            "status": STATUS_ERROR,
+                            "status_message": MSG_CONFLICT,
+                        });
+                        HttpResponse::Conflict().json(error_json)
+                    }
```
malcolmgreaves, line 589: This used to be an `HTTP 500` too, now it's a `409 Conflict`. I think this is also ok but same question here for review.

crates/server/src/errors.rs
Eric-Laurence: I can't think of any reason the Hub would have a problem with this. We can test to be sure, but I'd be surprised.

crates/server/src/errors.rs
Eric-Laurence: and same here

PR #396, opened 2026-03-27
Use a larger runner for Windows

CleanCut:
Decrease cache-miss (on windows) CI run time from 50 -> 40 minutes. Decrease cache-hit (on windows) CI run time from 30 -> 20 minutes.

- Use a larger runner for Windows (configured with @jcelliott)
  - The larger runner doesn't have a separate `D:` drive like the free runner does, so stop using that
- The `actions-rust-lang/setup-rust-toolchain` action was already calling `Swatinem/rust-cache` under-the-hood, and that was actually running, so make that explicit and set a couple flags to cache more
- Our `Swatinum/rust-cache` step wasn't actually getting used, so remove that
  - Note: this was configured to only upload cache on `main`, but we were actually uploading on every build through `setup-rust-toolchain` (above). `setup-rust-toolchain` doesn't plumb-through that option, but it wasn't using that option _anyway_...so it's no change.
- Use the `lld` linker that ships with Rust on Windows
- Add ability to specify `OXEN_TEST_RUN_DIR` to customize the output directory for tests that write directly to disk.
  - Our I/O-heavy tests can run faster if we run them on a ramdisk. Unfortunately, we currently use filename canonicalization that doesn't support ramdisks, so that needs to be fixed first before we can use this functionality to speed up CI.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -73,14 +73,10 @@ jobs:
           # Use ramdisk for cargo build artifacts
           echo "CARGO_TARGET_DIR=R:\target" >> $env:GITHUB_ENV
 
-          # Move repo data directory to ramdisk for faster test I/O
-          if (Test-Path "data") {
-            Copy-Item -Recurse -Path "data" -Destination "R:\data"
-            Remove-Item -Recurse -Force "data"
-          } else {
-            New-Item -ItemType Directory -Path "R:\data" | Out-Null
-          }
-          cmd /c mklink /J data R:\data
+          # Junction only the test runs directory to ramdisk for faster test I/O
+          New-Item -ItemType Directory -Path "R:\runs" | Out-Null
+          New-Item -ItemType Directory -Path "data\test" -Force | Out-Null
+          cmd /c mklink /J data\test\runs R:\runs
 
       - name: Setup Rust Toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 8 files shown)


malcolmgreaves (approved):

.github/workflows/ci_test.yml
```diff
@@ -146,9 +146,6 @@ jobs:
       - name: Build oxen rust project
         run: |
           cargo build --workspace
-        env:
-          TEMP: ${{ matrix.platform == 'windows' && 'D:\a\_temp' || runner.temp }}
-          TMP: ${{ matrix.platform == 'windows' && 'D:\a\_temp' || runner.temp }}
```
malcolmgreaves, line 151: I don't know how this works. What was this for?

CleanCut: The standard `windows-latest` runner currently has a `D:` drive that is larger and faster that `C:`. This just configured env vars so that `D:` would be used for temp files for various build tools. The larger runner does not have a separate `D:` drive, apparently.

CleanCut:
The screenshot below represents the best-case for splitting into separate Python and Rust tests (14m overall). This was only achievable artificially, by running the CI job twice in a row after hours when no one else was working. Splitting into 6 jobs with separate cache keys pretty much busts the cache in a single run. So even with no other activity I had to run the jobs back-to-back to get a cache hit before the cache eviction process took effect. In other words, this isn't a practical approach under GitHub actions with a 10GB cache, so I'll be merging the Python/Rust jobs back together.

I believe we can get our CI down around 5 minutes with a non-GitHub provider, if and when we want to tackle that.

Screenshot 2026-04-09 at 9 19 01 AM

CleanCut:
Calling this good for now. We've ended up reducing CI by about 10 minutes (50m -> 40m for cache misses, 30m -> 20m for cache hits).

Screenshot 2026-04-09 at 12 48 23 PM

PR #398, opened 2026-03-27
Implement S3VersionStore::store_version_from_reader

CleanCut:
I'm tackling the S3 implementation one method at a time, and there's many more methods to go, so it will be a bit until we get into a state where we can actually try it out. But I have unit tests [in a separate PR](https://github.com/Oxen-AI/Oxen/pull/404) (since they required some hefty dependency updates).

- Implement the S3VersionStore::store_version_from_reader
  - Now requires specifying the file size up front, which all callers can easily do.
  - Uploads files <= 100MB in one shot (as per AWS recommendations)
  - Determines a file part size dynamically based on the file size for files > 100MB
  - Does not write to disk
  - Uploads up to 16 file parts concurrently
  - Cancels the multipart upload if anything goes wrong.
- Add a new `OxenError::AwsS3Error` variant.
- Updated the aws crates
- Added instructions for claude to stop creating functions for random tiny bits of code just to call them exactly once and to stop deleting relevant comments🤞🏻


crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,165 @@ impl VersionStore for S3VersionStore {
...
+            let _ = client
+                .abort_multipart_upload()
+                .bucket(&self.bucket)
+                .key(&key)
+                .upload_id(&upload_id)
+                .send()
+                .await;
+        }
+
+        result
```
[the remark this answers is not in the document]
CleanCut: Wow, that's a really good catch. It took some time to fix this one. I added some descriptive comments to help folks understand the roadmap of the function as we go along, as it's not immediately obvious.

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
 
     async fn store_version_from_reader(
         &self,
-        _hash: &str,
-        _reader: &mut (dyn tokio::io::AsyncRead + Send + Unpin),
+        hash: &str,
+        reader: &mut (dyn tokio::io::AsyncRead + Send + Unpin),
```
malcolmgreaves: I like the design of treating `PART_SIZE` as a delineation between small and large files. And that large files are multi-part uploaded with `PART_SIZE` bytes per part. Very logical.

Also, I think it makes a ton of sense to define `MAX_CONCURRENT_UPLOADS` and `MAX_PARTS_WAITING_FOR_UPLOAD` to control the memory use during the upload process.

Though these are great parts of the design, I think they should be exposed to the caller. These behaviors should be changeable at runtime. Doing so offers flexibility to be able to adapt to different situations.

For instance, coderabbit points out that this implementation is unable to upload files larger than ~80GB. While uncommon, this isn't unheard of and it isn't a current limitation for oxen repositories. Since we're handling large amounts of data, I think we want to be intentional with our file size limits. And we should default to a larger-is-better approach whenever we can.

If we make these options parameters, then we can handle up to S3's max theoretical size of a 50TB object (`PART_SIZE` of 5GB). We will _absolutely_ run into local FS issues before that 😉 , but this is just a question of scale: we can get a 16 TiB max file on ext4 systems, and we should be able to stuff that into S3.

malcolmgreaves: I think this is a trait method, right? So I can see a few different solutions here:

1. Define these as internal fields on the `struct` that implements this trait for S3. Have the constructor accept these 3 options and use them here.

2. Define a trait type and use that here for the "configuration options." For this s3 implementation, define a new struct for this config and store the 3 values in it. Then, pass this struct into it at the call site. The non-s3 case may be nothing (e.g. `()`). e.g.
```rust
trait VersionStore {

  type Config;   

   async fn store_version_from_reader(&self, 
    hash: &str,
    reader: &mut (dyn tokio::io::AsyncRead + Send + Unpin),
    config: &Self::Config,
  ) -> ) -> Result<(), OxenError>;

...
}

I think (1) may be a bit cleaner for the trait case, but it means we'd have to re-construct or modify the struct in order to change behavior at runtime, versus changing the option and calling again. That may be an ok ergonomics tradeoff since we may not actually change the options too much at runtime.

I like (2) more for it's modularity. It seems more appropriate as a library-like function, however.

Maybe use (1) but have the actual implementation be a full parameterized function? Then this trait method implementation can just wrap it using the struct's set fields.

File: crates/lib/src/storage/s3.rs [hunk 3]
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+            .create_multipart_upload()
+            .bucket(&self.bucket)
+            .key(&key)
+            .send()
+            .await
+            .map_err(OxenError::aws_sdk_error)?;
+
+        let upload_id = upload
+            .upload_id()
+            .ok_or_else(|| OxenError::basic_str("S3 multipart upload missing upload_id"))?
```
malcolmgreaves: I would really like to cut down on the use of `OxenError::Basic` in the codebase. With @subygan's new changes to `OxenError` using the `[#error(..)]` macro, I think it's back to being quite easy to write new static-like errors.

Could you please change this to using a new `OxenError` variant?

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+                            .key(key)
+                            .upload_id(upload_id)
+                            .part_number(part_num)
+                            .body(ByteStream::from(data))
+                            .send()
+                            .await
+                            .map_err(OxenError::aws_sdk_error)?;
+
+                        let etag = resp.e_tag().map(|s| s.to_string()).ok_or_else(|| {
+                            OxenError::basic_str("S3 upload_part response missing ETag")
```
malcolmgreaves: Could you please change this to use a new `OxenError` variant? (https://github.com/Oxen-AI/Oxen/pull/398/changes#r3013028064)

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+                })
+                .buffer_unordered(MAX_CONCURRENT_UPLOADS)
+                .try_collect()
+                .await;
+            results
+        });
+
+        // Producer: read parts and feed them into the channel. This loop happens on this task.
+        let producer_result: Result<(), OxenError> = async {
+            let mut part_num: i32 = 1;
```
malcolmgreaves: They're 1-indexed! Didn't know that! :)

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+            tx.send((part_num, first_buf)).await.map_err(|_| {
+                OxenError::basic_str("Upload task terminated while sending first part")
```
malcolmgreaves: Could you please change this to use a new `OxenError` variant? (https://github.com/Oxen-AI/Oxen/pull/398/changes#r3013028064)

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+            loop {
+                let mut buf = vec![0u8; PART_SIZE];
+                let n = read_full(&mut reader, &mut buf).await?;
+                if n == 0 {
+                    // We have reached EOF
+                    break;
+                }
+                buf.truncate(n);
+                tx.send((part_num, buf)).await.map_err(|_| {
+                    OxenError::basic_str("Upload task terminated while sending parts")
```
malcolmgreaves: Could you please change this to use a new `OxenError` variant? (https://github.com/Oxen-AI/Oxen/pull/398/changes#r3013028064)

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+        // Close the channel so the upload task finishes draining any buffered parts.
+        drop(tx);
+
+        let result = match producer_result {
+            // The producer task completed reading the file and sent all parts to the upload task,
+            // so wait for the upload task to finish next.
+            Ok(()) => {
+                match upload_task_join_handle
+                    .await
+                    .map_err(|e| OxenError::basic_str(format!("Upload task panicked: {e}")))
```
malcolmgreaves: Could you please change this to use a new `OxenError` variant? (https://github.com/Oxen-AI/Oxen/pull/398/changes#r3013028064)

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +175,187 @@ impl VersionStore for S3VersionStore {
...
+                            .await
+                            .map_err(OxenError::aws_sdk_error)
+                            .map(|_| ())
+                    }
+                    Ok(Err(e)) | Err(e) => Err(e),
+                }
+            }
+            // The producer task wasn't able to successfully read the file and send the parts to the
+            // upload task, so we ensure the upload task is cancelled and awaited so that no
+            // in-flight upload_part calls race with abort_multipart_upload below.
```
malcolmgreaves: The in-line comments here are _fantastic_. Thank you for writing them so clearly. The whole code in this implementation is very well structured and easy to follow ❤️

crates/lib/src/storage/s3.rs
malcolmgreaves: If you don't want to add them into `OxenError`, then, since they only occur in this code, I would alternatively suggest making a new more-specific error `enum` for these errors. Then having a single new variant in `OxenError` that contains this new enum and making a new `impl From<that new enum> OxenError` to convert into an `OxenError`.

That's ☝️ the refactoring direction I'd like to migrate all of `oxen` into over time. More functions returning more locally-scoped errors. Ideally, errors map to coherent modules of the codebase so they're maximally specific and scoped. Then we have `impl From`s to convert them into an `OxenError`.

crates/lib/src/storage/s3.rs
```diff
@@ -500,6 +678,27 @@ impl VersionStore for S3VersionStore {
...
+async fn read_full(
+    reader: &mut (dyn tokio::io::AsyncRead + Send + Unpin),
+    buf: &mut [u8],
+) -> Result<usize, OxenError> {
+    let mut offset = 0;
+    while offset < buf.len() {
+        let n = reader
+            .read(&mut buf[offset..])
+            .await
+            .map_err(|e| OxenError::basic_str(format!("Failed to read from reader: {e}")))?;
```
malcolmgreaves: Could you please change this to use a new `OxenError` variant? (https://github.com/Oxen-AI/Oxen/pull/398/changes#r3013028064)

crates/lib/src/error.rs
```diff
@@ -147,6 +147,8 @@ pub enum OxenError {
     ImportFileError(StringError),
 
     // External Library Errors
+    #[error("AWS SDK error: {0}")]
+    AwsSdkError(Box<dyn std::error::Error + Send + Sync>),
```
malcolmgreaves: Is there not a more specific AWS error type than this dynamic set of traits? Aren't we using the AWS rust client that returns these errors? https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/types/error/index.html

Even if it requires making a wrapper, it would be great to have this be a specific type. As is, there would too many things that wouldn't be related to AWS SDK errors that could fit into an `OxenError::AwsSdkError`.

malcolmgreaves (commented):
Overall a really great PR! This is very clean and easy to follow along, which is wonderful -- the codebase needs more PRs like this one!

I have two major comments on this PR. The first is about making the part_size and other runtime options of the upload configurable. This is so that we can have flexibility to change at runtime as well as the ability to handle files larger than ~80GB, which is a limitation of this implementation.

The second comment is about `OxenError`. There's two sub-comments to this one. First is a request to use structured error variants instead of the `OxenError::Basic` variant for the different error conditions introduced here. The second one is to use a more specific error type for the AWS sdk error. The `dyn` trait object is too wide.

Both of my comments have some suggested implementations. Happy to discuss through those further. Ultimately, I'm more looking for the end state and don't favor any approach over the other.

crates/lib/src/storage/s3.rs
CleanCut: Ya, so weird!

crates/lib/src/storage/s3.rs
CleanCut: Thanks! I spent a lot of time trying to make sure others (and my future self) could follow along.

crates/lib/src/storage/s3.rs
CleanCut: I checked with [Josh in Slack yesterday](https://oxenai.slack.com/archives/C03EBLD4B7W/p1774911561475529?thread_ts=1774907480.567669&cid=C03EBLD4B7W), and he was fine starting with this file limit as an MVP.

I added too many bells and whistles as it is. This multi-part upload stuff is complicated, and we'll only see benefits if the single backend S3 connection was a bottleneck. I'm inclined to simplify it down to directly copy the reader stream up through to S3, sidestepping all capacity limits, and just see how it performs. So if we have problems with this setup, that's what we should do next.

crates/lib/src/error.rs
CleanCut: ✅ Tightened up.

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs
CleanCut: You know what, I'm just going to pull the trigger and simplify this. I'm a bit annoyed at myself for prematurely optimizing when we don't even know if we have a bottleneck. And then we ended up with this silly artificial limitation that will require making it even more complicated to work around. I should have gone for the simplest approach in the first place. I'll comment when I've got something...

CleanCut: ✅ Done. Way simpler now.

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +171,29 @@ impl VersionStore for S3VersionStore {
...
-        ))
+        let client = self.init_client().await?;
+        let key = self.generate_key(hash);
+
+        // Stream the reader directly to S3 via put_object. Allocate 8MB to optimize for large files.
+        let stream = tokio_util::io::ReaderStream::with_capacity(reader, 8 * 1024 * 1024);
+        let body = ReaderBody {
+            stream: std::sync::Mutex::new(stream),
+        };
+        let byte_stream = ByteStream::from_body_1_x(body);
```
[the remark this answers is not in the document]
CleanCut: And yet...it compiles.

crates/server/src/controllers/commits.rs
```diff
@@ -1245,7 +1242,7 @@ async fn unpack_tree_tarball(tmp_dir: &Path, data: &[u8]) -> Result<(), OxenErro
 
 async fn unpack_entry_tarball_async(
     repo: &LocalRepository,
-    compressed_data: &[u8],
+    compressed_data: Vec<u8>,
```
malcolmgreaves, line 1245: (minor comment) Do we need to consume this `compressed_data`? Or would a borrow on it still suffice for creating `let reader = Cursor::new(compressed_data);`? I feel like we can keep this one as-is, but perhaps I'm missing something?

malcolmgreaves (approved):
LGTM! 👏 Minor non-blocking comments.

crates/lib/src/error.rs
```diff
@@ -416,6 +429,10 @@ impl OxenError {
         OxenError::InvalidVersion(StringError::from(s.as_ref()))
     }
 
+    pub fn upload(s: &str) -> Self {
```
malcolmgreaves, line 432:
```suggestion
    /// Makes an OxenError::Upload error.
    pub fn upload(s: &str) -> Self {
```

crates/lib/src/error.rs
```diff
@@ -195,6 +203,11 @@ pub enum OxenError {
     #[error("{0}")]
     ImportFileError(StringError),
 
+    // External Library Errors
+    /// An error encountered dealing with S3
+    #[error("AWS S3 error: {0}")]
+    AwsS3Error(Box<SdkError<Box<dyn std::error::Error + Send + Sync>, HttpResponse>>),
```
malcolmgreaves: Thank you for tracking down this AWS error type! Minor comment -- there's a "Wrappers" section I documented in OxenError. This would be a good candidate to add under that section to preserve the organization :)

crates/lib/src/storage/s3.rs
```diff
@@ -500,6 +632,57 @@ impl VersionStore for S3VersionStore {
     }
 }
 
+async fn upload_part(
```
malcolmgreaves, line 642:
```suggestion
/// Uploads a single part in an ongoing multipart S3 upload operation.
async fn upload_part(
```

crates/lib/src/storage/s3.rs
```diff
@@ -171,13 +173,143 @@ impl VersionStore for S3VersionStore {
 
     async fn store_version_from_reader(
```
malcolmgreaves, line 180: If you can, it'd be helpful to add a short function comment here detailing how the multi-part upload algorithm you made here works (essentially what you have in the PR comment).

crates/server/src/controllers/commits.rs
CleanCut: A `&[u8]` won't work here because `store_version_from_reader` on line 1270 now requires owned data, and `file` derives it's lifetime from `compressed_data`. I'm not sure exactly _why_ the lifetime of the `&[u8]` is not deemed sufficient, but I can't find a way to get the compiler to allow it. Lucky for us, the callers didn't need the data after calling this, so it was an easy change.

crates/lib/src/error.rs
CleanCut: 😂 You reorganized the section right out from under me during this PR! This was in what became that section and ended up here instead during the final merge conflict resolution I just did!

crates/lib/src/storage/s3.rs
CleanCut: ✅ Done

crates/lib/src/storage/s3.rs
CleanCut: ✅ Done

crates/lib/src/error.rs
CleanCut: ✅ Done

crates/lib/src/error.rs
CleanCut: ✅ Done

PR #399, opened 2026-03-27
Fix conflicting names between the cli and the oxen-py crates

CleanCut:
Use a unique name for the `oxen-py` crate library to fix the warning from Cargo that `cli` and `oxen-py` produce artifacts with the same name that will become a hard error in the future.

This doesn't change anything from a Python user's perspective. They will still do `import oxen`, as the python module is named `oxen`, then under-the-hood it will access the Rust python interface through `oxen_py`.

### DIFF
--- a/crates/oxen-py/Cargo.toml
+++ b/crates/oxen-py/Cargo.toml
@@ -11,7 +11,7 @@ keywords = { workspace = true }
 publish = false
 
 [lib]
-name = "_oxen" # The native extension module. Users still `import oxen` via the Python package wrapper.
+name = "oxen_py" # The native extension module. Users still `import oxen` via the Python package wrapper.
 crate-type = ["cdylib"] # PyO3 needs this crate to be a cdylib so it can link against it.
 
 [dependencies]
--- a/crates/oxen-py/Cargo.toml
+++ b/crates/oxen-py/Cargo.toml
@@ -11,7 +11,7 @@ keywords = { workspace = true }
 publish = false
 
 [lib]
-name = "oxen" # Necessary so that we can import the crate as `import oxen` in Python.
+name = "_oxen" # The native extension module. Users still `import oxen` via the Python package wrapper.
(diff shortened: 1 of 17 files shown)


crates/oxen-py/Cargo.toml
```diff
@@ -11,7 +11,7 @@ keywords = { workspace = true }
 publish = false
 
 [lib]
-name = "oxen" # Necessary so that we can import the crate as `import oxen` in Python.
+name = "oxen_py" # The native extension module. Users still `import oxen` via the Python package wrapper.
```
CleanCut, line 14: This is the problem we're fixing. The cli also produces an `oxen` artifact, which won't be allowed in the same workspace in the future.

oxen-python/tests/test_data_frame.py
```diff
@@ -14,10 +14,8 @@ def test_data_frame_crud(celeba_remote_repo_fully_pushed):
 
     train_path = str(PurePath("annotations", "train.csv"))
     df = DataFrame(
-        remote_repo.identifier,
+        remote_repo,
         train_path,
-        host="localhost:3000",
-        scheme="http",
```
CleanCut, line 20: These tests were hard-coded to port 3000, but other tests already support running against any port, so we switch to using the same pattern on these ones as well.

malcolmgreaves (changes requested):
We don't want to change the name of the Python import. That is a large breaking change for users. From Python, it should remain as `import oxen`. It looks like this has changed the name to `oxen_py`. If I've misread these, please let me know.

Documentation in pyo3 here (at the start of the the `Cargo.toml` section ): https://pyo3.rs/v0.28.2/

malcolmgreaves (approved):
Prior review was wrong, LGTM! ([Discussion in Slack](https://oxenai.slack.com/archives/C0AJ635QE3D/p1774896239348269))

PR #401, opened 2026-03-29
disable debug logging for h2, reqwest, hyper_util libraries + request id in cli logs

subygan:
client side logs are very noisy when run with, `RUST_LOG=debug` because of logs from h2 library. this PR turns off debug logs from h2.

this can still be enabled by `RUST_LOG=debug,h2=debug`

- removes logs from, `h2`, `reqwest` and `hyper_util`.
- also makes the `requestId` visible in the client side


gschoeni (approved):
Good add, I was getting annoyed by this log too 😅

gschoeni (approved):

PR #402, opened 2026-03-29
sort dir by date

gschoeni:
New: Sort parameters on the directory listing endpoint
 
GET /api/repos/{namespace}/{repo_name}/dir/{resource} now accepts two new optional query parameters:
    - sort_by — "name" (default) or "date" (sorts by the entry's latest commit timestamp)
    - reverse — true or false (default false, which is ascending)
 
Examples:
    ?sort_by=date                  # oldest first
    ?sort_by=date&reverse=true     # newest first
    ?sort_by=name&reverse=true     # Z-A

### DIFF
--- a/crates/lib/src/api/client/dir.rs
+++ b/crates/lib/src/api/client/dir.rs
@@ -38,7 +38,7 @@ pub async fn list_with_opts(
     page: usize,
     page_size: usize,
     sort_opts: Option<&SortOpts>,
-    depth: Option<usize>,
+    depth: Option<isize>,
 ) -> Result<PaginatedDirEntries, OxenError> {
     let revision = revision.as_ref();
     let path = path.as_ref().to_string_lossy();
--- a/crates/lib/src/api/client/dir.rs
+++ b/crates/lib/src/api/client/dir.rs
@@ -5,6 +5,7 @@ use crate::error::OxenError;
 use crate::model::RemoteRepository;
 use crate::model::metadata::MetadataDir;
 use crate::model::metadata::generic_metadata::GenericMetadata;
+use crate::opts::SortOpts;
 use crate::view::entries::EMetadataEntry;
 use crate::view::{PaginatedDirEntries, PaginatedDirEntriesResponse};
 use crate::{api, constants};
@@ -27,9 +28,31 @@ pub async fn list(
     page: usize,
     page_size: usize,
 ) -> Result<PaginatedDirEntries, OxenError> {
+    list_with_opts(remote_repo, revision, path, page, page_size, None, None).await
+}
+
+pub async fn list_with_opts(
+    remote_repo: &RemoteRepository,
+    revision: impl AsRef<str>,
+    path: impl AsRef<Path>,
+    page: usize,
+    page_size: usize,
+    sort_opts: Option<&SortOpts>,
+    depth: Option<usize>,
+) -> Result<PaginatedDirEntries, OxenError> {
     let revision = revision.as_ref();
     let path = path.as_ref().to_string_lossy();
-    let uri = format!("/dir/{revision}/{path}?page={page}&page_size={page_size}");
+    let mut query_params = vec![format!("page={page}"), format!("page_size={page_size}")];
+    if let Some(sort_opts) = sort_opts {
+        query_params.push(format!("sort_by={}", sort_opts.sort_by));
+        if sort_opts.reverse {
+            query_params.push("reverse=true".to_string());
+        }
+    }
+    if let Some(depth) = depth {
+        query_params.push(format!("depth={depth}"));
+    }
+    let uri = format!("/dir/{revision}/{path}?{}", query_params.join("&"));
     let url = api::endpoint::url_from_repo(remote_repo, &uri)?;
 
     let client = client::new_for_url(&url)?;
@@ -101,10 +124,12 @@ mod tests {
     use crate::constants::DEFAULT_BRANCH_NAME;
     use crate::error::OxenError;
     use crate::model::StagedEntryStatus;
+    use crate::opts::{SortBy, SortOpts};
     use crate::repositories;
     use crate::test;
     use crate::util;
     use crate::view::entries::EMetadataEntry;
+    use tokio::time::sleep;
 
     use std::path::Path;
     use std::path::PathBuf;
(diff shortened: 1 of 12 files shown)


crates/server/src/controllers/dir.rs
```diff
@@ -49,6 +49,14 @@ pub async fn get(
         d if d < 0 => usize::MAX,
         d => d as usize,
     };
+    let sort_by = match query.sort_by.as_deref() {
+        Some("date") => SortBy::Date,
+        _ => SortBy::Name,
```
CleanCut: **Suggestion:** Let's raise an error if invalid input is received. It can be confusing to the user if the query succeeds but behaves differently than they intended.

```suggestion
        Some("name") => SortBy::Name,
        _ => return OxenError::BadRequest("invalid value for sort_by"),
```

crates/lib/src/opts/sort_opts.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+impl Default for SortOpts {
+    fn default() -> Self {
+        SortOpts {
+            sort_by: SortBy::Name,
+            reverse: false,
+        }
+    }
+}
```
CleanCut, line 71: **Suggestion**: We can tighten this up.

```suggestion
use derive_more::Display;
use serde::{Deserialize, Serialize};

#[derive(Clone, Debug, Default, Display, Serialize, Deserialize, PartialEq, Eq)]
#[serde(rename_all = "lowercase")]
// TODO: When we upgrade to derive_more 2.0.0, use #[display(rename_all = "lowercase")]
pub enum SortBy {
    #[default]
    #[display("name")]
    Name,
    #[display("date")]
    Date,
}

#[derive(Clone, Debug, Default, Serialize, Deserialize)]
pub struct SortOpts {
    pub sort_by: SortBy,
    pub reverse: bool,
}

```

crates/lib/src/api/client/dir.rs
```diff
@@ -491,4 +516,127 @@ mod tests {
...
+            let remote_repo = test::create_remote_repo(&local_repo).await?;
+
+            let file_a = local_repo.path.join("a_file.txt");
+            util::fs::write_to_path(&file_a, "content a")?;
+            repositories::add(&local_repo, &file_a).await?;
+            repositories::commit(&local_repo, "adding a_file")?;
+            repositories::push(&local_repo).await?;
+
+            sleep(std::time::Duration::from_millis(1100)).await;
+
```
CleanCut: We're using monotonic clocks already, so the sleep slows the test down without providing any benefit.
```suggestion
```

crates/lib/src/api/client/dir.rs
```diff
@@ -491,4 +516,127 @@ mod tests {
...
+            util::fs::write(dir_a.join("file_a1.txt"), "a1 content")?;
+            util::fs::write(dir_a_subdir.join("file_sub.txt"), "sub content")?;
+            util::fs::write(dir_b.join("file_b1.txt"), "b1 content")?;
+
+            repositories::add(&local_repo, &local_repo.path).await?;
+            repositories::commit(&local_repo, "Adding nested structure")?;
+            repositories::push(&local_repo).await?;
+
+            sleep(std::time::Duration::from_millis(1100)).await;
+
```
CleanCut:
```suggestion
```

crates/lib/src/repositories/entries.rs
```diff
@@ -397,10 +401,11 @@ mod tests {
     use uuid::Uuid;
 
     use crate::error::OxenError;
-    use crate::opts::PaginateOpts;
+    use crate::opts::{PaginateOpts, SortBy, SortOpts};
     use crate::repositories;
     use crate::test;
     use crate::util;
+    use tokio::time::sleep;
```
CleanCut, line 408:
```suggestion
```

crates/lib/src/repositories/entries.rs
```diff
@@ -1150,7 +1156,13 @@ mod tests {
             util::fs::write(dir_b.join("file_b1.txt"), "b1 content")?;
 
             repositories::add(&repo, &repo.path).await?;
-            let commit = repositories::commit(&repo, "Adding nested structure")?;
+            let _first_commit = repositories::commit(&repo, "Adding nested structure")?;
+
+            sleep(std::time::Duration::from_millis(1100)).await;
+
```
CleanCut, line 1162: Ditto.

```suggestion
```

crates/lib/src/api/client/dir.rs
```diff
@@ -101,10 +124,12 @@ mod tests {
     use crate::constants::DEFAULT_BRANCH_NAME;
     use crate::error::OxenError;
     use crate::model::StagedEntryStatus;
+    use crate::opts::{SortBy, SortOpts};
     use crate::repositories;
     use crate::test;
     use crate::util;
     use crate::view::entries::EMetadataEntry;
+    use tokio::time::sleep;
```
CleanCut:
```suggestion
```

crates/lib/src/api/client/dir.rs
```diff
@@ -26,10 +27,32 @@ pub async fn list(
     path: impl AsRef<Path>,
     page: usize,
     page_size: usize,
+) -> Result<PaginatedDirEntries, OxenError> {
+    list_with_opts(remote_repo, revision, path, page, page_size, None, None).await
+}
+
+pub async fn list_with_opts(
```
CleanCut, line 34: (Pragmatic change for this PR 👉🏻 ) ⚠️ Nothing calls this code with `sort_opts` and `depth` populated except for unit tests. So this doesn't actually work, yet. Everything that calls `list_root` or `list` should be evaluated to see if they should call `list_with_opts` instead.

(👇🏻 Holistic change to consider for the future...)

It would speed up development if we had one `list` function with an options struct that everyone calls. If some things use a particular config a lot, add a helper method to the options struct to pre-configure it rather than chaining functions with hard-coded literals to fill out options.

<details><summary>More discussion about that...</summary>
<br>
It is a code smell to use the "functions that pre-populate certain fields" pattern unless you're trying to maintain some public API compatibility. It slows things down by making it hard to refactor things without introducing a long tail of bugs to follow up on.
<br><br>
For example, we have: `list_root` that calls `list` that calls `list_with_opts`. At each level we hard-code literals for extra parameters for the level down. When we altered this chain, we didn't consider whether or not things up the chain should also use the new stuff. For example: Should `list_root` be able to sort by name? Specify depth? What about `list`? What about their callers--do the callers need to start calling a different function? We didn't touch a single one of those callers (nor call the new function at all 😬). 
<br><br>

Instead, we should create an options struct and fiddle with it at the call site, and everything should call something like this with their chosen options:

```rust
pub async fn list(list_options: ListOptions) -> Result<PaginatedDirEntries, OxenError>
```

(And whatever non-option parameters there are, if any)</details>

crates/lib/src/repositories/entries.rs
```diff
@@ -1202,22 +1216,46 @@ mod tests {
             if let Some(crate::view::entries::EMetadataEntry::MetadataEntry(e)) = dir_a_entry {
                 assert!(e.children.is_some());
                 let children = e.children.as_ref().unwrap();
-                // dir_a should have: file_a1.txt and subdir
-                assert_eq!(children.len(), 2);
+                // dir_a should have: file_a1.txt, file_a2.txt, and subdir
+                assert_eq!(children.len(), 3);
```
CleanCut, line 1223: We should test the default sorting, too.

```suggestion
                assert_eq!(children.len(), 3);
                assert_eq!(children[0].filename, "file_a1.txt");
                assert_eq!(children[1].filename, "file_a2.txt");
                assert_eq!(children[2].filename, "subdir");
```

CleanCut (commented):
- ⚠️ This PR doesn't yet work when you call the actual client. (see below)
- `depth` was added as a parameter--was that a bug fix that was just thrown in? I see it present in a lot of other places.
- A few suggestions

crates/server/src/controllers/dir.rs
gschoeni: Cool, I like when error messages are clear for invalid parameters rather than swallowing them

crates/lib/src/api/client/dir.rs
gschoeni: removed 👍

crates/lib/src/opts/sort_opts.rs
gschoeni: I tightened it up, but also improved the error message a little so it tells you what the valid params are

crates/lib/src/api/client/dir.rs
gschoeni: I like that as a rule of thumb, makes it much easier to reason about and track down and test behavior.

CleanCut: Exactly.

CleanCut (approved):
The minor depth bugs are probably valid. If you don't want to fix them now, it'd probably be sufficient to add a `// TODO ...` to follow up on later.

PR #404, opened 2026-03-30
Tests for S3VersionStore::store_version_from_reader

CleanCut:
_Note: I won't blame you if you *can't find* this PR!_ 🤣 (Look at the PR number)

These tests are in a separate PR from #398 because pulling in `s3s` as a dev dependency required a newer `chrono`, which required a newer `duckdb`, which required a newer `arrow`. It felt like isolating this in its own PR in case of side effects would be a good idea (and there was one observed side effect that had to be fixed).

We need to get over this hurdle so we can have unit tests for the S3 implementation.


crates/lib/Cargo.toml
```diff
@@ -122,7 +122,11 @@ zip = { workspace = true }
 
 [dev-dependencies]
 criterion = { workspace = true }
+hyper = { version = "1", features = ["server", "http1"] }
+hyper-util = { version = "0.1", features = ["tokio"] }
 mockito = { workspace = true }
+s3s = "0.13"
+s3s-fs = "0.13"
```
malcolmgreaves: Please update these to use dependencies from the workspace.

crates/lib/src/storage/s3.rs
```diff
@@ -722,3 +733,137 @@ impl Stream for ByteStreamAdapter {
...
+        let retrieved = store.get_version("bbcdef1234567890").await.unwrap();
+        assert_eq!(retrieved.len(), data.len());
+        assert_eq!(retrieved, data);
+    }
+
+    #[tokio::test]
+    async fn test_store_version_from_reader_exact_part_boundary() {
+        let (store, _tmp, _server) = setup().await;
+
+        // Exactly 16MB -- two full 8MB parts, no partial last part
```
malcolmgreaves, line 834: Great test cases 👏

crates/lib/src/storage/s3.rs
```diff
@@ -86,6 +86,17 @@ impl S3VersionStore {
         }
     }
 
+    #[cfg(test)]
```
malcolmgreaves, line 87: TIL -- This is a very neat way to write test-only utility code!

crates/lib/src/core/v_latest/workspaces/data_frames.rs
```diff
@@ -194,6 +194,11 @@ pub async fn rename(
     let new_db_path = repositories::workspaces::data_frames::duckdb_path(workspace, new_path);
     let new_db_path_parent = new_db_path.parent().unwrap();
 
+    // Close the cached connection before copying so DuckDB flushes its WAL to
+    // disk. Without this, the copy would include a WAL file that references
+    // the original catalog name, causing replay failures when the copy is opened.
+    df_db::remove_df_db_from_cache(&og_db_path)?;
+
```
malcolmgreaves, line 201: Excellent thing to document 👏

malcolmgreaves (approved):
Great PR w/ 1 requested change! Easy to follow. Great test cases and excellent in-line documentation. One requested change before merging -- please use workspace versions for `hyper`, `hyper-util`, `s3s`, and `s3s-fs`.

crates/lib/src/storage/s3.rs
CleanCut: Yep! Best to avoid compiling test helpers and including the bytes in the release.

crates/lib/src/core/v_latest/workspaces/data_frames.rs
CleanCut: Ya, calling out to external dependencies comes with lots of crazy behaviors you've got to account for. Might as well explain why.

crates/lib/Cargo.toml
CleanCut: ✅ Done

PR #406, opened 2026-04-01
Make `TMerkleTreeNode` dyn compatible + rm from_u8 panic!

malcolmgreaves:
**`trait TMerkleTreeNode` Changes**
First in a series of PRs to abstract the Merkle tree store so that we can provide different
backing implementations other than the existing custom file format based implementation.

This PR changes drops the `+ Serialize` constraint to `trait TMerkleTreeNode` to make it 
`dyn` Compatible (this was formally known as _`object safe`_). `Serialize` has a method
`serialize<S: Serializer>` that prevents it from being compiled as a virtual table lookup,
hence it's not `dyn` Compatible. [1]

The change here drops the `+ Serialize` constraint from the trait and instead adds a new
method `to_msgpack_bytes() -> Result<Vec<u8>, ...>`, which performs the serialization.

There's a blanket implementation that implements this updated trait for all existing concrete
node implementations: it uses their `Serialize` impl. to provide a generic implementation for
`to_msgpack_bytes` with a shared serialization error of `rmp_serde::encode::Error`.

Sites that were manually calling `serialize` on the node now call `to_msgpack_bytes()`.
There's a new conversion of `encode::Error` into an `OxenError` wrapper.

[1] https://doc.rust-lang.org/reference/items/traits.html#r-items.traits.dyn-compatible

**Return Serde Error More Often**
Several functions were returning an `OxenError` even though they were only ever returning
a serde serialization error (`rmp_serde::encode::Error`). Since there's an `OxenError`
variant + `From` implementation to convert these, the functions' signatures have been
modified to return the more tightly scoped serde error and have callers rely on using
a conversion via the `?` operator into an `OxenError`.

**`from_u8` changes**
`liboxen::model::merkle_tree::node_type::from_u8` has a breaking change: instead of a
`panic!` when an invalid `u8` value is supplied, it returns an `Err(.)`. A new error type
has been added `InvalidMerkleTreeNodeType` for this case and a new variant of
`OxenError::MerkleTreeError` has been added that wraps this error. Call sites have been
updated to propagate this error up to an `OxenError`.

### DIFF
--- a/crates/lib/src/model/merkle_tree/node_type.rs
+++ b/crates/lib/src/model/merkle_tree/node_type.rs
@@ -90,12 +90,6 @@ impl MerkleTreeNodeType {
     }
 
     /// Deserialize a `u8` value into a `MerkleTreeNodeType`.
-    /// Panics if the `u8` value is not a valid `MerkleTreeNodeType`.
-    pub fn from_u8_unwrap(val: u8) -> MerkleTreeNodeType {
-        Self::from_u8(val).expect("Invalid MerkleTreeNodeType: {val}")
-    }
-
-    /// Deserialize a `u8` value into a `MerkleTreeNodeType`.
     /// This function is 1:1 with `to_u8`: all outputs from `to_u8` result in an `Ok`.
     pub fn from_u8(val: u8) -> Result<MerkleTreeNodeType, InvalidMerkleTreeNodeType> {
         match val {
--- a/crates/lib/src/model/merkle_tree/node_type.rs
+++ b/crates/lib/src/model/merkle_tree/node_type.rs
@@ -78,6 +78,7 @@ impl MerkleTreeNodeType {
     */
 
     /// Serialize the node type into a stable `u8` value.
+    /// This function is 1:1 with `from_u8`.
     pub fn to_u8(&self) -> u8 {
         match self {
             MerkleTreeNodeType::Commit => 0u8,
@@ -91,10 +92,11 @@ impl MerkleTreeNodeType {
     /// Deserialize a `u8` value into a `MerkleTreeNodeType`.
     /// Panics if the `u8` value is not a valid `MerkleTreeNodeType`.
     pub fn from_u8_unwrap(val: u8) -> MerkleTreeNodeType {
-        from_u8(val).expect("Invalid MerkleTreeNodeType: {val}")
+        Self::from_u8(val).expect("Invalid MerkleTreeNodeType: {val}")
     }
 
     /// Deserialize a `u8` value into a `MerkleTreeNodeType`.
+    /// This function is 1:1 with `to_u8`: all outputs from `to_u8` result in an `Ok`.
     pub fn from_u8(val: u8) -> Result<MerkleTreeNodeType, InvalidMerkleTreeNodeType> {
         match val {
             0u8 => Ok(MerkleTreeNodeType::Commit),
@@ -154,7 +156,10 @@ impl<T: Serialize + MerkleTreeNodeIdType + Debug + Display> TMerkleTreeNode for 
 
 #[cfg(test)]
 mod tests {
-    use crate::model::merkle_tree::node::{CommitNode, DirNode, FileChunkNode, FileNode, VNode};
+    use crate::{
+        error::OxenError,
+        model::merkle_tree::node::{CommitNode, DirNode, FileChunkNode, FileNode, VNode},
+    };
 
     use super::*;
 
@@ -162,7 +167,11 @@ mod tests {
     fn test_nodes_implement_trait() {
         /// this only exists so we can check that all node types implement `TMerkleTreeNode`
         /// it will be a compile-time error if a node type does not implement the trait
-        fn is_tmerkletreenode<T: TMerkleTreeNode>(x: T) {}
+        fn is_tmerkletreenode<T: TMerkleTreeNode>(_: T)
+        where
+            OxenError: From<T::SerializationError>,
+        {
+        }
 
         is_tmerkletreenode(CommitNode::default());
         is_tmerkletreenode(DirNode::default());
--- a/crates/lib/src/model/merkle_tree/node_type.rs
+++ b/crates/lib/src/model/merkle_tree/node_type.rs
@@ -58,4 +58,23 @@ pub trait MerkleTreeNodeIdType {
     fn hash(&self) -> MerkleHash;
 }
 
-pub trait TMerkleTreeNode: MerkleTreeNodeIdType + Serialize + Debug + Display {}
+/// Trait for things that are Merkle tree nodes.
+///
+/// Critical functionality is to be able to serialize & deserialize nodes. The Debug & Display
+/// constraints are for convenience.
+///
+/// Since this trait is used as a parameter for generic functions, it must be object-safe. This means
+/// that it cannot extend `Serialize` since that has a `<S: Serializer>` parameter on the
+/// `serialize` method, which causes the type to not be implementable as a vtable lookup.
+///
+/// This is why the trait contains manual serialization (`to_msgpack_bytes`) and deserialization
+/// (`from_msgpack_bytes`) methods. For types that do implement `Serialize` and `Deserialize`,
+/// there's a blanket implementation that delegates the appropriate `serialize` & `deserialze`
+/// calls to these trait methods.
+pub trait TMerkleTreeNode: MerkleTreeNodeIdType + Debug + Display {
+    /// Serialize this node to a MsgPack byte array.
+    fn to_msgpack_bytes(&self) -> Vec<u8>;
+
+    /// Deserialize a node from a MsgPack byte array.
+    fn from_msgpack_bytes(bytes: &[u8]) -> Self;
+}
--- a/crates/lib/src/model/merkle_tree/node_type.rs
+++ b/crates/lib/src/model/merkle_tree/node_type.rs
@@ -110,7 +110,7 @@ impl MerkleTreeNodeType {
 }
 
 /// Allows for types to identify themselves as a specific kind of merkle tree node via the
-/// `MerkleTreeNodeType` enum. This is ciritical for defining the node's stable on-disk
+/// `MerkleTreeNodeType` enum. This is critical for defining the node's stable on-disk
 /// byte representation.
 pub trait MerkleTreeNodeIdType {
     /// The type of merkle tree node this node can serialize & deserialize into.
--- a/crates/lib/src/model/merkle_tree/node_type.rs
+++ b/crates/lib/src/model/merkle_tree/node_type.rs
@@ -14,13 +14,28 @@ use std::{
     fmt::{Debug, Display},
     hash::{Hash, Hasher},
 };
+use thiserror::Error;
 
+/// The type of node that we are storing in the merkle tree.
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Copy)]
 pub enum MerkleTreeNodeType {
+    /// A commit in the repository. From this commit node, we can traverse the tree to recover the
+    /// entire repository state at the commit.
     Commit,
+
+    /// A file in the repository.
     File,
+
+    /// A directory in the repository.
     Dir,
+
+    /// Directories can be very large, so we split their contents into (potentially) multiple
+    /// virtual directory nodes: a `VNode`. A `VNode` is like a directory: it contains some
+    /// number of files and directories. It is always a subset of a real directory. If a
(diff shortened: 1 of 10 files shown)


crates/lib/src/repositories/tree.rs
```diff
@@ -1077,11 +1078,35 @@ pub fn write_tree(repo: &LocalRepository, node: &MerkleTreeNode) -> Result<(), O
...
+    VNode: TMerkleTreeNode<SerializationError = S>,
+    DirNode: TMerkleTreeNode<SerializationError = S>,
+    FileNode: TMerkleTreeNode<SerializationError = S>,
+    // NOTE:
+    // We could have dropped everything here, have no `S` generic, and have this in the where:
+    //      N: TMerkleTreeNode<SerializationError = rmp_serde::encode::Error>,
+    // But, by doing it this way, we make it so that we can change the actual SerializationError
+    // type without needing to change this function's type. It works so long as the error type
+    // for all of the node's trait implementations align.
+{
```
CleanCut: The bounds complexity here should mostly (completely?) go away with the elimination of `SerializationError`.

<details>
<summary>We wouldn't need the `S` generic anyway, but that's irrelevant now.</summary>

My original comment before I saw that we could eliminate `SerializationError`:

Go ahead and drop the `S` generic and simplify this.

The extra bounds complexity is unnecessary. All nodes share the same `SerializationError` type already, and if we change it, we'll change it everywhere, including here.
</details>

crates/lib/src/model/merkle_tree/node_type.rs
```diff
@@ -41,21 +89,94 @@ impl MerkleTreeNodeType {
...
+            2u8 => Ok(MerkleTreeNodeType::VNode),
+            3u8 => Ok(MerkleTreeNodeType::File),
+            4u8 => Ok(MerkleTreeNodeType::FileChunk),
+            _ => Err(InvalidMerkleTreeNodeType(val)),
         }
     }
 }
 
+/// Allows for types to identify themselves as a specific kind of merkle tree node via the
+/// `MerkleTreeNodeType` enum. This is ciritical for defining the node's stable on-disk
```
CleanCut:
```suggestion
/// `MerkleTreeNodeType` enum. This is critical for defining the node's stable on-disk
```

crates/lib/src/model/merkle_tree/node_type.rs
```diff
@@ -41,21 +89,94 @@ impl MerkleTreeNodeType {
         }
     }
 
-    pub fn from_u8(val: u8) -> MerkleTreeNodeType {
+    /// Deserialize a `u8` value into a `MerkleTreeNodeType`.
+    /// Panics if the `u8` value is not a valid `MerkleTreeNodeType`.
+    pub fn from_u8_unwrap(val: u8) -> MerkleTreeNodeType {
+        Self::from_u8(val).expect("Invalid MerkleTreeNodeType: {val}")
+    }
+
```
CleanCut: Use `from_u8` instead. Returning an error is the correct thing to do in every place that this panicky function is used.

```suggestion
```

crates/lib/src/model/merkle_tree/node_type.rs
```diff
@@ -30,7 +45,40 @@ impl Hash for MerkleTreeNodeType {
     }
 }
 
+#[derive(Debug, Error)]
+#[error("Deserialization failure: Invalid MerkleTreeNodeType: {0}")]
+/// Failed to deserialize a `MerkleTreeNodeType` due to unknown `u8` value.
+pub struct InvalidMerkleTreeNodeType(u8);
```
CleanCut, line 51: This should be a new `OxenError` variant.

crates/lib/src/model/merkle_tree/node_type.rs
```diff
@@ -41,21 +89,94 @@ impl MerkleTreeNodeType {
...
+///
+/// This is why the trait contains a manual serialization (`to_msgpack_bytes`) method. For types
+/// that do implement `Serialize`, there's a blanket implementation that delegates the `serialize`
+/// call to the `to_msgpack_bytes` method.
+pub trait TMerkleTreeNode: MerkleTreeNodeIdType + Debug + Display {
+    /// The error type that can occur during serialization.
+    type SerializationError: std::error::Error + Send + Sync + 'static;
+
+    /// Serialize this node to a MsgPack byte array.
+    fn to_msgpack_bytes(&self) -> Result<Vec<u8>, Self::SerializationError>;
```
CleanCut: We don't need an associated type here. Use `OxenError`. That should help eliminate a lot of the extra complexity!

```suggestion
    /// Serialize this node to a MsgPack byte array.
    fn to_msgpack_bytes(&self) -> Result<Vec<u8>, OxenError>;
```

CleanCut (commented):
We can reduce the complexity here quite a bit and still accomplish the goal!

crates/lib/src/repositories/tree.rs
malcolmgreaves: This is an intentional part of the design. I don't want to overly rely on using `OxenError`, because that is a large enum where nearly all of the variants are not applicable to the code. There's only a single error -- serialization problem.

An explicit design I want to do more of is keep code as locally scoped as possible. Here, that meant only using a single serialization error type.

There's one way to simplify this, and that's to hardcode the serialization error type as a serde error. However, given the way I've documented this and explained it, and used names like "`SerializationError`", I feel that this is clear. The benefit is it gives the code the ability to switch to a new `SerializationError` without requiring this code to change. Which is good, since since this code is agnostic to the serialization details! :)

crates/lib/src/model/merkle_tree/node_type.rs
malcolmgreaves: Yeah I would like to make this change, that's why I'm introducing it here :) I thought I got all of these -- let me double check the call sites.

crates/lib/src/model/merkle_tree/node_type.rs
malcolmgreaves: There's a variant that wraps this. Keeping it separate is intentional -- it makes it clearer that code is only returning this as an error versus every single possible variant of an `OxenError`. It's used in the to/from u8 for the merkle tree node type. That code should only ever return this kind of error. I want to make that constraint clear in the types, hence why it's a standalone `struct` error here and there's a `#[from]` wrapper in `OxenError` for it.

crates/lib/src/model/merkle_tree/node_type.rs
malcolmgreaves: Explicitly not using `OxenError`. (_I explained this in an earlier comment above, but I'll reiterate a bit here_) That is too wide of a type -- it includes too many variants that should never be used here. Using it as the error type is confusing and allows the code to return invalid errors. It's only ever valid for it to return an error related to serialization.

This design is actually simpler if we stop to think about what all of this syntax is saying -- it's saying "this function only returns `Vec<u8>` or some single kind of error."

One issue with the oxen code today is it's very hard to understand what exact kind of error some piece of code returns if it has a `Result<_, OxenError>`. With this design, it's straightforward to see what specific error type the code returns by only looking at the **signature** (instead of reading through the entire definition and tracking down all of the possible variants of `OxenError` that are returned).

In turn, this design makes all of the concrete implementation code simpler: they all use `rmp_serde::decode::Error` instead of having `OxenError`. So we can look at those signatures and see "ok, the only thing that can go wrong is a serialization error" instead of "it can be any variant of an `OxenError`, which one(s) is it actually going to be?"

crates/lib/src/model/merkle_tree/node_type.rs
```diff
@@ -41,21 +89,94 @@ impl MerkleTreeNodeType {
...
+    /// Serialization errors come from serde.
+    type SerializationError = rmp_serde::encode::Error;
+
+    /// Uses serde to serialize this node.
+    fn to_msgpack_bytes(&self) -> Result<Vec<u8>, Self::SerializationError> {
+        let mut buf = Vec::new();
+        let x = self.serialize(&mut rmp_serde::Serializer::new(&mut buf));
+        x.map(|_| buf)
+    }
+}
```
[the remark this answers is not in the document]
malcolmgreaves: Good point in general, but we're ok making these kinds of breaking changes.

crates/lib/src/repositories/tree.rs
CleanCut: I understand this was a deliberate design choice. My feedback is that it is not a good direction to go in.

>  I don't want to overly rely on using OxenError, because that is a large enum where nearly all of the variants are not applicable to the code.

`OxenError` exists so we have a consistent error type to use throughout the `liboxen` codebase. In other words, it is intended that we use `OxenError` throughout `liboxen`.  In this context, there is no overly relying on it. I'd encourage us to lean into using it rather than working around it. There is nothing wrong with the error enum having a large number of variants. Fragmenting our error approach at this point is the wrong direction to go in.

This code will still be agnostic to serialization details using `OxenError`.

Hardcoding an error type specific to a single implementation into a trait is not a simplification.

crates/lib/src/model/merkle_tree/node_type.rs
CleanCut: Again, I understand your choices are intentional. My feedback is that there is a better approach.

> it makes it clearer that code is only returning this as an error versus every single possible variant of an OxenError...

My comment from before about leaning into `OxenError` and many variants being okay applies here.

> That code should only ever return this kind of error. I want to make that constraint clear in the types...

I don't know a good reason to assert that unknown implementations should only return one type of error.

Even if all implementations did return one type of error, or even this specific error, we're already following the industry best practice of using a single enum error type for our `liboxen` library. I don't see a reason to fragment our error handling approach.

crates/lib/src/model/merkle_tree/node_type.rs
CleanCut: Having a large number of variants in the error enum isn't a problem. It's [pretty normal](https://doc.rust-lang.org/std/io/enum.ErrorKind.html). It isn't clear to me why you find this confusing. On the other hand, having to search for scattered errors around the codebase and figure out when and where to use them would quickly become confusing.

The argument for dividing our single error enum into two error enums would be that there are two disjoint sets of errors that are used in mutually exclusive call stacks. We don't even have that situation.

> One issue with the oxen code today is it's very hard to understand what exact kind of error some piece of code returns if it has a Result<_, OxenError>.

I don't think that is an issue. My observation has been that almost nothing that handles an error cares which specific error it was. In the few places where we do care, we match on the OxenError variant that we care about. Where's the issue? It's already straightforward to see the variant type. Isn't that why we've been replacing the `BasicString` variants with more specific variants, at your request--just in case someone cares in the future and wants to match on the variant?

> In turn, this design makes all of the concrete implementation code simpler: they all use `rmp_serde::decode::Error` instead of having `OxenError`.

Your design makes the concrete implementation more complex. All the bounds stuff goes away when you switch to `OxenError`.

malcolmgreaves:
Made one sweeping change that hopefully simplifies things a bit for this review. Instead of having a generic `SerializationError` for the `trait`, I've hardcoded it to be a serde serialization error (`rmp_serde::encode::Error`). We lose the ability to change the error type w/o changing the trait, but after thinking on it a bit more, I feel like this is an ok tradeoff. If we do change how we serialize nodes into bytes, we will need to create a migration as the on-disk format will change.

This cuts down on a lot of the syntax -- there's no more `where` clauses constraining the `SerailizationError` to be convertible into an `OxenError` nor the constraint that these error types have to align in `p_write_tree`. The `OxenError::RmpEncodeError` already takes care of converting the serialization error into an `OxenError`.

I also made the change to use the new `from_u8` function that returns an error instead of `panic!`ing on an error. This has its own `OxenError::MerkleTreeError` and conversion from the more specific error type that is returned.

crates/lib/src/model/merkle_tree/node_type.rs
```diff
@@ -41,21 +89,94 @@ impl MerkleTreeNodeType {
         }
     }
 
-    pub fn from_u8(val: u8) -> MerkleTreeNodeType {
+    /// Deserialize a `u8` value into a `MerkleTreeNodeType`.
+    /// Panics if the `u8` value is not a valid `MerkleTreeNodeType`.
+    pub fn from_u8_unwrap(val: u8) -> MerkleTreeNodeType {
+        Self::from_u8(val).expect("Invalid MerkleTreeNodeType: {val}")
+    }
```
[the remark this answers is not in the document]
malcolmgreaves: Good catch! Would implement this but I decided to delete the old "from_u8" behavior in this PR.

CleanCut (approved):

PR #407, opened 2026-04-01
`get_by_name()` to return `Result<Branch, OxenError>` instead of `Result<Option<Branch>, OxenError>`

subygan:
There was a lot of boiler plate that we were handling None to return `OxenError::repo_not_found()` this PR simplifies that.

Important change is this:
https://github.com/Oxen-AI/Oxen/pull/407/changes#diff-36d674fb2bcad4130c6b6fff3f561bbcbfe2a28ed81a632f375c2ae7b9dbf442R26

### DIFF
--- a/crates/lib/src/repositories/branches.rs
+++ b/crates/lib/src/repositories/branches.rs
@@ -52,9 +52,15 @@ pub fn get_commit_id(repo: &LocalRepository, name: &str) -> Result<Option<String
     with_ref_manager(repo, |manager| manager.get_commit_id_for_branch(name))
 }
 
-/// Check if a branch exists
-pub fn exists(repo: &LocalRepository, name: &str) -> bool {
-    get_by_name(repo, name).is_ok()
+/// Check if a branch exists.
+/// Returns Ok(false) if the branch doesn't exist.
+/// Returns Err() if there was some other problem accessing the local repository.
+pub fn exists(repo: &LocalRepository, name: &str) -> Result<bool, OxenError> {
+    match get_by_name(repo, name) {
+        Ok(_) => Ok(true),
+        Err(OxenError::BranchNotFound(_)) => Ok(false),
+        Err(e) => Err(e),
+    }
 }
 
 /// Get the current branch
--- a/crates/lib/src/repositories/branches.rs
+++ b/crates/lib/src/repositories/branches.rs
@@ -22,8 +22,13 @@ pub fn list_with_commits(repo: &LocalRepository) -> Result<Vec<(Branch, Commit)>
     with_ref_manager(repo, |manager| manager.list_branches_with_commits())
 }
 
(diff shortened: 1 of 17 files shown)


crates/lib/src/repositories/checkout.rs
```diff
@@ -23,7 +23,7 @@ pub async fn checkout(
     if repositories::branches::exists(repo, value)? {
         if repositories::branches::is_checked_out(repo, value) {
             println!("Already on branch {value}");
-            return repositories::branches::get_by_name(repo, value);
+            return Ok(Some(repositories::branches::get_by_name(repo, value)?));
```
subygan: this is a little awkward but changing `checkout()` to return a `Result<Branch, OxenError>` is a bigger refactor

CleanCut (commented):
`get_by_name` is a good improvement!

The old form (and the current `get_by_name_maybe`) provides no value. Anything that uses `get_by_name_maybe` should either change to use `get_by_name` to be simpler, or use `with_ref_manager` directly otherwise.

subygan:
`with_ref_manager` is a little lower level access, which was I wanted to keep the `get_by_name_maybe` indirection to keep it clean.

I suggest just removing `get_by_name_maybe` and catch with where needed. wdyt?

```
    match repositories::branches::get_by_name(repo, name) {
        Ok(branch) => Ok(Some(branch)),
        Err(OxenError::BranchNotFound(_)) => Ok(None),
        Err(e) => Err(e),
    }
```

crates/lib/src/repositories/branches.rs
```diff
@@ -55,11 +53,8 @@ pub fn get_commit_id(repo: &LocalRepository, name: &str) -> Result<Option<String
 }
 
 /// Check if a branch exists
-pub fn exists(repo: &LocalRepository, name: &str) -> Result<bool, OxenError> {
-    match get_by_name(repo, name)? {
-        Some(_) => Ok(true),
-        None => Ok(false),
-    }
+pub fn exists(repo: &LocalRepository, name: &str) -> bool {
```
malcolmgreaves: I could also see this being:
```
match get_by_name(repo,name) {
  Ok(_) => Ok(true),
  Err(OxenError::BranchNotFound(_)) => Ok(false),
  error => error,
}
```

And keeping it as `Result<bool, OxenError>` and adding the following to the docstring:
```
/// Check if a branch exists. Returns Ok(false) if the branch doesn't exist. Returns Err() if there was some other problem accessing the local repository.
```

But a-ok to keep it as `-> bool` too. I guess if there's some other error we will encounter that somewhere else?

subygan: this is what I had implemented first but this makes sense I'll keep the Result<book, OxenError>

CleanCut (approved):

PR #410, opened 2026-04-02
Use streaming in S3 version of copy_version_to_path

CleanCut:
(Step 4 of the S3 implementation)

Fix `copy_version_to_path` to stream the S3 response body directly to the destination file instead of loading the whole file in memory.


malcolmgreaves (approved):
Great work! Nice use of a stream reader + `tokio::io::copy` to avoid holding it in-memory. One comment around buffered writing.

crates/lib/src/storage/s3.rs
```diff
@@ -492,17 +494,25 @@ impl VersionStore for S3VersionStore {
...
-            tokio::fs::create_dir_all(parent)
+            create_dir_all(parent)
                 .await
                 .map_err(|e| OxenError::basic_str(format!("Failed to create parent dirs: {e}")))?;
         }
-        tokio::fs::write(dest_path, &data)
+
+        let mut file = File::create(dest_path)
+            .await
+            .map_err(|e| OxenError::basic_str(format!("Failed to create file: {e}")))?;
```
malcolmgreaves: Could you please double check if this is doing buffered writing?

My memory may be wrong, but IIRC `tokio::fs::File` is unbuffered writes. If it's wrapped as a `BufWriter` and that's passed to `tokio::io::copy` then I believe it will  be using buffered writes.

Something like:
```rust
let file = File::create(...).await?;
let mut writer = tokio::io::BufWriter::new(file);

let mut stream = ...

tokio::io::copy(&mut stream, &mut writer).await?;
```

CleanCut: Thanks for the reminder to think about buffered I/O! Buffering behavior in Tokio is super tricky, so you've really got to be aware of all the pieces if you want to get optimal buffering.

- In this case, `tokio::io::copy` does it's own buffering. So technically, _yes_, we're doing buffered writing here. _But_ `tokio::io::copy` uses an 8KB buffer, which doesn't do much as far as buffering goes when we're dealing with a large file that (presumably) already has more than 8KB available. We can do better.
- If we wrap our file in `BufWriter`...essentially nothing changes. `tokio::io::copy` ignores `BufWriter`'s buffer, but...
- We can switch to `tokio::io::copy_buf` that uses `BufWriter`'s buffer instead of allocating it's own buffer! Unfortunately, `BufWriter`'s default buffer is _also_ 8KB, which doesn't change anything, but...
- If we configure `BufWriter` with a larger buffer and use `tokio::io::copy_buf` , then we can significantly decrease the amount of syscalls we're doing!

So I pumped it up to 10MB: 😄 

```rust
let file = ...
let mut writer = tokio::io::BufWriter::with_capacity(10 * 1024 * 1024, file);
let mut stream = ...
tokio::io::copy_buf(...).await?;
```

malcolmgreaves: Wonderful! Thanks for explaining how this is working. I wasn't aware of `tokio::io::copy`'s internal buffering. I'm glad to know that we can use and configure a `BufWriter` to get precise control over buffering behavior.

PR #411, opened 2026-04-03
bugfix: properly remove nested directories

malcolmgreaves:
Oxen was not correctly removing nested directories from the Merkle tree when committing.
The root cause was mixing up paths: nodes contain file names but things like the dir cache
keep repository-relative paths. This PR correctly fixes some mixups here. When using nodes,
relative paths are reconstructed by walking the path to that node from the root. Additionally,
there was a related bug where `oxen rm -r` was not removing nested directories. This required
a similar fix to subbing-in the full path so that comparisons in the rest of the codebase worked
as expected.

This PR also deletes the removed directories from the `dir_hash_db` at the end of
`fn write_commit_entries`. These were stale cache entries that were causing successive
reads of the Merkle tree to incorrectly show deleted dirs.

Extensive new tests have been added to illuminate this bug and prevent it from cropping up again.


experiments/todo_plans/relative_path_type_safety_plan.md
malcolmgreaves, line 1: I started on, but then decided it was too large of a scope for a bugfix, a plan to introduce new types for repo-relative paths vs. file names vs. full filesystem paths. I believe that having this in the type system will let us avoid this kind of bug in the future. I put the plan down here and will implement it in a follow up.

CleanCut: The `experiments` directory appears to have experimental Rust projects rather than documentation. Where do you think would be a good place to put this? Maybe `plans/` or `docs/plans` or `todos/` directory, or something like that?

crates/lib/src/repositories/commits/commit_writer.rs
```diff
@@ -829,6 +852,30 @@ fn write_commit_entries(
...
+        for child in removed_children {
+            if child.status == StagedEntryStatus::Removed
+                && let EMerkleTreeNode::Directory(_) = &child.node.node
+                && let Ok(child_path) = child.node.maybe_path() {
+                    let path_str = child_path.to_string_lossy();
+                    log::debug!("deleting removed dir hash: {path_str:?}");
+                    str_val_db::delete(dir_hash_db, path_str)?;
+                }
+        }
+    }
```
CleanCut, line 843: **Suggestion:** Put the logic from `cach_invalidate_dir_hash_db` inline. Discussion: One-use functions are usually a code smell. Here we're basically adding a dozen extra lines and comments to partition the end of the previous function into a separate function without gaining any benefit. I've been [trying to add instructions to CLAUDE.md to get it to stop doing that](https://github.com/Oxen-AI/Oxen/blob/c989dcba21ec1f51a57836e54dbe5c70dd8cda93/.claude/CLAUDE.md#L110), but it looks like I'll need to tweak it.

crates/lib/src/repositories/add.rs
```diff
@@ -665,4 +671,321 @@ A: Oxen.ai
         })
         .await
     }
+
+    /// Given the relative path (posibly nested) to a file and contents it should have,
```
CleanCut:
```suggestion
    /// Given the relative path (possibly nested) to a file and contents it should have,
```

crates/lib/src/repositories/add.rs
```diff
@@ -665,4 +671,321 @@ A: Oxen.ai
...
+        );
+    }
+
+    use async_walkdir::WalkDir;
+
+    /// Checks that only the expected relative paths exist from the given root,
+    /// ignoring the '.oxen/' folder.
+    async fn expect_fs(root: &Path, expected_relative: &[&Path]) {
+        expect_filesystem(root, expected_relative, &[".oxen"]).await
+    }
```
CleanCut: **Suggestion:** Call `expect_filesystem` directly. The 5 lines above exist only to provide the literal `&[".oxen"]`

experiments/todo_plans/relative_path_type_safety_plan.md
```diff
@@ -0,0 +1,315 @@
+# Relative Path Type Safety Plan
+
+## Context
+
+The Oxen codebase uses raw `PathBuf` for paths that carry three fundamentally different semantics:
+- **Absolute filesystem paths** — e.g., `/home/user/repo/1/2/3`
+- **Paths relative to the repo root** — e.g., `1/2/3/file.txt`
```
CleanCut: There are at least two sub-cases for this:
- Paths of checked out version files
- Paths of internal files `.oxen/...`

CleanCut (approved):
⚠️ I advise avoiding adding any `unwrap` calls and removing any that you touch. In general, we should eliminate all the `.unwrap()` calls in our code, except tests. Panics introduce risks of production downtime and data corruption. The particular usage here looks like it's _probably_ not going to cause a problem, but there's just no good reason to use them.

Otherwise, LGTM though I didn't dig in and learn the semantics of the operation very deeply, so it would be great if Greg could take a look as well.

crates/lib/src/repositories/commits/commit_writer.rs
malcolmgreaves: I find that encapsulating the logic in a function makes it easier to understand what it is supposed to be doing. I also did this in the event that there is another place where we need to invalidate the cache we can reuse the function.

experiments/todo_plans/relative_path_type_safety_plan.md
malcolmgreaves: Having a convention for where we put plans, and other kinds of documents, makes sense to me from a long term perspective. 

Though, I was erring on the side on not introducing a new convention here. This is a somewhat temporary file. It will eventually be removed when the idea is implemented. So we may quickly end up with an empty directory again. Does it make sense to invest in a convention, teach and document it, if it's not permanent? 

Perhaps we could invest in some documentation to make the convention more permanent. I think that's a good idea for codebase organization. However, I don't think it's the priority right now. I'd like to wait and see if there's another naturally-arising situation where someone needs to document a plan. Once there's a pattern of us needing this, I'd like to pull the trigger on adding some organizational conventions.

subygan:
@malcolmgreaves  I posted this in slack must've gotten missed,

>This PR is awesome but fixes it at the add & commit layer.
>
> the bug is actuall at the rm layer. so that would still persist. for example,
> in a directory like, `1/2/3/file.txt`
> `rm -rf 1/2/3`
> `oxen rm 1/2/3`
> would still fail to stage that directory properly. similar to how `oxen add .` failed.

I also created an example of a simpler fix to this issue here,

https://github.com/Oxen-AI/Oxen/pull/417/changes

malcolmgreaves:
Moved the "plan" out from this PR @CleanCut 

Integrated your change and test case @subygan . There's a 2nd related bug here that needs a similar fix, so I pulled yours changes in here and merged.

PR #413, opened 2026-04-03
mut unneeded for RmOpts construction

malcolmgreaves:
`RmOpts` allows for setting all fields at construction time. Several
places in the code were unnecessarily creating a `mut RmOpts` to set
fields right after construction time instead of directly setting the
`struct` directly during construction and using a non-`mut` declaration.

### DIFF
--- a/crates/lib/src/api/client/diff.rs
+++ b/crates/lib/src/api/client/diff.rs
@@ -1135,8 +1135,11 @@ who won the game?,The packers beat up on the bears,packers
                 util::fs::remove_file(&repo_filepath)?;
             }
 
-            let mut rm_opts = RmOpts::from_path(Path::new("images").join("cats"));
-            rm_opts.recursive = true;
+            let rm_opts = RmOpts {
+                path: PathBuf::from("images").join("cats"),
+                staged: false,
+                recursive: true,
+            };
             repositories::rm(&repo, &rm_opts)?;
             repositories::commit(&repo, "Removing cat images")?;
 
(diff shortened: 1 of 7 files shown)


crates/lib/src/api/client/diff.rs
```diff
@@ -1135,8 +1135,11 @@ who won the game?,The packers beat up on the bears,packers
                 util::fs::remove_file(&repo_filepath)?;
             }
 
-            let mut rm_opts = RmOpts::from_path(Path::new("images").join("cats"));
-            rm_opts.recursive = true;
+            let rm_opts = RmOpts {
+                path: PathBuf::from("images").join("cats"),
+                staged: false,
+                recursive: true,
+            };
```
malcolmgreaves, line 1142: Note to delete this and the other commented-out code in a follow-up PR.

crates/lib/src/core/v_latest/add.rs
```diff
@@ -153,8 +153,11 @@ pub async fn add_files(
     let gitignore = oxenignore::create(repo);
 
     let mut paths_to_remove = HashSet::new();
-    let mut rm_opts = RmOpts::new();
-    rm_opts.recursive = true;
+    let rm_opts = RmOpts {
+        path: PathBuf::from(""),
+        staged: false,
+        recursive: true,
+    };
```
CleanCut, line 160: Here's a cool Rust feature you can use here if you want to:

Whenever you want to customize one or more fields of struct, but leave the rest the default values, you can use _struct update syntax_.

The defaults for `path, staged, recursive` are `"", false, false` so in this block we're just customizing `recursive`:


```suggestion
    let rm_opts = RmOpts {
        recursive: true,
        ..Default::default()
    };
```

Note: You can't have a comma (`,`) after the `..Default::default()`--and that needs to be the last thing in the list.

CleanCut (approved):

crates/lib/src/core/v_latest/add.rs
malcolmgreaves: Follow up: https://github.com/Oxen-AI/Oxen/pull/414

PR #415, opened 2026-04-03
Make clippy pre-commit hook also run on test code

malcolmgreaves:
Makes the pre-commit hooks for `cargo clippy` run on all targets: crates, tests,
and benchmarking code are all included now.

Applies all suggestions from `cargo clippy` on the rest of the codebase.

Also makes the test user config have the same name that bin/test-rust produces.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
 
       - id: cargo-clippy
         name: cargo clippy
-        entry: bash -c 'cargo clippy --workspace --no-deps -- -D warnings'
+        entry: bash -c 'cargo clippy --workspace --all-targets --no-deps -- -D warnings'
         language: system
         files: \.rs$
         pass_filenames: false
--- a/crates/lib/src/api/client/tree.rs
+++ b/crates/lib/src/api/client/tree.rs
@@ -535,10 +535,10 @@ mod tests {
             let dir_count = entries
                 .filter_map(|entry| match entry {
                     Ok(e) => {
-                        if let Ok(file_type) = e.file_type() {
-                            if file_type.is_dir() {
-                                return Some(1);
-                            }
+                        if let Ok(file_type) = e.file_type()
+                            && file_type.is_dir()
+                        {
+                            return Some(1);
                         }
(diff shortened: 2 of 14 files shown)


.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
 
       - id: cargo-clippy
         name: cargo clippy
-        entry: bash -c 'cargo clippy --workspace --no-deps -- -D warnings'
+        entry: bash -c 'cargo clippy --workspace --tests --no-deps -- -D warnings'
```
CleanCut: ⚠️ Okay, double-check me, but I believe that `--tests` is a _toggle_ that switches to _only_ checking tests. It's that way for a lot of commands, and it drives me freaking crazy. If it's doing that here, you need to run the command twice, once with `--tests` and once without.

CleanCut (approved):
⚠️ Mind the gap!

.pre-commit-config.yaml
malcolmgreaves: lemme double check !

CleanCut: Also, we _might_ want to use different settings for clippy for tests vs clippy for normal code.

For example: We probably don't ever want to use `.unwrap()` in application code, so it'd be good to have clippy disallow it. However, in tests...who cares? A crash just fails the test, and that's fine.

malcolmgreaves: Looking at `cargo clippy --help` it states:

> See all options with cargo check --help.

And `cargo check`'s documentation on `--tests` here https://doc.rust-lang.org/cargo/commands/cargo-check.html says:

> Check all targets that have the `test = true` manifest flag set. By default this includes the library and binaries built as unittests, and integration tests. Be aware that this will also build any required dependencies, so the lib target may be built twice (once as a unittest, and once as a dependency for binaries, integration tests, etc.). Targets may be enabled or disabled by setting the test flag in the manifest settings for the target.

So it does run against the non-test and test code!

But this was a very useful exercise, because I realized _why am I not using `--all-targets`_?

So I have changed this to use `--all-targets` :)

PR #423, opened 2026-04-09
Detect missing file data before pushing from a shallow clone

jcelliott:
A shallow clone has merkle tree metadata for all commits but only version store files for the fetched commits. When pushing to a new remote or force-pushing, the full history is walked and the server may request commits whose file data isn't available locally. This spot-checks one file per missing commit before starting uploads and gives a clear error directing the user to run `oxen pull --all`.

You can test by:
1. Create a new clone of a repo with multiple commits.
2. Creating a new remote and configure the repo with that remote
3. Push to the new remote

Before: You just get `No such file or directory`
After: You get a nice error message

### DIFF
--- a/crates/lib/src/core/v_latest/push.rs
+++ b/crates/lib/src/core/v_latest/push.rs
@@ -409,6 +409,25 @@ async fn push_commits(
         })
         .collect::<Result<Vec<(Commit, PushCommitInfo)>, OxenError>>()?;
 
+    // Spot-check that file data exists locally before pushing. A shallow clone
+    // has merkle tree metadata for all commits but only the version store files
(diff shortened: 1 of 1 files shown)


CleanCut (approved):
It looks like this does what you want to for the normal push path.

❓ Question: What about the `--missing files` and `--revalidate` codepaths? It looks like those are handled before the spot check. Do they need the same sort of spot check / instructional message?

jcelliott:
@CleanCut great question. Those are less likely to be an issue, because they're specifically debugging/repair tools. But we should make the experience nicer where we can. I'll add a follow-up PR.

PR #424, opened 2026-04-09
Fix push --missing-files from shallow clone

jcelliott:
- Fix panic when latest_remote_commit is None (pushing to a new remote with no commits). Use is_some_and() instead of unwrap().
- Add spot-check in list_and_push_missing_files for missing local file data, with a message directing the user to re-run from a full clone. The --revalidate path flows through the same function and is also covered.
- Only checking the first file to keep this lightweight. It's possible (but unlikely) that the first file exists but a later one doesn't. The behavior will still be correct, just a less friendly error message.

### DIFF
--- a/crates/lib/src/core/v_latest/push.rs
+++ b/crates/lib/src/core/v_latest/push.rs
@@ -215,7 +215,10 @@ async fn push_missing_files(
         let commit = repositories::commits::get_by_id(repo, commit_id)?
             .ok_or_else(|| OxenError::commit_id_does_not_exist(commit_id))?;
         list_and_push_missing_files(repo, remote_repo, None, &commit).await?;
-    } else if head_commit.id == latest_remote_commit.clone().unwrap().id {
+    } else if latest_remote_commit
+        .as_ref()
+        .is_some_and(|rc| head_commit.id == rc.id)
+    {
         //both remote and local are at same commit
 
         let history = repositories::commits::list_from(repo, &head_commit.id)?;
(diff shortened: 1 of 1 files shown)


CleanCut (approved):

crates/lib/src/core/v_latest/push.rs
```diff
@@ -244,6 +247,18 @@ async fn list_and_push_missing_files(
...
 
+    if let Some(entry) = missing_files.first() {
+        let version_store = repo.version_store()?;
+        if !version_store.version_exists(&entry.hash()).await? {
+            return Err(OxenError::basic_str(format!(
+                "Cannot push missing files for commit '{}' (\"{}\"): file data is not available locally.\n\
+                 This usually means the repository was cloned without full history.\n\
+                 To repair the remote, re-run this command from a clone that has the full history.",
+                head_commit.id, head_commit.message
+            )));
```
malcolmgreaves, line 258: Could you add this as a new variant to `OxenError` with this message instead of using it as an `OxenError::Basic`?

As just a place holder for the name, the structure isn't writing too much more code than using `basic_str`:
```rust
#[error("Cannot push missing files for commit '{commit_id}' (\"{commit_message}\"): file data is not available locally.\n\
                 This usually means the repository was cloned without full history.\n\
                 To repair the remote, re-run this command from a clone that has the full history.")
MissingFilesPush{commit_id: String, commit_message: String},
```
(_Maybe use `(String,String)` and use `{0}` and `{1}`, but I felt that 2 things of the same type are hard to distinguish._)


```suggestion
            return Err(OxenError::MissingFilesPush{
                commit_id: head_commit.id.clone(), 
                commit_message: head_commit.commit_message.clone(),
            });
```

malcolmgreaves (approved):
Great fix! One suggested change -- trying to reduce the use of `OxenError::Basic`. Proposal for a new variant for this case.

crates/lib/src/core/v_latest/push.rs
jcelliott: @malcolmgreaves I added a similar, but slightly different, message in the previous PR: https://github.com/Oxen-AI/Oxen/pull/423.
How would you handle those two variants? I think the context and potential fix is different enough to warrant separate messages even though they are similar.

jcelliott:
@malcolmgreaves I'll address the custom error in a follow-up.

PR #425, opened 2026-04-09
Ramdisk for tests

CleanCut:
Experiment with using a ramdisk, and the changes needed to enable doing that.

The hope here is that we can speed up tests, especially in CI. On my M5, I only saw a ~5% speedup. Hopefully the speedup in CI will be more significant, especially on Windows (which is by far the slowest).

I suspect that most of our test time is currently being eaten by sync code in oxen-server that's blocking the tokio runtime during concurrent tests, but that will take a lot of fixing to speed up, so I'll be more than happy if we can squeeze out 10% improvement with simple usage of a ramdisk.

- Rust changes: the standard library's `canonicalize` method doesn't work on ramdisks on Windows, which is where we need the ramdisk most (even without a ramdisk, macos and linux run tests more than twice as fast as Windows when Windows is on 4x beefier hardware). The Rust code changes work around this by using a custom implementation that has a fallback on Windows.

---

**Update**: Excellent results! ~11% speedup. From 13.5m to 12m.

image

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -94,6 +94,34 @@ jobs:
         with:
           tool: nextest@0.9.114
 
+      # Ramdisk for test data
+      - name: Setup ramdisk (Linux)
+        if: matrix.platform == 'linux'
+        run: |
+          RAMDISK_PATH=$(mktemp -d /dev/shm/oxen_test_XXXXXX)
+          echo "OXEN_TEST_RUN_DIR=$RAMDISK_PATH/test" >> "$GITHUB_ENV"
+          echo "SYNC_DIR=$RAMDISK_PATH/sync" >> "$GITHUB_ENV"
+          mkdir -p "$RAMDISK_PATH/test" "$RAMDISK_PATH/sync"
+
+      - name: Setup ramdisk (macOS)
+        if: matrix.platform == 'macos'
+        run: |
+          RAMDISK_DEV=$(hdiutil attach -nomount ram://8388608 | xargs)  # 4 GB
+          diskutil eraseDisk APFS OxenTest "$RAMDISK_DEV"
+          echo "OXEN_TEST_RUN_DIR=/Volumes/OxenTest/test" >> "$GITHUB_ENV"
+          echo "SYNC_DIR=/Volumes/OxenTest/sync" >> "$GITHUB_ENV"
+          mkdir -p /Volumes/OxenTest/test /Volumes/OxenTest/sync
+
+      - name: Setup ramdisk (Windows)
+        if: matrix.platform == 'windows'
+        run: |
+          choco install imdisk -y
+          imdisk -a -s 16G -m R: -p "/fs:ntfs /q /y"
+          echo "OXEN_TEST_RUN_DIR=R:\test" >> $env:GITHUB_ENV
+          echo "SYNC_DIR=R:\sync" >> $env:GITHUB_ENV
+          mkdir R:\test
+          mkdir R:\sync
+
       # Oxen Rust Tests
       - name: Setup test directories (Linux/macOS)
         if: matrix.platform != 'windows'
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -12,8 +12,8 @@ permissions:
   contents: read
 
 jobs:
-  test:
-    name: Test Suite
+  rust-test:
+    name: Rust Tests
     runs-on: ${{ matrix.os }}
     defaults:
       run:
@@ -73,7 +73,9 @@ jobs:
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
         with:
-          save-if: ${{ github.ref == 'refs/heads/main' }} # Only save cache on main; always restore.
+          prefix-key: rust-test
+          # TODO: Commented out to test caching. Uncomment before merging.
+          #          save-if: ${{ github.ref == 'refs/heads/main' }} # Only save cache on main; always restore.
           cache-all-crates: true
           cache-on-failure: true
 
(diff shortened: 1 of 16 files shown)


.github/workflows/ci_test.yml
```diff
@@ -89,6 +89,34 @@ jobs:
...
+      - name: Setup ramdisk (Windows)
+        if: matrix.platform == 'windows'
+        run: |
+          choco install imdisk -y
+          imdisk -a -s 16G -m R: -p "/fs:ntfs /q /y"
+          echo "OXEN_TEST_RUN_DIR=R:\test" >> $env:GITHUB_ENV
+          echo "SYNC_DIR=R:\sync" >> $env:GITHUB_ENV
+          mkdir R:\test
+          mkdir R:\sync
+
```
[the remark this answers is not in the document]
CleanCut: I'd rather start by just failing if it doesn't work.

crates/lib/src/util/fs.rs
```diff
@@ -1229,15 +1229,21 @@ pub fn is_canonical(path: impl AsRef<Path>) -> Result<bool, OxenError> {
...
-            "path {path:?} cannot be canonicalized due to err {err:?}"
-        ))),
+        Err(_) => {
+            // Fallback: convert to absolute path without symlink resolution.
+            if path.is_absolute() {
+                Ok(path.to_path_buf())
+            } else {
+                Ok(std::env::current_dir()?.join(path))
+            }
+        }
```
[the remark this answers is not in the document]
CleanCut: Good catch!

crates/lib/src/util/fs.rs
```diff
@@ -1937,6 +1943,66 @@ pub fn validate_and_normalize_path(path: impl AsRef<Path>) -> Result<PathBuf, Ox
...
+        } else {
+            if let Some(parent) = file_dst.parent() {
+                create_dir_all(parent)?;
+            }
+            file.unpack(&file_dst).await.map_err(|e| {
+                crate::error::OxenError::basic_str(format!(
+                    "Failed to unpack {}: {e}",
+                    file_dst.display()
+                ))
+            })?;
```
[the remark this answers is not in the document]
CleanCut: Good catch!

crates/server/src/controllers/commits.rs
```diff
@@ -1270,8 +1270,34 @@ async fn unpack_entry_tarball_async(
...
+                    }
+                    _ => continue,
+                }
+            }
+            if let Some(parent) = dest.parent() {
+                util::fs::create_dir_all(parent)?;
+            }
+            file.unpack(&dest)
                 .await
                 .map_err(|e| OxenError::basic_str(format!("Failed to unpack file: {e}")))?;
```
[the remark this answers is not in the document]
CleanCut: Good catch!

CleanCut:
Windows CI is failing on a bug in duckdb that was [fixed upstream](https://github.com/duckdb/duckdb/pull/21652), but not yet released. We need to wait for a duckdb release and the corresponding Rust wrapper release and then we can try again.

crates/lib/src/util/fs.rs
```diff
@@ -1003,14 +1003,30 @@ pub fn is_canonical(path: impl AsRef<Path>) -> Result<bool, OxenError> {
...
+                // to Uncategorized rather than Unsupported, so also check the raw code.
+                || (cfg!(windows) && e.raw_os_error() == Some(1)) =>
+        {
+            // Fallback: convert to absolute path without symlink resolution. This is needed on
+            // filesystems whose drivers don't implement canonicalization (e.g. Windows imdisk
+            // ramdisks).
+            if path.is_absolute() {
+                Ok(path.to_path_buf())
+            } else {
+                Ok(std::env::current_dir()?.join(path))
```
malcolmgreaves: Will this resolve `../` correctly? If e.g. `path = "../../../"`, will the resulting path preserve these `..`? I think that `join` just does a naive concatenation. 

If so, this can lead to a bug where we return `Ok` when we really should return `Err`. For example, if we have `path="../../../../../../../file.txt"` when the current dir is `/home/myuser/` this will silently return an invalid path.

I think it's important that we don't subvert the intention behind this `canonicalize` function. This PR proposes to make it not perform its original function in some special situations. To do this, I don't think it should still be thought of as a canonicalization function.

Do we actually need canonicalization in all cases where it's being used? Or would an absolute path suffice?

If you can verify that every place that is using `canonicalize` can actually use an absolute path, then let's change this function to return absolute paths and rename it. This way, the hack added here to work for ramdisks in the CI tests won't be something that changes semantics.

crates/lib/src/util/fs.rs
```diff
@@ -1664,6 +1680,66 @@ pub fn validate_and_normalize_path(path: impl AsRef<Path>) -> Result<PathBuf, Ox
     Ok(normalized)
 }
 
+/// Unpack an async-tar archive to a destination directory without calling `canonicalize`. This is
+/// needed because `archive.unpack()` and `entry.unpack_in()` internally call
+/// `std::fs::canonicalize`, which fails on filesystems that don't support it (e.g. Windows imdisk
+/// ramdisks). Path traversal is checked by rejecting parent components.
+pub async fn unpack_async_tar_archive<R: futures_util::AsyncRead + Unpin>(
```
malcolmgreaves, line 1687: Please avoid introducing new code that uses `OxenError::Basic`.

crates/server/src/controllers/commits.rs
```diff
@@ -1236,8 +1236,38 @@ async fn unpack_entry_tarball_async(
...
                 .await?;
         } else {
-            // For non-version files, unpack to hidden dir
-            file.unpack_in(&hidden_dir)
+            // We manually construct the path and check for traversal instead of using unpack_in(),
+            // because unpack_in() calls canonicalize() internally, which fails on Windows ramdisks
+            // (imdisk).
+            let entry_type = file.header().entry_type();
+            if !entry_type.is_file() && !entry_type.is_dir() {
+                return Err(OxenError::basic_str(format!(
```
malcolmgreaves, line 1244: Please avoid introducing new code that uses `OxenError::Basic`.

malcolmgreaves (commented):
I always like to see a PR that decreases CI times! 🚀  

In this PR though I have a few concerns with changing the semantics of `fn canonicalize` in order to eek out an improvement in our CI testing times. I have a more detailed in-line comment, but I'll summarize it as: can oxen get away with always using absolute paths where `canonicalize` is currently used? If so, then we can make that change + this ramdisk use to speed up our CI tests.

crates/lib/src/util/fs.rs
CleanCut: ✅ Good catch. I switched it to `std::path::absolute` instead of manual construction. That does resolve all relative items like `..` correctly.

Falling back to an absolute path only happens on windows (the `cfg!(windows)` and when the highly unusual error code that the imdisk driver returns is encountered (`e.raw_os_error() == Some(1)`). We know in this case that symlink resolution will not be needed. In all other cases, the path has been fully canonicalized, so we should be good to go.

crates/lib/src/util/fs.rs
CleanCut: ✅ Done

crates/server/src/controllers/commits.rs
CleanCut: ✅ Done

crates/lib/src/util/fs.rs
malcolmgreaves: Thanks for investigating this and making the `std::path::absolute` change!

malcolmgreaves (approved):

PR #430, opened 2026-04-11
Make metrics opt-in and gated by new feature flag

malcolmgreaves:
**Make Metrics Opt-In**
Makes starting the `/metrics` endpoint an explicit opt-in. Users must explicitly
set `OXEN_METRICS_PORT` for `oxen-server`'s `/metrics` to function.
`OXEN_METRICS_PORT=off` still disabled metrics, which is the default behavior now.

**`metrics` compilation feature**
All `metrics`-crate using features of all oxen crates are now gated behind a new
compilation feature flag called `"metrics"`. Collection in `liboxen`, `oxen`, and
`oxen-py` and serving in `oxen-server` is disabled in the default build.

The `production` flag includes `metrics`.

For `oxen-server`: if a user supplies `OXEN_METRICS_PORT` when `metrics` is not
enabled, then a warning is logged telling the user to recompile with `metrics` enabled.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -21,8 +21,9 @@ path = "src/main.rs"
 
 [features]
 default = []
+metrics = ["liboxen/metrics"]
 otel = ["liboxen/otel"]
-production = ["otel", "default"]
+production = ["metrics", "otel", "default"]
 
 [dependencies]
 async-trait = { workspace = true }
--- a/README.md
+++ b/README.md
@@ -261,8 +261,9 @@ tail -f ~/.oxen/logs/oxen-server.2026-04-06 | jq 'select(.level == "ERROR")'
 
 ## Prometheus Metrics
 
-`oxen-server` exposes a Prometheus-compatible metrics endpoint.
-See [Prometheus Metrics](crates/server/README.md#prometheus-metrics) for details.
+`oxen-server` can expose a Prometheus-compatible metrics endpoint. Requires
+the `metrics` compile-time feature (included in `production`) and `OXEN_METRICS_PORT`
+at runtime. See [Prometheus Metrics](crates/server/README.md#prometheus-metrics) for details.
 
 ## OpenTelemetry Tracing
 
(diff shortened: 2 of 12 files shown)


crates/lib/src/util/metrics.rs
CleanCut, line 1: It would be cleaner to have `#[cfg(feature = "metrics")]` above each `metrics::{...}!` macro call rather than maintaining custom macros that produce mock objects that will need more no-op methods added as we try to use more features of the `metrics` macros.

crates/cli/Cargo.toml
```diff
@@ -21,8 +21,9 @@ path = "src/main.rs"
 
 [features]
 default = []
+metrics = ["liboxen/metrics"]
```
CleanCut, line 24: Why do we need to provide a way to omit the metrics code?

CleanCut (approved):
This feels like unnecessary complexity to me. Is there some problem with having metrics functionality included by default?

crates/cli/Cargo.toml
malcolmgreaves: Happy to explain what I'm thinking here. If we weren't so performance-conscious, I would definitely say that we don't have any need for this extra compile-time complexity on this + telemetry.

Incorporating both of these features has some very small, but not-zero, overhead. There was an earlier effort in the codebase to add some profiling for performance testing and this is gated by a feature flag (`perf-logging`).

Going forward, I want to make sure we have the ability to strip-out any introspection-related code. And I'd like all of of this code to be opt-in.

crates/lib/src/util/metrics.rs
malcolmgreaves: Hmmm...initially, I was thinking that it'd be a bit too verbose to do it this way. Do you feel like it's actually easier to write 
```rust
#[cfg(feature = "metrics")]
metrics::counter("...").increment(1);
```

Rather than:
```rust
oxen_metrics_counter("...").increment(1);
```

If you feel like you + others won't get tired writing out the 2 lines each time then I can do a find-and-replace in this PR.

malcolmgreaves: I implemented this and I think you're right -- it's not that verbose and it does cut down on us maintaining a facade. 💯

PR #436, opened 2026-04-11
Make `list_files_in_dir` async

CleanCut:
Another step towards making all the things async.

- Convert `list_files_in_dir` to async
   -  Return `Result<Vec<PathBuf>, OxenError>` instead of silently ignoring errors
- Propagate the error through callers
  - Update `check_if_upload_complete_and_unpack` to also return `Result<(), OxenError>`
- Add unit tests for `list_files_in_dir`

### DIFF
--- a/crates/server/src/controllers/commits.rs
+++ b/crates/server/src/controllers/commits.rs
@@ -7,7 +7,6 @@ use liboxen::constants::VERSION_FILE_NAME;
 
 use liboxen::core::commit_sync_status;
 use liboxen::error::OxenError;
-use liboxen::error::StringError;
 use liboxen::model::{Commit, LocalRepository};
 use liboxen::opts::PaginateOpts;
 use liboxen::perf_guard;
@@ -881,10 +880,10 @@ async fn check_if_upload_complete_and_unpack(
             unpack_compressed_data(&files, repo).await?;
         } else {
             let filename = filename.ok_or_else(|| {
-                OxenError::MissingFileName(StringError::new(
+                OxenError::MissingFileName(
                     "check_if_upload_complete_and_unpack must supply filename if !compressed"
                         .into(),
-                ))
+                )
             })?;
             unpack_to_file(&files, repo, &filename)?;
         }
--- a/crates/server/src/controllers/commits.rs
+++ b/crates/server/src/controllers/commits.rs
@@ -797,7 +797,7 @@ pub async fn upload_chunk(
                     // but we should find a more elegant solution because we're
                     // doing a lot of extra work unpacking tarballs multiple
                     // times.
-                    check_if_upload_complete_and_unpack(
+                    if let Err(err) = check_if_upload_complete_and_unpack(
                         &repo,
                         tmp_dir,
                         total_chunks,
@@ -805,7 +805,12 @@ pub async fn upload_chunk(
                         query.is_compressed,
                         query.filename.to_owned(),
                     )
-                    .await;
+                    .await
+                    {
+                        log::error!("upload_chunk unpack failed: {err:?}");
+                        return Ok(HttpResponse::InternalServerError()
+                            .json(StatusMessage::internal_server_error()));
+                    }
 
                     Ok(HttpResponse::Ok().json(StatusMessage::resource_created()))
                 }
@@ -834,8 +839,8 @@ async fn check_if_upload_complete_and_unpack(
     total_size: usize,
     is_compressed: bool,
     filename: Option<String>,
-) {
-    let mut files = util::fs::list_files_in_dir(&tmp_dir);
+) -> Result<(), OxenError> {
+    let mut files = util::fs::list_files_in_dir(&tmp_dir).await?;
 
     log::debug!(
         "check_if_upload_complete_and_unpack checking if complete... {} / {}",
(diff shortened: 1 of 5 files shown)


crates/server/src/controllers/commits.rs
```diff
@@ -797,15 +797,20 @@ pub async fn upload_chunk(
...
                         total_chunks,
                         size,
                         query.is_compressed,
                         query.filename.to_owned(),
                     )
-                    .await;
+                    .await
+                    {
+                        log::error!("upload_chunk unpack failed: {err:?}");
+                        return Ok(HttpResponse::InternalServerError()
```
malcolmgreaves: IIRC we can automatically make an OxenHttpError from an OxenError. And if memory serves me right still, it will still be a 500 error. If this is the case, please use the .into conversion here.

crates/server/src/controllers/commits.rs
```diff
@@ -834,30 +839,24 @@ async fn check_if_upload_complete_and_unpack(
...
     log::debug!(
         "check_if_upload_complete_and_unpack checking if complete... {} / {}",
         files.len(),
         total_chunks
     );
 
-    if total_chunks < files.len() {
-        return;
+    if files.len() < total_chunks {
+        return Ok(());
```
malcolmgreaves, line 851: Is this right? This appears to be the opposite of what this function was doing before. Or is there something else going on that I'm not seeing?

crates/server/src/controllers/commits.rs
```diff
@@ -879,55 +877,27 @@ async fn check_if_upload_complete_and_unpack(
...
-                None => {
-                    log::error!(
-                        "check_if_upload_complete_and_unpack must supply filename if !compressed"
-                    );
-                }
-            }
+            let filename = filename.ok_or_else(|| {
+                OxenError::basic_str(
+                    "check_if_upload_complete_and_unpack must supply filename if !compressed",
+                )
```
malcolmgreaves: Please introduce a new error variant for this case.

malcolmgreaves (commented):
Generally this looks great! I have a few comments I'd like to see addressed. Some are merely seeking information/confirmation. Glad to see more asynchronous things be `async`!

crates/server/src/controllers/commits.rs
CleanCut: We are returning a `HttpResponse` here, not an `OxenHttpError`. There is no conversion from `OxenError` to `HttpResponse`.

crates/server/src/controllers/commits.rs
CleanCut: This is correct.

This was supposed to check and see if we have received all the parts of the file (`files.len()`) that we expect (`total_chunks`), and return early, so we don't waste time doing the full checking if we haven't even received all the parts. But since the logic was inverted, we always did the full checking, which didn't make the function incorrect, just wasteful.

crates/server/src/controllers/commits.rs
CleanCut: Could you explain your thought process on why we should add a new variant here?

As far as I'm aware, this doesn't meet any of the conditions that would call for a new variant:

- Need to handle the variant differently than other variants, for example
  - match on the variant
  - convert the variant to something different up the stack
  - display the variant differently
- Need to group foreign errors into a variant

We make [extensive use](https://github.com/Oxen-AI/Oxen/blob/0185ff738feb25355e720b68653d096624300813/crates/lib/src/error.rs#L560-L805) of the `Basic` variant.

Sidenote: I think the `Basic` variant would be better-named `InternalError`.

crates/server/src/controllers/commits.rs
malcolmgreaves: Oh interesting, I wasn't reading closely enough. Why is this an `Ok` if it's a 500 error? 🤔

I think this should instead be an `Err(..)` since we have an explicit failure.

crates/server/src/controllers/commits.rs
malcolmgreaves: Got it, thanks for explaining the context! If you have a moment, could you please create a short 1 or 2 sentence docstring for this function add this part to it? I'd like to make sure we preseve this important information on the correct use of this function :)

crates/server/src/controllers/commits.rs
```diff
@@ -879,55 +878,28 @@ async fn check_if_upload_complete_and_unpack(
...
-                },
-                None => {
-                    log::error!(
-                        "check_if_upload_complete_and_unpack must supply filename if !compressed"
-                    );
-                }
-            }
+            let filename = filename.ok_or_else(|| {
+                OxenError::MissingFileName(StringError::new(
+                    "check_if_upload_complete_and_unpack must supply filename if !compressed"
```
malcolmgreaves, line 883: (Very minor -- nothing requested for change here) IIRC there's a `From<&str> for StringError` so you should be able to just have `OxenError::MissingFileName("...".into())`.

crates/server/src/controllers/commits.rs
```diff
@@ -879,55 +878,28 @@ async fn check_if_upload_complete_and_unpack(
...
+                        .into(),
+                ))
+            })?;
+            unpack_to_file(&files, repo, &filename)?;
```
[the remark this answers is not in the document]
CleanCut: ✅ Good catch, fixed.

crates/server/src/controllers/commits.rs
CleanCut: ✅ Done

malcolmgreaves (approved):

crates/server/src/controllers/commits.rs
malcolmgreaves:
```suggestion
                        return Err(err.into())
```

CleanCut: ✅ Done

Issue #10, opened 2023-02-21
[Feature] Any chance to provide a windows build release?


leomeih:
A lot of assets needs to be preprocessed in windows platform because there're far more tools available.

Jasper-Bekkers:
> Yep we have a pull request with fixes for windows [here](https://github.com/Oxen-AI/Oxen/pull/94), will add a binary to the next release

For it all to work the follow-up is required too

gschoeni:
Windows build is integrated into our CI / release builds now, feel free to try it the binaries [here](https://github.com/Oxen-AI/Oxen/releases)

PR #437, opened 2026-04-12
Add raw PUT support and oxen-based-on optimistic concurrency

paul-hammant:
Two additions to the file endpoint:

**Raw PUT**: 
  
The PUT handler now accepts non-multipart requests where the HTTP body is the file content directly. Commit metadata comes from oxen-commit-author/email/message headers. The existing multipart PUT is unchanged.

**oxen-based-on header**: 

Optional optimistic concurrency for PUT and DELETE. Clients pass the oxen-revision-id (from GET) back as oxen-based-on. If the file has been modified since that revision, the server returns 400. If omitted, no check is performed.

 - 1 file changed, 6 new tests (all verify stored content round-trip)

This addresses item #4 on https://paulhammant.com/2020/01/19/vcs-nirvana/   That facillitates #10. 

As it happens. I have another PR that needs this one to be merged first.

### DIFF
--- a/crates/server/src/controllers/file.rs
+++ b/crates/server/src/controllers/file.rs
@@ -5,7 +5,8 @@ use crate::params::{app_data, parse_resource, path_param, query_param};
 use actix_multipart::form::text::Text;
 use actix_multipart::form::{FieldReader, Limits, MultipartForm};
 use actix_multipart::{Field, MultipartError};
-use actix_web::{HttpRequest, HttpResponse, web};
+use actix_web::{FromRequest, HttpMessage, HttpRequest, HttpResponse, web};
+use futures_util::StreamExt as _;
 use futures_util::TryStreamExt as _;
 use futures_util::future::LocalBoxFuture;
 use liboxen::core::staged::get_staged_db_manager;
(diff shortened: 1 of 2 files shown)


paul-hammant:
Anything left for Claude and I to respond to, @coderabbitai ?

paul-hammant:
We're inclined to thing that the 5 are resolved on the two open comments from you above:

## Finding 1 (in-memory buffering):

Already discussed — the multipart upload path buffers identically (line 88, FieldReader for MultipartTempFileNew). This is the pre-existing design pattern, not introduced by this  PR.
                                                                                                                                                                                         
##  Finding 2 (oxen-based-on for directory uploads):                

This is intentional. oxen-based-on provides optimistic concurrency for a specific file — it checks whether that file has been modified since the client last read it. For directory  uploads (multipart with files[]), the target is a directory, not a file, so there's no single file revision to validate against. The header is correctly ignored in this case. Per-file concurrency checking for batch directory uploads would be a different feature with different semantics.  

And thus we think we are condluded for remediation tasks for this PR, ahead of #OxenDevTeam review

paul-hammant:
This branch ressurected from stale [PUT_should_be_able_to_overwrite_a_HEAD_revision8 Pull request](https://github.com/Oxen-AI/Oxen/pull/416)

crates/server/src/controllers/file.rs
```diff
@@ -253,9 +254,102 @@ pub async fn get(
...
+            .await
+            .map_err(|e| OxenHttpError::BadRequest(format!("Multipart parse error: {e}").into()))?;
+        put_multipart(req, form.into_inner()).await
+    } else {
+        // Raw body PUT
+        let mut bytes = web::BytesMut::new();
+        while let Some(item) = payload.next().await {
+            bytes.extend_from_slice(&item.map_err(|_| OxenHttpError::FailedToReadRequestPayload)?);
+        }
+        put_raw(req, bytes.freeze()).await
```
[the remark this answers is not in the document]
malcolmgreaves: This is a great call-out -- we are very sensitive to performance and reliability regressions. We don't want to have anyone use this endpoint to upload a large file in one go. It's intentional to keep this as a multipart API only.

crates/server/src/controllers/file.rs
```diff
@@ -529,6 +672,40 @@ pub async fn mv(req: HttpRequest, body: String) -> actix_web::Result<HttpRespons
...
+        .and_then(|v| v.to_str().ok())
+    {
+        Some(v) => v,
+        None => return Ok(()), // No concurrency check requested
+    };
+
+    let node = repositories::tree::get_node_by_path(repo, commit, path)?;
+    let node = match node {
+        Some(n) if n.is_file() => n,
+        _ => return Ok(()), // File doesn't exist, treat as new file creation
```
malcolmgreaves, line 693: I believe this may be a bug. When `n` is an existing directory, this function seems like it treats it as a new file creation? I believe this should just be `Some(n) => n`, otherwise it seems like we don't get a chance to check if `"oxen-based-on"` matches for a directory.

I could be wrong here -- I can't recall from the top of my head right now if this endpoint is used when we're versioning directories. For verification, I'd like to see that this performs as expected if we're creating dir nodes or shown that we don't use directories in the PUT endpoint.

crates/server/src/controllers/file.rs
```diff
@@ -529,6 +672,40 @@ pub async fn mv(req: HttpRequest, body: String) -> actix_web::Result<HttpRespons
...
+    let claimed = match req
+        .headers()
+        .get("oxen-based-on")
+        .and_then(|v| v.to_str().ok())
+    {
+        Some(v) => v,
+        None => return Ok(()), // No concurrency check requested
+    };
+
+    let node = repositories::tree::get_node_by_path(repo, commit, path)?;
```
malcolmgreaves, line 690: This is actually a rather expensive operation, so we want to compute it less frequently if we can. This function is called from a place where it also gets the  `node` by its path, so a reference to it should be used here instead.

crates/server/src/controllers/file.rs
```diff
@@ -529,6 +672,40 @@ pub async fn mv(req: HttpRequest, body: String) -> actix_web::Result<HttpRespons
...
+    path: &Path,
+) -> Result<(), OxenHttpError> {
+    let claimed = match req
+        .headers()
+        .get("oxen-based-on")
+        .and_then(|v| v.to_str().ok())
+    {
+        Some(v) => v,
+        None => return Ok(()), // No concurrency check requested
+    };
```
malcolmgreaves, line 688: Could you please refactor this into a private helper function: `oxen_based_on: &HttpRequest -> Option<&str>`. Callers then should do something like:
```rust
if let Some(parent) = oxen_based_on(req) {
  check_oxen_based_on(repo, commit, path)?;
}
```

malcolmgreaves (approved):

malcolmgreaves (changes requested):
Hey Paul 👋  Thanks for making this contribution! I've tried to leave guiding comments throughout this review on the details. At a high level though, we both want some of and don't want to take on some of this PR's changes.

We don't want to change the PUT endpoint to accept entire files. It's intentional in the design that this is a multipart upload API. We're biased toward versioning large files. And while performance is always a #1 consideration in our designs and implementations, we are willing to take a small performance hit with some of the multi-part upload when we're versioning small file because it makes our API surface area smaller. One thing we definitely _don't_ want to have happen is having a raw PUT endpoint end up being misused on large files.

However, the optimistic concurrency changes here are excellent! 💯 This is a great idea and we want to have this in `oxen-server`. There's a few tweaks to the implementation that the oxen team would like to see, but overall this idea is brillant.

Would you be willing to transform this PR a bit into adding optimistic concurrency to `oxen-server`? We'd like to see this expanded to the `mv` and `delete` HTTP endpoints too, please ❤️

CleanCut:
Paul indicated in other communication channels that he's not going to continue here.

PR #438, opened 2026-04-12
Add webhook notification system

paul-hammant:
New webhook subsystem for notifying external services on file changes.

  - **Model**: Webhook, WebhookAddRequest, WebhookConfig types
  - **Storage**: JSON-file-backed per-repo at .oxen/webhooks/
  - **Path matching**: Hierarchical prefix matching (webhook on /data fires
    for /data/train/img.png; webhook on / catches everything)
  - **Notifier**: Async HTTP POST delivery with per-webhook rate limiting
    and auto-removal after 5 consecutive failures
  - **Endpoints**: POST /webhooks/add, GET /webhooks, DELETE /webhooks/{id},
    GET/PUT /webhooks/config, GET /webhooks/stats

5 new files, 6 modified files, 8 unit tests for the DB layer.

Connected apps including OxenSync would need this.

### DIFF
--- a/crates/lib/src/core/db/webhooks.rs
+++ b/crates/lib/src/core/db/webhooks.rs
@@ -1,0 +1,286 @@
+use crate::error::OxenError;
+use crate::model::webhook::{Webhook, WebhookAddRequest, WebhookConfig};
+use std::path::{Path, PathBuf};
+
+pub struct WebhookDB {
+    webhooks_path: PathBuf,
+    config_path: PathBuf,
+}
+
+impl WebhookDB {
+    pub fn new(repo_dot_oxen_path: &Path) -> Result<Self, OxenError> {
+        let dir = repo_dot_oxen_path.join("webhooks");
+        std::fs::create_dir_all(&dir)?;
+        Ok(WebhookDB {
+            webhooks_path: dir.join("webhooks.json"),
+            config_path: dir.join("config.json"),
+        })
+    }
+
+    pub fn add_webhook(&self, req: WebhookAddRequest) -> Result<Webhook, OxenError> {
+        let mut webhooks = self.list_all_webhooks()?;
+        let webhook = Webhook {
+            id: uuid::Uuid::new_v4().to_string(),
+            path: req.path,
+            webhook_url: req.webhook_url,
+            webhook_secret: uuid::Uuid::new_v4().to_string(),
+            purpose: req.purpose,
+            contact: req.contact,
+            consecutive_failures: 0,
+        };
+        webhooks.push(webhook.clone());
+        self.save_webhooks(&webhooks)?;
+        Ok(webhook)
+    }
+
+    pub fn list_all_webhooks(&self) -> Result<Vec<Webhook>, OxenError> {
+        if !self.webhooks_path.exists() {
+            return Ok(vec![]);
+        }
+        let content = std::fs::read_to_string(&self.webhooks_path)?;
+        if content.trim().is_empty() {
+            return Ok(vec![]);
+        }
+        let webhooks: Vec<Webhook> = serde_json::from_str(&content)?;
+        Ok(webhooks)
+    }
+
+    pub fn remove_webhook(&self, id: &str) -> Result<(), OxenError> {
+        let mut webhooks = self.list_all_webhooks()?;
+        webhooks.retain(|w| w.id != id);
+        self.save_webhooks(&webhooks)?;
+        Ok(())
+    }
+
+    /// Find all webhooks whose path prefix matches the given changed file path.
+    /// A webhook registered at `/data` will match changes to `/data/file.txt`,
+    /// `/data/subdir/file.txt`, etc. A root webhook (`/` or empty) matches everything.
+    pub fn list_webhooks_for_path(&self, changed_path: &str) -> Result<Vec<Webhook>, OxenError> {
+        let webhooks = self.list_all_webhooks()?;
+        let normalized_changed = normalize_webhook_path(changed_path);
+        Ok(webhooks
+            .into_iter()
+            .filter(|w| {
+                let normalized_webhook = normalize_webhook_path(&w.path);
+                if normalized_webhook.is_empty() {
+                    return true; // Root webhook matches everything
+                }
+                normalized_changed.starts_with(&normalized_webhook)
+                    && (normalized_changed.len() == normalized_webhook.len()
+                        || normalized_changed.as_bytes().get(normalized_webhook.len())
+                            == Some(&b'/'))
+            })
+            .collect())
+    }
+
+    pub fn increment_failure(&self, id: &str) -> Result<u32, OxenError> {
+        let mut webhooks = self.list_all_webhooks()?;
+        let mut count = 0;
+        for w in &mut webhooks {
+            if w.id == id {
+                w.consecutive_failures += 1;
+                count = w.consecutive_failures;
+            }
+        }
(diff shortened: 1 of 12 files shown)


paul-hammant:
Some things we didn't act on from CodeRabbit commentary

1. Atomic file writes (webhooks.rs read→modify→write):                                                                                                                                    
   
  ▎ Acknowledged — the read→modify→write pattern here matches how the rest of the codebase handles similar JSON/config files. Webhook CRUD is low-frequency and serialized by the HTTP   
  request handler. Happy to add file locking if the maintainers think it's warranted, but didn't want to introduce a pattern that doesn't exist elsewhere in the codebase.
                                                                                                                                                                                         
2.  HMAC signature header:                                                                                                                                                                 
   
  ▎ Agreed this would be a good addition. The webhook_secret field is stored for this purpose. Leaving it for a follow-up since it adds a crate dependency (hmac/sha2) and I'd want to   
  align on the header name and signature format with the maintainers first.
                                                                                                                                                                                         
3.  In-memory streaming for large uploads:                                                                                                                                                 
   
  ▎ The multipart PUT path has the same in-memory buffering pattern (see FieldReader for MultipartTempFileNew — it collects all chunks into a Vec<u8>). This is a pre-existing design    
  choice, not something introduced in this PR. A streaming refactor would need to touch FileContents and process_and_add_files across both code paths.

paul-hammant:
Anything left for Claude and I to respond to, @coderabbitai ?

paul-hammant:
Hey @coderabbitai 

## Re SSRF validation
                                                                                                                                                                                         
We've intentionally deferred this to match the current security posture of the codebase. The import endpoint (controllers/import.rs on main) performs outbound HTTP fetches from   user-supplied URLs with identical scheme-only validation — no private IP blocking. Adding SSRF protection to webhooks alone would create an inconsistent security boundary.
                                                                                                                                                                                         
Happy to submit a separate follow-up PR that adds private-range blocking to both the import and webhook endpoints together, so the protection is consistent. Flagging it here as a     known gap rather than a partial fix.
                                                                                                                                                                                         
## Re  Repo-resolution boilerplate                                   

This pattern is the established convention across the entire codebase — branches.rs alone repeats it 7 times, and every other controller does the same. A helper extraction would be   a codebase-wide refactor orthogonal to this PR. Leaving as-is to match existing style.

paul-hammant:
Thanks @coderabbitai

paul-hammant:
This branch ressurected from stale [PUT_should_be_able_to_overwrite_a_HEAD_revision8 Pull request](https://github.com/Oxen-AI/Oxen/pull/416)

crates/lib/src/core/db/webhooks.rs
```diff
@@ -0,0 +1,286 @@
...
+
+pub struct WebhookDB {
+    webhooks_path: PathBuf,
+    config_path: PathBuf,
+}
+
+impl WebhookDB {
+    pub fn new(repo_dot_oxen_path: &Path) -> Result<Self, OxenError> {
+        let dir = repo_dot_oxen_path.join("webhooks");
+        std::fs::create_dir_all(&dir)?;
```
malcolmgreaves, line 13: We're trying to ensure that all I/O operations use `async`. (_Not every place does it, but that's some legacy we're dealing with. For performance, new code is trying to use `async` 100% of the time when it's appropriate._ ) Otherwise, this becomes a thread-blocking operation, which means overall throughput in the system drops.

Most of the `std::fs::*` functions have equivalent `async` counterparts in `tokio::fs::*`. IIRC `tokio::fs::create_dir_all(&dir).await?` is what you'd need here.

crates/lib/src/core/db/webhooks.rs
```diff
@@ -0,0 +1,286 @@
...
+            path: req.path,
+            webhook_url: req.webhook_url,
+            webhook_secret: uuid::Uuid::new_v4().to_string(),
+            purpose: req.purpose,
+            contact: req.contact,
+            consecutive_failures: 0,
+        };
+        webhooks.push(webhook.clone());
+        self.save_webhooks(&webhooks)?;
+        Ok(webhook)
```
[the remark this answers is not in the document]
malcolmgreaves: This ☝️ is correct. Rename is atomic. But making an explicit mutex lock is a good design choice too.

However, storing this webhook information in files and doing FS interaction adds a lot of latency. A more performance-conscious approach would be to have an embedded DB here that would manage this state.

crates/lib/src/core/webhooks.rs
malcolmgreaves, line 1: As a note -- since all of the webhooks code is only used in the server, all of this should live under `crates/oxen-server` instead of `crates/lib`.

crates/lib/src/core/webhooks.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// Notify all webhooks that match the given changed path.
+    /// Returns the number of webhooks that were notified (not rate-limited).
+    /// Respects the repo's webhook config: skips delivery if disabled.
+    pub async fn notify_path_changed(
+        &self,
+        repo_path: &Path,
+        changed_path: &str,
+    ) -> Result<usize, OxenError> {
+        let oxen_dir = repo_path.join(".oxen");
```
malcolmgreaves, line 53: Prefer using the dedicated constant instead of a raw `".oxen"` string wherever possible.

crates/server/src/services/webhooks.rs
malcolmgreaves, line 1: Great modularization of the `webooks()` routes 👏

crates/server/README.md
```diff
@@ -76,9 +76,42 @@ curl -H "Authorization: Bearer $TOKEN" "http://0.0.0.0:3000/api/repos"
 
 #### Create Repository
 ```bash
-curl -H "Authorization: Bearer $TOKEN" -X POST -d '{"name": "MyRepo"}' "http://0.0.0.0:3000api/repos"
+curl -H "Authorization: Bearer $TOKEN" -X POST -d '{"name": "MyRepo"}' "http://0.0.0.0:3000/api/repos"
```
malcolmgreaves, line 79: Thank you 😅

malcolmgreaves (changes requested):
This is a really interesting feature for `oxen-server`! I can see how it is useful to have a system that notifies others on actions that happen in the server.

However, unfortunately we do not want to increase `oxen-server`'s scope to also manage webhooks. There's two major reasons why we don't want this in `oxen-server`:
1. performance & reliability
2. separation of concerns & scope

For (1), this feature would introduce additional work on potentially _every_ HTTP call. The implemented approach here scales linearly with the number of subscribers -- `oxen-server` has to respond to all `N` subscribers per-HTTP endpoint. This has the potential to add a non-zero amount of overhead to the server, slowing down other critical operations. Additionally, the implementation uses files on disk to store webhook state. Access here is rather slow.

For (2), it adds an entirely new area of state for `oxen-server` to manage. This adds state, which would maybe require making a migration script to migrate older repositories. It would at least have to be done in a backwards-compatible way, so there's some additional work to be done for this -- newer clients can't use webhooks on older servers.

More importantly than the compatibility issue is it changes a philosophy in `oxen-server`. It only stores state related to the repositories themselves. There's nothing else in `oxen-server` that is stateful _except_ for the actual data being versioned. This PR changes the philosophy, which we don't want to do at this time.

CleanCut:
Paul indicated in other communication channels that he's not going to continue here.

PR #439, opened 2026-04-13
S3: Implement `store_version_chunk`

CleanCut:
Step 5 of the S3 Project: Implement `S3VersionStore::store_version_chunk`

This one is pretty straightforward.

- Flatten `version_dir` from `{prefix}/{hash[0:2]}/{hash[2:]}` to `{prefix}/{hash}`, because s3 doesn't have limitations on how many "files" it can have in a "directory" (they're all just string keys)
- Change `store_version_chunk` signature to accept `Bytes` instead of `&[u8]` so the s3 implementation can move bytes into `ByteStream` without copying
- Add `chunk_key` helper to format the chunk key
- Rename `init_client` to `client`, since its primary purpose is to return the client.

### DIFF
--- a/crates/lib/src/storage/local.rs
+++ b/crates/lib/src/storage/local.rs
@@ -250,7 +250,7 @@ impl VersionStore for LocalVersionStore {
         &self,
         hash: &str,
         offset: u64,
-        data: &[u8],
+        data: Bytes,
     ) -> Result<(), OxenError> {
         let chunk_dir = self.version_chunk_dir(hash, offset);
         fs::create_dir_all(&chunk_dir).await?;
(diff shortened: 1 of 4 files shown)


crates/lib/src/storage/local.rs
```diff
@@ -250,15 +250,15 @@ impl VersionStore for LocalVersionStore {
         &self,
         hash: &str,
         offset: u64,
-        data: &[u8],
+        data: Bytes,
```
malcolmgreaves, line 253: Could this be &Bytes? I see it's only borrowed below.

malcolmgreaves (approved):
LGTM minor comment about &Bytes.

crates/lib/src/storage/local.rs
CleanCut: This is a trait method, and the S3 implementation needs the owned bytes in order to stream it to S3 without copying. So the local implementation has to have the same signature. Fortunately, we don't need the data after anywhere that we're calling this method, so we're totally fine consuming it.

PR #441, opened 2026-04-13
S3: Implement `delete_version`

CleanCut:
Step 6 of the S3 Project: Implement `S3VersionStore::delete_version`

- List objects under `{prefix}/{hash}/` and batch-delete via `DeleteObjects`
- Add `list_objects_with_prefix()` helper method to list files in s3, handling pagination
- Add `delete_objects()` helper method to delete files in s3.
- For `OxenError`: rename the variant `AwsS3Error` → `AwsError` and replace the `aws_s3_error` method with `From` impls for `SdkError` and `BuildError,` so `?` propagates directly
- Implement `S3VersionStore::delete_version`
- Add tests for `S3VersionStore::delete_version`

### DIFF
--- a/crates/lib/src/storage/s3.rs
+++ b/crates/lib/src/storage/s3.rs
@@ -165,20 +165,16 @@ impl S3VersionStore {
             if let Some(errors) = resp.errors
                 && !errors.is_empty()
             {
-                let msg = errors
+                let key_failures = errors
                     .iter()
                     .map(|e| {
-                        format!(
-                            "{}: {}",
+                       (
                             e.key.as_deref().unwrap_or("?"),
                             e.message.as_deref().unwrap_or("?")
                         )
                     })
-                    .collect::<Vec<_>>()
-                    .join("; ");
-                return Err(OxenError::basic_str(format!(
-                    "delete_objects: some keys failed to delete: {msg}"
-                )));
+                    .collect::<Vec<_>>();
+                return Err(OxenError::DeleteFailure{key_failures});
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/lib/src/storage/s3.rs
```diff
@@ -110,6 +110,80 @@ impl S3VersionStore {
...
+                            "{}: {}",
+                            e.key.as_deref().unwrap_or("?"),
+                            e.message.as_deref().unwrap_or("?")
+                        )
+                    })
+                    .collect::<Vec<_>>()
+                    .join("; ");
+                return Err(OxenError::basic_str(format!(
+                    "delete_objects: some keys failed to delete: {msg}"
+                )));
```
malcolmgreaves: Please add this as a new structured error to `OxenError`.

Just to illustrate the idea, one approach is to keep the key + error strings as a tuple -- 

```suggestion
                let key_failures = errors
                    .iter()
                    .map(|e| {
                       (
                            e.key.as_deref().unwrap_or("?"),
                            e.message.as_deref().unwrap_or("?")
                        )
                    })
                    .collect::<Vec<_>>();
                return Err(OxenError::DeleteFailure{key_failures});
```


With a new variant:
```
#[error("delete_objects: some keys failed to delete: {key_failures}")
S3DeleteFailure{key_failures: Vec<(String, String)>)
```

malcolmgreaves (approved):
LGTM with one requested change. Like the overall implementation and the new test! 👏

crates/lib/src/storage/s3.rs

PR #442, opened 2026-04-13
S3: Implement `get_version_chunk`

CleanCut:
Implement `S3VersionStore::get_version_chunk` (step 7). This one was straightforward.

- Issue a range request on the combined `{prefix}/{hash}/data` key instead of fetching the whole object.

Resolves ENG-861

### DIFF
--- a/crates/lib/src/storage/s3.rs
+++ b/crates/lib/src/storage/s3.rs
@@ -620,14 +620,41 @@ impl VersionStore for S3VersionStore {
 
     async fn get_version_chunk(
         &self,
-        _hash: &str,
-        _offset: u64,
-        _size: u64,
+        hash: &str,
+        offset: u64,
+        size: u64,
     ) -> Result<Vec<u8>, OxenError> {
-        // TODO: Implement S3 version chunk retrieval
-        Err(OxenError::basic_str(
-            "S3VersionStore get_version_chunk not yet implemented",
-        ))
+        if size == 0 {
+            return Ok(Vec::new());
+        }
+
+        let client = self.client().await?;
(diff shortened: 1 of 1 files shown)


Eric-Laurence (commented):
This looks right, but I do not know how to test it. My review here may not be enough to merge it.

crates/lib/src/storage/s3.rs
```diff
@@ -620,14 +620,41 @@ impl VersionStore for S3VersionStore {
...
-        _size: u64,
+        hash: &str,
+        offset: u64,
+        size: u64,
     ) -> Result<Vec<u8>, OxenError> {
-        // TODO: Implement S3 version chunk retrieval
-        Err(OxenError::basic_str(
-            "S3VersionStore get_version_chunk not yet implemented",
-        ))
+        if size == 0 {
```
subygan, line 627: It'd be neat if we can handle this in the client directly instead of here.

subygan (approved):

crates/lib/src/storage/s3.rs
```diff
@@ -1073,4 +1100,48 @@ mod tests {
         assert!(!store.version_exists(hash_a).await.unwrap());
         assert!(store.version_exists(hash_b).await.unwrap());
     }
+
+    #[tokio::test]
+    async fn test_get_version_chunk_mid_file() {
+        let (store, _tmp, _server) = setup().await;
```
subygan: love these tests.

crates/lib/src/storage/s3.rs
CleanCut: That would be nice

PR #444, opened 2026-04-13
Cleanup, recover, and flush write-ahead-log for duckdb

gschoeni:
We have been getting some errors with the duckdb write-ahead-log file. This PR makes it so we cleanup that file if it is corrupted and does a better job flushing it. It may be corrupted because the server was killed or OOM or a variety of reasons.

This is the error I was seeing before these changes:

Screenshot 2026-04-13 at 2 07 15 PM


EloyMartinez:
👀

PR #445, opened 2026-04-14
Josh/fast workspace lookup

jcelliott:
Implement constant-time lookup for workspaces, instead of iterating over all the workspaces in the repository.

Creates a workspace index per-repo (RocksDB) that maps names to IDs.

Avg time on my machine with Greg's 9604 workspace repo goes from 1.35s to 10.5ms (~128x speedup) on the get workspace endpoint:
`curl localhost:3000/api/repos/<namespace>/<repo_name>/workspaces/<workspace-name>`

The first request to get a workspace will trigger building the index if it doesn't already exist. You can also migrate manually with `oxen migrate up add_workspace_name_index <path/to/repo>` (on your server-side repo).

### DIFF
--- a/crates/lib/src/command/migrate/m20260408_add_workspace_name_index.rs
+++ b/crates/lib/src/command/migrate/m20260408_add_workspace_name_index.rs
@@ -1,0 +1,122 @@
+use std::path::Path;
+
+use super::Migrate;
+
+use crate::core::workspaces::workspace_name_index;
+use crate::error::OxenError;
+use crate::model::LocalRepository;
+use crate::model::Workspace;
+use crate::repositories;
+use crate::util;
+use crate::util::progress_bar::{ProgressBarType, oxen_progress_bar};
+
+pub struct AddWorkspaceNameIndexMigration;
+
+impl Migrate for AddWorkspaceNameIndexMigration {
+    fn name(&self) -> &'static str {
+        "add_workspace_name_index"
+    }
+
+    fn description(&self) -> &'static str {
+        "Creates a RocksDB index mapping workspace names to IDs for O(1) lookup"
+    }
+
+    fn up(&self, path: &Path, all: bool) -> Result<(), OxenError> {
+        if all {
+            run_on_all_repos(path)?;
+        } else {
+            let repo = LocalRepository::from_dir(path)?;
+            run_on_one_repo(&repo)?;
+        }
+        Ok(())
+    }
+
+    fn down(&self, path: &Path, all: bool) -> Result<(), OxenError> {
+        if all {
+            revert_all_repos(path)?;
+        } else {
+            let repo = LocalRepository::from_dir(path)?;
+            revert_one_repo(&repo)?;
+        }
+        Ok(())
+    }
+
+    fn is_needed(&self, repo: &LocalRepository) -> Result<bool, OxenError> {
+        // Needed if workspace directory exists but index does not
+        let workspaces_dir = Workspace::workspaces_dir(repo);
+        Ok(workspaces_dir.exists() && !workspace_name_index::index_exists(repo))
+    }
+}
+
+fn run_on_all_repos(path: &Path) -> Result<(), OxenError> {
+    println!("🐂 Collecting namespaces to migrate...");
+    let namespaces = repositories::list_namespaces(path)?;
+    let bar = oxen_progress_bar(namespaces.len() as u64, ProgressBarType::Counter);
+    println!("🐂 Migrating {} namespaces", namespaces.len());
+    for namespace in namespaces {
+        let namespace_path = path.join(namespace);
+        let repos = repositories::list_repos_in_namespace(&namespace_path);
+        for repo in repos {
+            match run_on_one_repo(&repo) {
+                Ok(_) => {}
+                Err(err) => {
+                    log::error!(
+                        "Could not create workspace name index for repo {:?}\nErr: {}",
+                        repo.path.canonicalize(),
+                        err
+                    );
+                }
+            }
+        }
+        bar.inc(1);
+    }
+    Ok(())
+}
+
+fn run_on_one_repo(repo: &LocalRepository) -> Result<(), OxenError> {
+    let workspaces_dir = Workspace::workspaces_dir(repo);
+    if !workspaces_dir.exists() {
+        return Ok(());
+    }
+
+    log::info!("Creating workspace name index for repo: {:?}", repo.path);
+
+    workspace_name_index::with_index(repo, |idx| idx.rebuild_from_disk(repo))
+}
+
+fn revert_all_repos(path: &Path) -> Result<(), OxenError> {
+    println!("🐂 Collecting namespaces to revert...");
+    let namespaces = repositories::list_namespaces(path)?;
+    let bar = oxen_progress_bar(namespaces.len() as u64, ProgressBarType::Counter);
+    println!("🐂 Reverting {} namespaces", namespaces.len());
+    for namespace in namespaces {
+        let namespace_path = path.join(namespace);
+        let repos = repositories::list_repos_in_namespace(&namespace_path);
+        for repo in repos {
+            match revert_one_repo(&repo) {
+                Ok(_) => {}
+                Err(err) => {
+                    log::error!(
+                        "Could not revert workspace name index for repo {:?}\nErr: {}",
+                        repo.path.canonicalize(),
+                        err
+                    );
+                }
+            }
+        }
+        bar.inc(1);
+    }
+    Ok(())
+}
+
+fn revert_one_repo(repo: &LocalRepository) -> Result<(), OxenError> {
+    workspace_name_index::remove_from_cache(&repo.path)?;
+    let index_dir = repo
+        .path
+        .join(crate::constants::OXEN_HIDDEN_DIR)
+        .join(crate::constants::WORKSPACE_NAME_INDEX_DIR);
+    if index_dir.exists() {
+        util::fs::remove_dir_all(&index_dir)?;
+    }
+    Ok(())
+}
--- a/crates/lib/src/command/migrate/m20260408_add_workspace_name_index.rs
+++ b/crates/lib/src/command/migrate/m20260408_add_workspace_name_index.rs
@@ -62,7 +62,7 @@ fn run_on_all_repos(path: &Path) -> Result<(), OxenError> {
                 Err(err) => {
                     log::error!(
                         "Could not create workspace name index for repo {:?}\nErr: {}",
-                        repo.path.canonicalize(),
+                        util::fs::canonicalize(repo.path),
                         err
                     );
                 }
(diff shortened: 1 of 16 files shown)


crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,327 @@
...
+fn index_dir(repo: &LocalRepository) -> PathBuf {
+    repo.path
+        .join(OXEN_HIDDEN_DIR)
+        .join(WORKSPACE_NAME_INDEX_DIR)
+}
+
+/// Check if the workspace name index DB exists for this repo.
+pub fn index_exists(repo: &LocalRepository) -> bool {
+    index_dir(repo).exists()
+}
```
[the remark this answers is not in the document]
jcelliott: Good find, but probably not worth addressing. Even for my test repo with an unusually high number of workspaces, generating the index took ~1.5 seconds.

crates/lib/src/command/migrate/m20260408_add_workspace_name_index.rs
```diff
@@ -0,0 +1,124 @@
...
+    for namespace in namespaces {
+        let namespace_path = path.join(namespace);
+        let repos = repositories::list_repos_in_namespace(&namespace_path);
+        for repo in repos {
+            match run_on_one_repo(&repo) {
+                Ok(_) => {}
+                Err(err) => {
+                    log::error!(
+                        "Could not create workspace name index for repo {:?}\nErr: {}",
+                        repo.path.canonicalize(),
```
CleanCut: I need to add a lint for this. The `canonicalize` from the standard library doesn't support ramdisks on Windows, so my (open) PR to use ramdisks improves `util::fs::canonicalize` and switches everything to use it.


```suggestion
                        util::fs::canonicalize(repo.path),
```

crates/lib/src/command/migrate/m20260408_add_workspace_name_index.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+fn revert_one_repo(repo: &LocalRepository) -> Result<(), OxenError> {
+    workspace_name_index::remove_from_cache(&repo.path)?;
+    let index_dir = repo
+        .path
+        .join(crate::constants::OXEN_HIDDEN_DIR)
+        .join(crate::constants::WORKSPACES_DIR)
+        .join(crate::constants::WORKSPACE_NAME_INDEX_DIR);
+    if index_dir.exists() {
```
CleanCut: There's a function in `workspace_name_index.rs` that already does most of this.

```suggestion
    if index_exists(repo) {
```

crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Check if a name exists in the index. O(1).
+    pub fn has_name(&self, name: &str) -> bool {
+        match self.db.get_pinned(name.as_bytes()) {
+            Ok(Some(_)) => true,
+            Ok(None) => false,
+            Err(err) => {
+                log::error!("Error checking workspace name index for '{name}': {err}");
+                false
+            }
```
CleanCut: Since every place that receives a `false` converts it to an error, we should return an error when there's an error rather than just converting it to a `false` and have it immediately converted back to an error without the actual underlying error information.

crates/lib/src/repositories/workspaces.rs
```diff
@@ -312,6 +418,14 @@ pub fn delete(workspace: &Workspace) -> Result<(), OxenError> {
 
     log::debug!("workspace::delete cleaning up workspace dir: {workspace_dir:?}");
 
+    // Remove from name index before deleting the workspace directory
+    if let Some(ref name) = workspace.name
+        && workspace_name_index::index_exists(&workspace.base_repo)
+    {
+        let idx = workspace_name_index::get_index(&workspace.base_repo)?;
+        idx.delete(name)?;
```
CleanCut: Using `?` here prevents the rest of the cleanup if something goes wrong. We need to log and continue.

```suggestion
        match workspace_name_index::get_index(&workspace.base_repo) {
            Ok(idx) => {
                if let Err(e) = idx.delete(name) {
                    log::error!("workspace::delete error removing workspace index: {e:?}");
                }
            }
            Err(e) => log::error!("workspace::delete error finding workspace index: {e:?}"),
        }
```

crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,330 @@
...
+                        "Error iterating workspace name index: {err}"
+                    )));
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// List all (name, workspace_id) entries. Primarily for debugging/testing.
+    pub fn list(&self) -> Result<Vec<(String, String)>, OxenError> {
```
CleanCut: It's only for tests.

```suggestion
    /// List all (name, workspace_id) entries for debugging/testing.
    #[cfg(test)]
    pub fn list(&self) -> Result<Vec<(String, String)>, OxenError> {
```

crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,330 @@
...
+        .join(WORKSPACE_NAME_INDEX_DIR)
+}
+
+/// Check if the workspace name index DB exists for this repo.
+pub fn index_exists(repo: &LocalRepository) -> bool {
+    index_dir(repo).exists()
+}
+
+/// Removes this repository's workspace name index DB from the cache.
+pub fn remove_from_cache(repository_path: impl AsRef<Path>) -> Result<(), OxenError> {
```
CleanCut: This needs to be called from `maybe_cleanup_repo` in `crates/lib/src/test.rs`.

⚠️ This may be one of the causes for the Windows CI failures.

crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,330 @@
...
+    let dir = util::fs::oxen_hidden_dir(&repository_path)
+        .join(WORKSPACES_DIR)
+        .join(WORKSPACE_NAME_INDEX_DIR);
+    let mut instances = DB_INSTANCES.lock();
+    let _ = instances.pop(&dir); // drop immediately
+    Ok(())
+}
+
+/// Removes from cache including children paths (useful for test cleanup).
+pub fn remove_from_cache_with_children(repository_path: impl AsRef<Path>) -> Result<(), OxenError> {
```
CleanCut: Only needed for tests.

```suggestion
#[cfg(test)]
pub fn remove_from_cache_with_children(repository_path: impl AsRef<Path>) -> Result<(), OxenError> {
```

⚠️ ALSO, this needs to be called in `cleanup_sync_dir` in `crates/server/src/test.rs`. This may be another cause for the Windows CI failures.

crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,330 @@
...
+                }
+            }
+        }
+        Ok(results)
+    }
+
+    /// Rebuild the index from existing workspace configs on disk.
+    /// Clears all existing entries first, then scans `.oxen/workspaces/` directories.
+    pub fn rebuild_from_disk(&self, repo: &LocalRepository) -> Result<(), OxenError> {
+        self.clear()?;
```
CleanCut, line 200: We're already rebuilding the entire database. It may be both simpler and faster to drop the entire database and create a new one, rather than using write operations to clear it first.

crates/lib/src/repositories/workspaces.rs
```diff
@@ -191,6 +235,67 @@ pub fn create_with_name(
...
+fn validate_create_constraints(
+    base_repo: &LocalRepository,
+    commit: &Commit,
+    workspace_name: &Option<String>,
+    is_editable: bool,
+) -> Result<(), OxenError> {
+    let has_index = workspace_name_index::index_exists(base_repo);
+
+    // Fast path: use the index for name checks when we don't need to iterate for non-editable
+    if has_index && is_editable {
```
CleanCut, line 249: It looks like we have the concept of a unique, non-editable workspace that's just used to provide a read-only place for DuckDB to run queries against.  The lookup for that non-editable workspace takes the slow path.

I think this is okay, because I think we should split out duckdb/polars/all-the-dataframe-stuff into a separate project entirely, at which point it would be accessing the bytes in some completely different way.

CleanCut (commented):
Q: Did we want to move towards using LMDB (instead of RocksDB) when we can?

crates/lib/src/core/workspaces/workspace_name_index.rs
```diff
@@ -0,0 +1,330 @@
...
+        util::fs::create_dir_all(&dir).map_err(|e| {
+            OxenError::basic_str(format!(
+                "Failed to create workspace name index directory: {e}"
+            ))
+        })?;
+    }
+
+    let opts = db::key_val::opts::default();
+    let db = DB::open(&opts, dunce::simplified(&dir)).map_err(|e| {
+        OxenError::basic_str(format!("Failed to open workspace name index database: {e}"))
```
malcolmgreaves: Please avoid introducing new code that uses `OxenError::Basic`. My review feedback for this is here: https://github.com/Oxen-AI/Oxen/pull/454

jcelliott:
> Q: Did we want to move towards using LMDB (instead of RocksDB) when we can?

I think eventually that would be good, assuming the LMDB work goes well. RocksDB has generally worked well for these small KV lookups, but LMDB should work just as well for that and it would be nice to remove a dependency.

crates/lib/src/repositories/workspaces.rs
```diff
@@ -191,6 +235,67 @@ pub fn create_with_name(
...
+        if let Some(name) = workspace_name {
+            // Check name doesn't collide with an existing workspace name
+            let idx = workspace_name_index::get_index(base_repo)?;
+            if idx.has_name(name) {
+                return Err(OxenError::basic_str(format!(
```
malcolmgreaves: `Err(OxenError::WorkspaceAlreadyExists(name.to_string()))`

malcolmgreaves (commented):
LGTM on the algorithm an implementation. Requesting changes on the new errors kinds that are introduced here. This speedup is excellent! 👏

crates/lib/src/core/workspaces/workspace_name_index.rs
jcelliott: Added to the server tests. I didn't add the `#[cfg(test)]` because it would only be available in the *lib's* tests, not in the server crate tests.

crates/lib/src/core/workspaces/workspace_name_index.rs
jcelliott: Faster, definitely. Probably not simpler, since we already have an open handle to the DB in this context we would have to manage that plus the cache.

This should only ever be called once per repo, so I'm going to skip changing this unless you feel like it's important.

CleanCut: Nah, not that important. Especially if we only expect to call it once.

CleanCut (approved):

PR #447, opened 2026-04-14
S3: Implement `list_version_chunks`

CleanCut:
S3 plan step 9: Implement `list_version_chunks`

Add `chunks_prefix()` helper and implement `list_version_chunks` using `list_objects_with_prefix` to enumerate chunk keys and parse offsets. Includes s3s tests for happy path, empty, and hash isolation.

### DIFF
--- a/crates/lib/src/storage/s3.rs
+++ b/crates/lib/src/storage/s3.rs
@@ -111,6 +111,11 @@ impl S3VersionStore {
         format!("{}/chunks/{}", self.version_dir(hash), offset)
     }
 
+    /// Get the S3 key prefix for all chunks of a version
+    fn chunks_prefix(&self, hash: &str) -> String {
+        format!("{}/chunks/", self.version_dir(hash))
+    }
+
     /// List all S3 object keys under a given prefix, following continuation tokens.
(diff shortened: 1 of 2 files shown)


crates/lib/src/storage/s3.rs
```diff
@@ -620,48 +625,30 @@ impl VersionStore for S3VersionStore {
...
 
-        let bytes = resp
-            .body
-            .collect()
-            .await
-            .map_err(|e| {
-                OxenError::basic_str(format!("get_version_chunk: body collect failed: {e}"))
-            })?
-            .into_bytes();
+    async fn list_version_chunks(&self, hash: &str) -> Result<Vec<u64>, OxenError> {
```
malcolmgreaves, line 638: Could you please add a short doc comment on this function explaining how the `u64`s are to be used. If it's already implemented, could you also have this comment include a pointer about where it should be used?

malcolmgreaves (approved):
LGTM w/ one documentation request.

crates/lib/src/storage/s3.rs
CleanCut: ✅ Done. It's documented on the `VersionStore` trait definition of the method. I updated that documentation to explain the return value and how to use it.

PR #449, opened 2026-04-14
S3: Implement `combine_version_chunks` (step 10)

CleanCut:
This is step 10 of the S3 project.

- Implement `S3VersionStore::combine_version_chunks`
- Simplify the trait signature: remove the unused `cleanup` parameter (always clean up) 
- Change the return type from `PathBuf` to `()` since the return value was unused.
- Guarantee sorted order from `list_version_chunks` so callers don't need to sort.
- Add tests

SIDENOTE: Much (or all?) of the optimizations from the parallel upload between client and server are lost since we've got to store the chunks in S3, then retrieve the chunks from S3, then combine and (re)store them back into S3. That's a lot of manual complexity for no benefit, here. One of my later steps is to look into removing the manual chunking/combining entirely in favor of the single, streaming upload from client to server to S3.

### DIFF
--- a/crates/lib/src/storage/s3.rs
+++ b/crates/lib/src/storage/s3.rs
@@ -748,7 +748,7 @@ impl VersionStore for S3VersionStore {
                     .collect()
                     .await
                     .map_err(|e| {
-                        OxenError::basic_str(format!(
+                        OxenError::internal_error(format!(
                             "Failed to read chunk body at offset {offset}: {e}"
                         ))
                     })?
--- a/crates/lib/src/storage/s3.rs
+++ b/crates/lib/src/storage/s3.rs
@@ -708,7 +708,6 @@ impl VersionStore for S3VersionStore {
     }
 
     async fn combine_version_chunks(&self, hash: &str) -> Result<(), OxenError> {
-        // 1. List chunk offsets (already sorted by list_version_chunks)
         let offsets = self.list_version_chunks(hash).await?;
         if offsets.is_empty() {
             return Ok(());
@@ -718,7 +717,6 @@ impl VersionStore for S3VersionStore {
         let client = self.client().await?;
         let key = self.generate_key(hash);
 
-        // 2. Create multipart upload
         let upload = client
             .create_multipart_upload()
             .bucket(&self.bucket)
@@ -730,7 +728,6 @@ impl VersionStore for S3VersionStore {
             .ok_or_else(|| OxenError::upload("S3 multipart upload missing upload_id"))?
             .to_string();
 
-        // 3. Stream chunks and upload as multipart parts
         const MIN_PART_SIZE: usize = 5 * 1024 * 1024; // 5 MB, S3 minimum
         let mut part_buf: Vec<u8> = Vec::new();
         let mut part_num: i32 = 1;
@@ -740,7 +737,6 @@ impl VersionStore for S3VersionStore {
             for (i, offset) in offsets.iter().enumerate() {
                 let is_last_chunk = i == offsets.len() - 1;
 
-                // Download chunk from S3
                 let resp = client
                     .get_object()
                     .bucket(&self.bucket)
@@ -758,7 +754,6 @@ impl VersionStore for S3VersionStore {
                     })?
                     .into_bytes();
 
-                // Append to part buffer
                 part_buf.extend_from_slice(&chunk_bytes);
 
                 // Upload part when buffer is large enough (or on last chunk)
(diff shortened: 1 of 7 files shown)


crates/lib/src/storage/s3.rs
```diff
@@ -706,15 +707,128 @@ impl VersionStore for S3VersionStore {
...
-    async fn combine_version_chunks(
-        &self,
-        _hash: &str,
-        _cleanup: bool,
-    ) -> Result<PathBuf, OxenError> {
-        // TODO: Implement S3 version chunk combination
-        Err(OxenError::basic_str(
-            "S3VersionStore combine_version_chunks not yet implemented",
-        ))
+    async fn combine_version_chunks(&self, hash: &str) -> Result<(), OxenError> {
```
malcolmgreaves, line 710: The comments between line 764 and 788 are excellent! They're great at documenting the logic around buffer handling. 

However, the other comments here don't feel like they're on the same level. Line 711, 721, 733, 743, 761, 798, and 810 aren't as helpful because what they're explaining is readily evident from the code. For review, please expand on them to make them more relevant or remove them.

crates/lib/src/storage/version_store.rs
malcolmgreaves, line 1: Great doc comments here 👏

crates/lib/src/storage/s3.rs
```diff
@@ -706,15 +707,128 @@ impl VersionStore for S3VersionStore {
...
+                    .send()
+                    .await?;
+                let chunk_bytes = resp
+                    .body
+                    .collect()
+                    .await
+                    .map_err(|e| {
+                        OxenError::basic_str(format!(
+                            "Failed to read chunk body at offset {offset}: {e}"
+                        ))
```
malcolmgreaves: Please avoid introducing new code that uses `OxenError::Basic`.

malcolmgreaves (commented):
Overall great PR! It makes sense to have the offsets always be sorted and to always perform cleanup. And changing the trait so that `combine_version_chunks` doesn't return a local path also makes sense given how this is being used.

I do think it's worth taking a pause to look at the design of the version store trait and see what it would take to change it. As you pointed out, it's not great that we're manually handling multipart file uploads & downloads for S3 when S3 can handle that automatically. It seems a bit odd that our interface is at this level of detail 🤔 However, it's fair to keep that to another PR.

Some minor comments to be addressed. Everything else LGTM.

crates/lib/src/storage/s3.rs
CleanCut: ✅ Done

crates/lib/src/storage/s3.rs

crates/lib/src/storage/s3.rs
```diff
@@ -706,15 +707,121 @@ impl VersionStore for S3VersionStore {
...
+                    .bucket(&self.bucket)
+                    .key(self.chunk_key(hash, *offset))
+                    .send()
+                    .await?;
+                let chunk_bytes = resp
+                    .body
+                    .collect()
+                    .await
+                    .map_err(|e| {
+                        OxenError::internal_error(format!(
```
malcolmgreaves, line 751: The new `InternalError` has the same problem as `OxenError::Basic`. It does not convey useful information about the source of the error.

This is a failed to read chunk body offset error.

This file has a lot of specific error conditions that are throwing away useful structured information by putting it into a string. I suggest using this pattern to re-establish the structured information:

Define in this module:
```rust
pub enum S3StoreError {
   #[error("Failed to read chunk body at offset {0}: {1}")]
   ChunkRead(u64, <request error>),

   #[error("S3 multipart upload missing upload_id")
   MissingUploadId,

   <etc.>
}
```

Then, in `OxenError`, define a wrapper for all of the specific errors that crop up in the S3 version store:
```
pub enum OxenError {
   ...

   #[error("{0}")]
   S3StoreError(#[from] crate::storage::s3::S3StoreError)
}
```

This approach:

1. keeps error information structured
2. keeps error definitions close to the code that uses them
3. is a scalable pattern: this "S3StoreError" can change and the conversion to `OxenError` remains stable

For maximum benefit, the `VersionStore` trait should define its own associated type for errors and have its methods return `Result<..., Self::Error>`.

malcolmgreaves: @CleanCut this idea ☝️ can come in a later refactoring PR. I'll approve this one.

malcolmgreaves (approved):

PR #450, opened 2026-04-15
Collapse `Entry` newtype into `CommitEntry`

CleanCut:
After the recent [dead code removal](#449), the `Entry` type does nothing other than wrap the `CommitEntry` type. This PR removes the `Entry` type to use `CommitEntry` directly.

- Deleted the entire `Entry` type  from `commit_entry.rs`
- Pruned a bunch of dead code from `CommitEntry`
- Changed on `CommitEntry`:
  -  `Hash`/`PartialEq` now use the content hash field. That's what the `Entry` wrapper was doing which was used by the fetch command, and nothing used `CommitEntry`'s path-based hashing.

### DIFF
--- a/crates/lib/src/api/client/versions.rs
+++ b/crates/lib/src/api/client/versions.rs
@@ -3,8 +3,7 @@ use crate::api::client;
 use crate::api::client::internal_types::LocalOrBase;
 use crate::constants::{AVG_CHUNK_SIZE, max_retries};
 use crate::error::OxenError;
-use crate::model::entry::commit_entry::Entry;
-use crate::model::{LocalRepository, MerkleHash, RemoteRepository};
+use crate::model::{CommitEntry, LocalRepository, MerkleHash, RemoteRepository};
 use crate::util::{self, concurrency, hasher};
 use crate::view::versions::{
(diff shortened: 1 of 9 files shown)


crates/lib/src/api/client/versions.rs
```diff
@@ -529,32 +528,32 @@ pub async fn multipart_batch_upload(
...
         // if it's not the first try and the file is not in the retry list, skip
-        if !files_to_retry.is_empty() && !retry_hashes.contains(&file_hash) {
+        if !files_to_retry.is_empty() && !retry_hashes.contains(file_hash) {
             continue;
         }
 
-        let data = version_store.get_version(&file_hash).await?;
+        let data = version_store.get_version(file_hash).await?;
         let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
         std::io::copy(&mut data.as_slice(), &mut encoder)?;
```
malcolmgreaves, line 541: Nice opportunity to switch to `tokio::fs::copy` and use `async-compression`'s `GzipEncoder!

malcolmgreaves (approved):
LGTM. Thanks for updating the doc comment on `CommitEntry` too!

PR #455, opened 2026-04-16
add optional metadata to the workspace object

gschoeni:
Right now the hub stores workspace metadata in the database, which means it is easy for the hub and server to get out of sync. This adds a construct for arbitrary user provided json to be stored in the workspace object.

It also adds pagination to the workspaces call.

### DIFF
--- a/crates/lib/src/api/client/workspaces.rs
+++ b/crates/lib/src/api/client/workspaces.rs
@@ -11,23 +11,19 @@ use crate::api;
 use crate::api::client;
 use crate::error::OxenError;
 use crate::model::RemoteRepository;
-use crate::view::workspaces::{ListWorkspaceResponseView, WorkspaceResponseWithStatus};
+use crate::opts::PaginateOpts;
+use crate::view::workspaces::{
+    ListWorkspaceResponseView, PaginatedWorkspaces, UpdateWorkspaceMetadataRequest,
+    WorkspaceResponseWithStatus,
+};
 use crate::view::workspaces::{NewWorkspace, WorkspaceResponse};
 use crate::view::{StatusMessage, WorkspaceResponseView};
 
-pub async fn list(remote_repo: &RemoteRepository) -> Result<Vec<WorkspaceResponse>, OxenError> {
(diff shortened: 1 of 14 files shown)


crates/lib/src/api/client/workspaces.rs
```diff
@@ -53,25 +49,44 @@ pub async fn get_by_name(
     name: impl AsRef<str>,
 ) -> Result<Option<WorkspaceResponse>, OxenError> {
     let name = name.as_ref();
-    let url = api::endpoint::url_from_repo(remote_repo, &format!("/workspaces?name={name}"))?;
+    let response = list_with_params(remote_repo, None, Some(name)).await?;
+    match response.entries.len() {
```
EloyMartinez, line 53: are we sure names are unique?

gschoeni: We want them to be..but yeah let me double check

EloyMartinez:
👀

CleanCut:
> It also adds pagination to the workspaces call.

Does this have anything to do with the metadata? 

I'm super confused. I don't understand what the metadata is for. I don't understand the design for the workspaces, or how they're used on the server vs the client. From what I can tell on the server a workspace is just supposed to be a staging area for composing a remote commit...so why would there ever be a need to have multiple pages of them? Why do we need to store metadata on them?

Help me out, here. 🏳️

gschoeni:
> > It also adds pagination to the workspaces call.
> 
> Does this have anything to do with the metadata?
> 
> I'm super confused. I don't understand what the metadata is for. I don't understand the design for the workspaces, or how they're used on the server vs the client. From what I can tell on the server a workspace is just supposed to be a staging area for composing a remote commit...so why would there ever be a need to have multiple pages of them? Why do we need to store metadata on them?
> 
> Help me out, here. 🏳️

Yes let's do a workspace deep dive! It is a different concept than lives in most VCS so want to get your brain on it too.

jcelliott:
Other branch was merged. Rebased this against main.

PR #461, opened 2026-04-20
Cleanup: Remove `WorkspaceCommit`

CleanCut:
This PR removes `WorkspaceCommit` and just uses `Commit` directly.

A long time ago, the hub didn't include the `parent_ids` field in it's schema for a commit.  The `WorkspaceCommit` struct appears to exist just to drop the `parent_ids` field from `Commit` to send the info up to the hub. Since the hub gained the `parent_ids` field in a migration in Jan 2025, this hacky conversion is no longer necessary--we can just send the `Commit`.


Eric-Laurence (approved):
Looks good, I'm a big fan of all this clean up. _Off with their heads! Desloppify!_

PR #465, opened 2026-04-21
Delete unused `open_read_write_if_not_exists`

malcolmgreaves:
Function `MerkleDbNode::open_read_write_if_not_exists` is unused.
Functionality can be recovered using a `MerkleDbNode::exists` check
before and a `MerkleDbNode::open_read_write` call.


CleanCut (approved):
❤️

PR #467, opened 2026-04-21
Changelog instructions

CleanCut:
Let's start updating the changelog -- claude can help!


Eric-Laurence (approved):
Makes sense to me. Perhaps we should do something similar for OxenHub.

CleanCut:
True! I think it's a good idea for anything with user-facing releases.

PR #468, opened 2026-04-21
Add `--script` argument to `generate-dataset`

CleanCut:
The `bin/generate-dataset` script wouldn't run on my local setup without adding the `--script` flag to the first line.

The other changes were done automatically by whatever python-formatter Zed uses by default. 🤷🏻‍♂️

### DIFF
--- a/bin/generate-dataset
+++ b/bin/generate-dataset
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S uv run
+#!/usr/bin/env -S uv run --script
 # /// script
 # requires-python = ">=3.11"
 # dependencies = [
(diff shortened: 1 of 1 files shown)


malcolmgreaves (commented):
Odd that this wasn't working for you! I can run these w/o incident. It's good to add `--script` to the shebang here though! 

Formatting should be handled by the pre-commit hooks.

I am not sure how these Python scripts lost their `.py` suffix, but that explains why they are not being picked up by the pre-commit hooks anymore.

I made https://github.com/Oxen-AI/Oxen/pull/469 to fix this. And I also ended up adding `--script` to all of the Python scripts too.

CleanCut:
Closing in favor of #469

PR #470, opened 2026-04-21
MerkleDbNode error definition + `HexHash`, move `node_db_prefix`

malcolmgreaves:
Refactors the unique errors for `MerkleDbNode`'s operations into a new error enum.
The new `MerkleDbError` variant replacements maintain the same error messages.
`OxenError` has a new `#[from]` wrapper for this error type.

Introduces a new `HexHash` type, which is internal to the `liboxen` crate. This is a zero-
sized wrapper around the `String` produced by encoding a `MerkleHash`'s 128-bit
unsigned integer into a hexadecimal formatted string. Construction is only possible via
this hex-formatting route: it's impossible to mix-up a regular `String` and a `HexHash`.

The `node_db_prefix` function has been moved to become a method of `HexHash` now.
This also avoids extra `String` allocations as it creates the 2-layered nested directory
of `{hash prefix}/{hash suffix}`, which is used for both Merkle node & versioned file storage.

Also, fixes a bare `.unwrap()` in `r_create_dir_node`. The function now propagates the
`Err` to the caller instead of causing a `panic!`.

### DIFF
--- a/crates/lib/src/core/db/merkle_node/merkle_node_db.rs
+++ b/crates/lib/src/core/db/merkle_node/merkle_node_db.rs
@@ -58,6 +58,7 @@ use crate::error::OxenError;
 use crate::model::LocalRepository;
 use crate::model::MerkleHash;
 use crate::model::merkle_tree::merkle_hash::HexHash;
+use crate::model::merkle_tree::node_type::InvalidMerkleTreeNodeType;
 use crate::util;
 
 use crate::model::merkle_tree::node::{
@@ -80,6 +81,7 @@ pub fn node_db_prefix(hash: &MerkleHash) -> PathBuf {
     Path::new(dir_prefix).join(dir_suffix)
 }
 
+/// An absolute path to the directory for the Merkle node's `node` and `children` files.
 pub fn node_db_path(repo: &LocalRepository, hash: &MerkleHash) -> PathBuf {
     let dir_prefix = node_db_prefix(hash);
     repo.path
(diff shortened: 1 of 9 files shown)


crates/lib/src/core/db/merkle_node/merkle_node_db.rs
```diff
@@ -67,14 +69,19 @@ use crate::model::merkle_tree::node::{
...
+/// The first directory name is the first 3 characters of the hex-encoded hash. The second
+/// is the remaining characters.
 pub fn node_db_prefix(hash: &MerkleHash) -> PathBuf {
-    let hash_str = hash.to_string();
-    let dir_prefix_len = 3;
-    let dir_prefix = hash_str.chars().take(dir_prefix_len).collect::<String>();
-    let dir_suffix = hash_str.chars().skip(dir_prefix_len).collect::<String>();
-    Path::new(&dir_prefix).join(&dir_suffix)
+    let as_hex = HexHash::new(hash);
+    let hash_str = as_hex.as_str();
```
CleanCut: A couple things you could do for ergonomics if you wanted to:

- Sometimes it is difficult to remember what separate type/function to use to convert something, but easy to look for conversion methods on the type you already have. We could add a method like this to `MerkleHash`:

```rust
    pub fn to_hex_hash(&self) -> HexHash {
        HexHash(format!("{self}"))
    }
```

- If you make a reference to `HexHash` act like a `&str`, then you don't need a method that does an explicit conversion anymore:

```rust
impl Deref for HexHash {
    type Target = str;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

If you do both above, then this code becomes:

```suggestion
    let hash_str = hash.to_hex_hash();
```

and then you could remove all of this:

```rust
impl HexHash {
    #[inline(always)]
    pub fn new(hash: &MerkleHash) -> Self {
        Self(format!("{hash}"))
    }

    pub(crate) fn as_str(&self) -> &str {
        &self.0
    }
}
```

CleanCut (approved):
There's a few obvious typos that a quick `cargo check` will let you fix. The `pre-commit` hooks must not have run.

crates/lib/src/core/db/merkle_node/merkle_node_db.rs
```diff
@@ -84,6 +91,45 @@ pub fn node_db_path(repo: &LocalRepository, hash: &MerkleHash) -> PathBuf {
...
+    #[error("Cannot encode a Merkle node: {0}")]
+    Encode(#[from] rmp_serde::encode::Error),
+    #[error("Cannot decode a Merkle node: {0}")]
+    Decode(#[from] rmp_serde::decode::Error),
+    #[error("{0}")]
+    TypeMismatch(#[from] InvalidMerkleTreeNodeType),
+    #[error("Failed to create directory: {0}")]
+    DirCreate(Box<OxenError>), // TODO: replace with FsError from upcoming refactoring PR
+    #[error("Failed to open file: {0}")]
+    Open(Box<OxenError>), // TODO: replace with FsError from upcoming refactoring PR
```
malcolmgreaves, line 107: In progress here: https://github.com/Oxen-AI/Oxen/pull/471

crates/lib/src/core/db/merkle_node/merkle_node_db.rs
malcolmgreaves: I like `to_hex_hash` for `MekleHash` -- I'll add that one.

For `&str`, I want to be careful with this. In general, the intention is to not treat a `HexHash` as a string. I want there to be a little bit of friction so that a string can't be confused for a hexidecimal formatted `MerkleHash` value. With that context, I think having a `Deref` implementation for `HexHash` will make it too easy to break the type safety of the `HexHash` wrapper.

I added the `as_str` to make this prefix-suffix split function avoid two `String` allocations. 🤔 I could instead move this logic into `HexHash`. That way I can avoid having `as_str` all together and have the function return the 2-layered nested directories directly. I think this makes sense since this directory structure is a primary use case of hex-formatting the hash value.

PR #472, opened 2026-04-21
Add merkle tree interfaces for reading & writing

malcolmgreaves:
New interface traits: `MerkleReader`, `MerkleWriter`, and unifying `MerkleStore`.

**`MerkleStore`**
A `MerkleStore` is just a `MerkleReader` + `MerkleWriter` where their `Error` type
is the same. Includes a blanket impl. for any type that has a reader & writer impl.
with equal `Error` types.

**`MerkleReader`**
Can retrieve Merkle tree nodes from storage given a hash. Supports existence
checks, retrieving the node, and retrieving a node's children, if applicable.

**`MerkleWriter`**
Produces a `MerkleWriteSession` that allows for writing arbitrary Merkle tree nodes
to the underlying physical store. Within the write session, writing a node creates a
new `NodeWriteSession` instance. This instance can write a node as well as a node's
children. Both session types have `finish`, which ensures their writes go to the
underlying physical store. 

These tiered write sessions match how the Oxen code writes Merkle tree nodes.
Additionally, this design allows for implementations to be eager, buffer writes in-
memory, or provide transaction semantics.

### DIFF
--- a/crates/lib/src/model/merkle_tree/merkle_writer.rs
+++ b/crates/lib/src/model/merkle_tree/merkle_writer.rs
@@ -1,0 +1,93 @@
+use crate::error::IntoOxenError;
+use crate::model::{MerkleHash, TMerkleTreeNode};
+
+/// Interface for writing to a Merkle tree store.
+pub trait MerkleWriter: Send + Sync {
+    /// The error type for the Merkle tree's underlying storage layer.
+    ///
+    /// Backends may use whichever error type is natural for their storage
+    /// (e.g. [`MerkleDbError`] for the [`FileBackend`]). The `Into<OxenError>`
+    /// bound on the associated type propagates as an implied bound at every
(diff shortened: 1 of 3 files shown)


crates/lib/src/model/merkle_tree/merkle_writer.rs
malcolmgreaves, line 1: _Why are there lifetimes here?_ The upcoming LMDB implementation will need to maintain a `heed::RwTxn` for some explicit lifetime in the `MerkleWriteSession`. That same read-write transaction will be borrowed in the `NodeWriteSession` too, so that also needs to support an explicit lifetime.

malcolmgreaves:
@coderabbitai please re-review this entire PR

malcolmgreaves:
STACKED PR: Do not merge until https://github.com/Oxen-AI/Oxen/pull/496 has been merged.

PR #473, opened 2026-04-21
Add per-repo migration run endpoint

jcelliott:
Add migration endpoint so external services can manage migrations.


CleanCut (approved):
Sometime soon we need to go over the concurrency model for repo operations. i.e. How do we handle concurrent writes+writes, or writes+reads?

PR #474, opened 2026-04-21
[migration] Make hex-encoded hash always 32 characters

malcolmgreaves:
Changes `MerkleHash`'s string representation to always be 32 characters.
In hexidecimal, every digit encodes 4 bits: 128 bits / 4 = 32 digits
"ffffffffffffffffffffffffffffffff" is u128::MAX. This makes the conversion
stable for all possible `MerkleHash` values.

This fixes a bug in the hex-hash `{hash prefix}/{hash suffix}` directory
creation method. There, a hash value of `0` was being converted to `"0"`,
meaning that the directories were ill-formated.

This change requires a migration, because existing reposistories could have
a different `prefix` written (_no leading zeros written_). A new migration
command, `m20260422_zero_pad_hash_dirs`, has been added for this situation.

### DIFF
--- a/crates/cli/src/cmd/migrate.rs
+++ b/crates/cli/src/cmd/migrate.rs
@@ -2,6 +2,7 @@ use std::path::Path;
 
 use async_trait::async_trait;
 use clap::{Arg, Command};
+use liboxen::command::migrate::set_zero_pad_hash_dirs_skip_collision_check;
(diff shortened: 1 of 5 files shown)


CleanCut (approved):

malcolmgreaves:
Pausing this PR -- it requires a migration of existing repositories.

PR #475, opened 2026-04-21
Add `--config-dir` flag for `oxen` and `oxen-server`

CleanCut:
The goals of this PR are:

1. Make a way to specify the config dir
2. Replace the `TEST` env var hack with specifying the config dir.

This is the followup to https://github.com/Oxen-AI/Oxen/pull/462


malcolmgreaves (commented):
.

malcolmgreaves (approved):
`--config-dir` is a great addition to the CLI + I'm _very_ glad we're getting rid of `TEST`!

One thing for follow up later (_not necessarily by you, unless you want to take a crack at this! :)_) --_ it's a larger change, but we really should delete the entire notion of a stateful user and auth in the library. Every place where we call `::get()` should be changed to accept the `*Config` as a parameter. This change will mesh well with this PR's changes because we'll go from (env var OR --config-dir) --> `{User,Auth}Config` -> use throughout.

PR #479, opened 2026-04-22
Add `oxen clean` command to remove untracked files and directories

CleanCut:
Adds a first-class equivalent of `git clean -fd` scoped to Oxen working trees. Two drivers:

- ENG-888 recovery: an interrupted `oxen pull` leaves untracked files the next pull refuses to merge over, and customers currently work around this with a custom bash script. `oxen clean -f` replaces the script.
- ENG-924 recently removed a silent-deletion side effect from `oxen restore <dir>`, which left Oxen with no supported way to remove untracked content. `oxen clean` is the sanctioned home for that operation.

Behavior:

- `oxen clean` with no flags is a dry-run — prints `Would remove <path>` per candidate and makes no filesystem changes. `oxen clean -f` actually removes.
- Untracked directories are always included (no `-d` flag). `oxen status` already classifies a directory as "untracked" only when its whole subtree is untracked, so this is safe by construction.
- `.oxen/` is never touched — a defense-in-depth check on top of status's own exclusion.
- `.oxenignore`-matched files are always preserved; the filter is inherited from status rather than reapplied.
- Optional positional `[paths...]` scope the operation.

Fixes ENG-925.

### DIFF
--- a/crates/lib/src/core/v_latest/clean.rs
+++ b/crates/lib/src/core/v_latest/clean.rs
@@ -5,6 +5,7 @@
 
 use std::path::{Path, PathBuf};
 
+use async_walkdir::WalkDir;
 use futures::{StreamExt, TryStreamExt, stream};
 
 use crate::constants::OXEN_HIDDEN_DIR;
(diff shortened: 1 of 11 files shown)


crates/lib/src/core/v_latest/clean.rs
```diff
@@ -0,0 +1,114 @@
...
+                total_bytes = total_bytes.saturating_add(meta.len());
+            }
+        }
+    }
+
+    // Clean up files and directories
+    if opts.force {
+        for f in &files {
+            println!("Removing {}", f.display());
+            util::fs::remove_file(repo.path.join(f))?;
```
malcolmgreaves: Minor, non blocking, just a mild perf consideration: since this function is async, this remove file and the remove dir all functions below could also be async.

Alternatively, I don't see any uses of async in this function, so it could be changed to non-async.

crates/lib/src/core/v_latest/clean.rs
```diff
@@ -0,0 +1,114 @@
...
+    // Calculate total bytes of files and directories to be cleaned
+    let mut total_bytes = 0u64;
+    for f in &files {
+        if let Ok(meta) = util::fs::metadata(repo.path.join(f)) {
+            total_bytes = total_bytes.saturating_add(meta.len());
+        }
+    }
+    for d in &dirs {
+        let abs = repo.path.join(d);
+        for entry in jwalk::WalkDir::new(&abs).into_iter().filter_map(|e| e.ok()) {
```
malcolmgreaves: There is an async walk dir crate in the dev dependencies FYI.

malcolmgreaves (approved):
Nice implementation and nice comments! Minor non blocking feedback. I know this is a priority feature so I'm happy to get it out the door ASAP.

What's the rationale for requiring --force for changes versus having a --dry-run flag? Either approach works. I'd say that it's valuable to stay consistent with what the other oxen commands have as a convention around these flags.

CleanCut:
> What's the rationale for requiring --force for changes versus having a --dry-run flag? Either approach works. I'd say that it's valuable to stay consistent with what the other oxen commands have as a convention around these flags.

Just to be more similar to `git` where it makes sense. My understanding from Greg & Josh is that we should mostly mimic `git` unless we have reasons not to. If anyone has strong feelings about the behavior of this subcommand, I'm fine changing it.

crates/lib/src/core/v_latest/clean.rs
CleanCut: Nice catch! I completely spaced that!

✅ Updated to run batches of concurrent, async file operations.

crates/lib/src/core/v_latest/clean.rs
CleanCut: Another good catch!

✅ Switched to async

PR #481, opened 2026-04-22
Resume interrupted pulls when files already match merge target

CleanCut:
We were comparing the working-tree file only to `HEAD`, but when a prior pull was `SIGTERM`'d after restoring files to the merge target but before advancing `HEAD,` those successfully-restored files looked "modified from `HEAD"` and the next pull failed citing that it could not overwrite changed files, even though the file was what we wanted to change _to_.

This PR changes the logic so that during the merge phase of a pull, we also compare the file in the working directory to the file in the target branch, and if they are the same, then we continue. This allows us to re-pull if a previous pull was interrupted in the merge phase, at least for files that were completely updated to the target.

Closes ENG-888.

### DIFF
--- a/crates/lib/src/core/v_latest/index/restore.rs
+++ b/crates/lib/src/core/v_latest/index/restore.rs
@@ -246,6 +246,11 @@ pub fn should_restore_partial_node(
             // If modified times are different, check hashes
             let hash = MerkleHash::new(util::hasher::u128_hash_file_contents(&working_path)?);
 
+            // File already matches the merge target — no-op.
+            if hash == *file_node.hash() {
+                return Ok(true);
+            }
+
             let base_node_hash = base_node.hash;
             if hash != base_node_hash {
                 return Ok(false);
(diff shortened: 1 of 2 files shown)


crates/lib/src/core/v_latest/index/restore.rs
malcolmgreaves, line 1: 👏 Excellent in-line comments: they really illustrate what's going on. Makes reading & reviewing the code a lot easier! 🚀

crates/lib/src/repositories/merge.rs
```diff
@@ -2151,4 +2151,53 @@ mod tests {
...
+            let merge_branch = repositories::branches::current_branch(&repo)?.unwrap();
+
+            // Back on main: checkout restores world.txt to commit A's content.
+            repositories::checkout(&repo, &og_branch.name).await?;
+            assert_eq!(util::fs::read_from_path(&world_file)?, "World");
+
+            // Simulate an interrupted fast-forward merge: the previous pull
+            // rewrote world.txt to the target's content on disk but crashed
+            // before advancing HEAD. HEAD still points to commit A.
+            util::fs::write_to_path(&world_file, "World2")?;
```
malcolmgreaves, line 2186: Clever! 👏

malcolmgreaves (approved):

PR #482, opened 2026-04-22
Add Python bindings for `oxen clean`

CleanCut:
Added Python bindings for the `oxen clean` CLI command we added recently.

```python
repo.clean()            # dry-run — returns a CleanResult listing candidates
repo.clean(force=True)  # actually delete
```

Fixes ENG-926.

### DIFF
--- a/oxen-python/tests/test_clean.py
+++ b/oxen-python/tests/test_clean.py
@@ -28,7 +28,7 @@ def test_clean_default_is_dry_run(tmp_path):
 
     assert result.applied is False
     assert os.path.exists(untracked), "dry-run must not delete anything"
-    assert "scratch.txt" in {os.path.basename(p) for p in result.files}
+    assert any(filter(lambda p: os.path.basename(p) == "scratch.txt", result.files))
 
 
 def test_clean_force_removes_untracked_file(tmp_path):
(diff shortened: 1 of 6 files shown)


oxen-python/tests/test_clean.py
```diff
@@ -0,0 +1,110 @@
...
+    repo, _ = _init_repo_with_one_tracked_file(str(tmp_path / "r"))
+    untracked = os.path.join(repo.path, "scratch.txt")
+    with open(untracked, "w") as f:
+        f.write("stale")
+
+    result = repo.clean()
+
+    assert result.applied is False
+    assert os.path.exists(untracked), "dry-run must not delete anything"
+    assert "scratch.txt" in {os.path.basename(p) for p in result.files}
```
malcolmgreaves: Very minor -- you can avoid the `set` allocation by using `any` and `filter`:

```python
assert any(filter(lambda p: os.path.basename(p) == "scratch.txt", result.files))
```

`any` will return `True` and stop evaluating the generator the first time it encounters something that is `True`.

malcolmgreaves (approved):

oxen-python/tests/test_clean.py
CleanCut: ✅ Done

PR #488, opened 2026-04-24
Share `reqwest::Client` instead of rebuilding per request

CleanCut:
Every API helper in `liboxen` was building a fresh `reqwest::Client` for every request, so reqwest's connection pool, TLS session resumption, (and possibly in the future, HTTP/2 multiplexing) never got a chance to kick in. Every request established its own new connection and went through the whole handshake process.

It was pretty straightforward to switch to reuse a cached client per host/user-agent combination.

Closes ENG-938.

### DIFF
--- a/crates/lib/src/api/client.rs
+++ b/crates/lib/src/api/client.rs
@@ -97,7 +97,7 @@ fn new_for_host(host: String, should_add_user_agent: bool) -> Result<Client, Oxe
     // Double-check under the write lock; another thread may have inserted while we built.
     let mut cache = CLIENT_CACHE
         .write()
-        .map_err(|_| OxenError::client_cache_poisoned())?;
+        .map_err(|_| OxenError::ClientCachePoisoned)?;
     Ok(cache.entry(key).or_insert(client).clone())
 }
 
--- a/crates/lib/src/api/client.rs
+++ b/crates/lib/src/api/client.rs
@@ -12,6 +12,8 @@ use crate::view::http;
 pub use reqwest::Url;
 use reqwest::retry;
 use reqwest::{Client, ClientBuilder, header};
+use std::collections::HashMap;
+use std::sync::{LazyLock, RwLock};
 use std::time;
 
 pub mod branches;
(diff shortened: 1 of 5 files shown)


crates/lib/src/error.rs
```diff
@@ -593,6 +598,11 @@ impl OxenError {
         OxenError::InternalError(StringError::from(s.as_ref()))
     }
 
+    /// Make a new OxenError::ClientCachePoisoned error.
+    pub fn client_cache_poisoned() -> Self {
+        OxenError::ClientCachePoisoned
+    }
+
```
malcolmgreaves: Minor suggestion -- It's about the same length to type `OxenError:: client_cache_poisoned` vs. `OxenError:: ClientCachePoisoned`, but IMO there's a bit of a benefit to using the variant directly since the function can only express the return type as the entire enum.
```suggestion
```

crates/lib/src/api/client.rs
```diff
@@ -62,30 +76,46 @@ pub fn new_for_url_no_user_agent(url: &str) -> Result<Client, OxenError> {
...
+    // Slow path: build outside the write lock so concurrent first-time callers
+    // for *different* hosts don't serialize on one mutex.
+    let client = builder_for_host(host, should_add_user_agent)?
+        .timeout(time::Duration::from_secs(constants::timeout()))
+        .build()?;
+
+    // Double-check under the write lock; another thread may have inserted while we built.
+    let mut cache = CLIENT_CACHE
+        .write()
+        .map_err(|_| OxenError::client_cache_poisoned())?;
```
malcolmgreaves: Minor suggestion -- 
```suggestion
        .map_err(|_| OxenError::ClientCachePoisoned)?;
```

malcolmgreaves (commented):
What will happen to a long running `oxen-server`? It seems like this new `CLIENT_CACHE` will grow unbounded and there doesn't appear to be any mechanism that will ever remove a client?

I would like to ensure that this cache of clients will evict old or otherwise unused clients. It's a great idea to not re-create HTTP clients and reuse them.

Wouldn't it be more direct to instead refactor the codebase so that the HTTP client is passed along to the functions that require it? For instance, in the client, we are short lived, so we only need to create a client once and use that. IIUC the client wouldn't have a need for the cache.

On the server, once a client disconnects, wouldn't this be the right moment to close the HTTP client and no longer keep it around?

jcelliott:
@malcolmgreaves The client module is only used by the CLI (and Python library), so we shouldn't need to worry about this cache growing. This looks like a great optimization for client operations (especially for long-running push/pull ops) 👍

malcolmgreaves:
Oh thanks for pointing that out, I completely missed that here. 😅  We can leave the refactoring for another day!

malcolmgreaves (approved):

crates/lib/src/error.rs
CleanCut: ✅ Agreed

crates/lib/src/api/client.rs
CleanCut: ✅ Done

PR #489, opened 2026-04-25
Fix add-row panic and null-drop on workspace list columns

jcelliott:
Adding a row to a typed list column (`list[u32]`, `list[i8]`, `list[date]`, etc.) panicked the server because `value_to_tosql` only handled six element dtypes explicitly and fell through to `panic!` for the rest. Lists of strings or booleans also silently dropped null elements via `into_iter().flatten()`.

Replace the per-dtype branches in `any_val_to_json` and `value_to_tosql` with a single `series_to_json_array` helper that walks the polars `Series` element-by- element and renders `None` as JSON `null`. Wrap list/struct/embedding placeholders in `INSERT`/`UPDATE` SQL with `CAST(? AS <sql_type>)` so the bind target is unambiguous instead of relying on DuckDB's implicit string-to-list cast.

Apply the same placeholder logic to `modify_row_with_polars_df` and `modify_rows_with_polars_df` so update has parity with insert.


gschoeni (approved):
Validated and tested locally 👍

PR #491, opened 2026-04-28
Publish oxen-cli & oxen-server to crates.io

malcolmgreaves:
Changes the publish step to publish all crates in the workspace that have `publish=true`
in their `Cargo.toml`s. This now publishes `oxen-cli` (crates/cli) and `oxen-server`
(crates/server) to crates.io. `oxen-py` is **not** published.

Updates the `Cargo.toml` to make `liboxen` a workspace dependency for the CLI and server.
This allows `liboxen` to be a `path` dependency during local development and pinned to
a specific version for resolution on crates.io.

The `bin/bump-version` script has been updated to update the version in `Cargo.toml`
for the `liboxen` workspace dependency.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,9 +41,16 @@ jobs:
           args: --skip-existing ${{ github.workspace }}/oxen-wheels/*.whl
           maturin-version: v1.8.5
 
-  publish_liboxen_crate:
-    # See https://crates.io/docs/trusted-publishing for details
-    name: 🦀 Publish liboxen crate to crates.io
+  publish_crates:
+    # Publishes every workspace crate with `publish = true` (the default) to crates.io:
+    #   - liboxen        → library crate
+    #   - oxen-cli       → installs the `oxen` binary via `cargo install oxen-cli`
+    #   - oxen-server    → installs the `oxen-server` binary via `cargo install oxen-server`
+    # `oxen-py` is skipped automatically because its Cargo.toml sets `publish = false`.
+    # cargo publishes in topological dep order (liboxen first) and waits for index propagation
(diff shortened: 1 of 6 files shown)


.github/workflows/publish.yml
```diff
@@ -59,6 +59,47 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: rust-lang/crates-io-auth-action@v1
+        id: auth
+      - name: Upload to crates.io
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
+        run: |
+          cd ${{ github.workspace }}/crates/server
+          cargo publish
```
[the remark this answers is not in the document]
malcolmgreaves: invalid:
```
error: unexpected argument '--idempotent' found
```

.github/workflows/publish.yml
```diff
@@ -59,6 +59,47 @@ jobs:
...
+
```
CleanCut: ⚠️ You don't need extra steps for each crate. Change the existing step to use `cargo publish --workspace` and adjust the step name to say we're publishing all the crates. Then the step will apply to all crates with `publish = true` in their `Cargo.toml` files.

CleanCut (approved):

.github/workflows/publish.yml
malcolmgreaves: Thanks for the feedback! I will update.

CleanCut: This also lets us rely on `cargo` to do all the right things with publishing dependencies first, making sure they're available, then publishing the next crates that depend on them, etc. Managing the dependencies and waiting steps in a workflow is...not fun.

PR #499, opened 2026-04-29
Convert inferred JSON[] columns to VARCHAR[] in jsonl data frame sources

jcelliott:


gschoeni (approved):
tested and verified locally 👍

PR #503, opened 2026-04-29
Add cat command to print the file contents, similar to git cat-file

haifengkao:
I know nothing about Rust. Codex wrote the whole stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5405,6 +5405,7 @@ dependencies = [
  "tempfile",
  "time",
  "tokio",
+ "tokio-stream",
  "tracing",
  "url",
  "uuid",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -40,6 +40,7 @@ serde_json = { workspace = true }
 tempfile = { workspace = true }
 time = { workspace = true }
 tokio = { workspace = true }
+tokio-stream = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true }
 uuid = { workspace = true }
--- a/crates/cli/src/cmd.rs
+++ b/crates/cli/src/cmd.rs
@@ -12,6 +12,9 @@ pub use branch::BranchCmd;
 pub mod checkout;
 pub use checkout::CheckoutCmd;
 
+pub mod cat;
+pub use cat::CatCmd;
+
 pub mod clean;
 pub use clean::CleanCmd;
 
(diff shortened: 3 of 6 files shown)


malcolmgreaves (commented):
Hi @haifengkao 👋 Thanks for taking the time to make a pull request into Oxen! This looks like it would be convenient for small text files.

I have a few questions about this change. Have you thought about what the intended UX should be if this command is run on a very large text file? What about on non-text binary data? IIRC running `cat` on binary data will often garble the terminal.

Is there a specific use case you have in mind that motivated this change? Would e.g. using `oxen download -o /tmp/contents && cat /tmp/contents && rm /tmp/contents` work? Or even perhaps `oxen download -o /dev/stdout`?

haifengkao:
@malcolmgreaves 

The main use case is a VS Code extension for mixed text/binary repositories, not interactive terminal browsing.

  The workflow I am targeting is game modding: a mod can contain both text files and binary assets. For text files, the extension wants to show diffs.
  For binary files, it often only needs to know that the file changed, or fetch the bytes for a custom viewer/comparison. To support that, the extension needs a way to read a file from a local Oxen repo at a specific revision or from the staged area, without restoring it into the worktree and without  creating/managing a temporary file.

  `oxen cat` is intended as a plumbing command for that purpose, similar in spirit to `git cat-file -p <rev>:<path>`. It emits the file contents to stdout so a tool can capture them.

  For very large files or binary data, the extension/UI layer should usually inspect metadata first and decide whether to call this command or how to display the result. The terminal-garbling concern is real if a human runs it directly in an interactive terminal, but that is not the primary intended UX for this command.

  `oxen download -o /tmp/contents && cat /tmp/contents && rm /tmp/contents` does not cover this use case well because it targets remote downloads, writes
  to disk, and does not address local revision or staged reads. `oxen download -o /dev/stdout` is also Unix-specific and still belongs to the remote-download workflow.
  
  (cat is not enough, I need --json as well)

CleanCut:
Hi, @haifengkao! I'll be moving forward with this feature. Since the code has moved forward quite a bit since you submitted this PR, I took the liberty of re-implementing your request in #667, so I'll close this PR.

> (cat is not enough, I need --json as well)

I'm guessing you mean you need the metadata in json format? You can get that from `oxen info` once #668 is merged!

CleanCut:
Included in [v0.51.1](https://github.com/Oxen-AI/Oxen/releases/tag/v0.51.1)

PR #504, opened 2026-04-29
Implemented Merkle{Packer,Unpacker} for FileBackend

malcolmgreaves:
**`MerkleTransport`**
Implemented the new `MerkleTransport` traits for packing up
and unpacking Merkle tree nodes as they're sent on the wire
from the oxen client to server. Preserves the existing tar-gz
formats specific to the FileBackend's physical store layout.

Supports the two unique tar-gz paths:
- upload: only captures `{prefix}/{suffix}/{node,children}`
- everything else: `tree/nodes/{prefix}/{suffix}/{node,children}`

**Progress Bar in Bytes**
Changes the progress bar for uploads & downloads to use bytes/s
instead of nodes/s. Nodes are not evenly sized, so progress will
look like it stops when handling a large file. This now hooks
into the reader & writer types that the trait uses.

### DIFF
--- a/crates/lib/src/core/db/merkle_node/file_backend.rs
+++ b/crates/lib/src/core/db/merkle_node/file_backend.rs
@@ -1,19 +1,37 @@
-use crate::core::db::merkle_node::merkle_node_db::{MerkleDbError, MerkleNodeDB};
+use std::collections::HashSet;
+use std::io::{Read, Write};
+use std::path::{Component, Path, PathBuf};
+
+use flate2::Compression;
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
+use tar::Archive;
+use tempfile::TempDir;
+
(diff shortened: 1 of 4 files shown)


crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -216,6 +575,987 @@ mod tests {
         })
     }
 
+    /// Gunzip + collect tar entries into a deterministic map for byte-compat comparison.
```
malcolmgreaves, line 578: All test code helpers from here down. This is quite verbose, but the point is to extract the existing logic for Merkle tree node packing & unpacking on `main` and preserve that here. This way, we can run both the copied old code as well as the newly refactored code and prove that they behave the same.

malcolmgreaves:
STACKED PR: Do not merge until https://github.com/Oxen-AI/Oxen/pull/502 has been merged.

PR #510, opened 2026-05-01
Add merkle tree interfaces for reading & writing

malcolmgreaves:
New interface traits: `MerkleReader`, `MerkleWriter`, and unifying `MerkleStore`.

**`MerkleStore`**
A `MerkleStore` is just a `MerkleReader` + `MerkleWriter` where their `Error` type
is the same. Includes a blanket impl. for any type that has a reader & writer impl.
with equal `Error` types.

**`MerkleReader`**
Can retrieve Merkle tree nodes from storage given a hash. Supports existence
checks, retrieving the node, and retrieving a node's children, if applicable.

**`MerkleWriter`**
Produces a `MerkleWriteSession` that allows for writing arbitrary Merkle tree nodes
to the underlying physical store. Within the write session, writing a node creates a
new `NodeWriteSession` instance. This instance can write a node as well as a node's
children. Both session types have `finish`, which ensures their writes go to the
underlying physical store.

These tiered write sessions match how the Oxen code writes Merkle tree nodes.
Additionally, this design allows for implementations to be eager, buffer writes in-
memory, or provide transaction semantics.

### DIFF
--- a/crates/lib/src/model/merkle_tree/merkle_reader.rs
+++ b/crates/lib/src/model/merkle_tree/merkle_reader.rs
@@ -1,0 +1,60 @@
+use crate::error::OxenError;
+use crate::model::{
+    MerkleHash,
+    merkle_tree::node::{EMerkleTreeNode, MerkleTreeNode},
+};
+
+/// Interface for read-only access to Merkle tree nodes representing commits & directories.
+///
+/// The `exists`, `get_node`, and `get_children` methods only work on [`MerkleHash`] values
+/// that map to commits or directories (and by extension, virtual directory nodes).
+///
+/// A file node is _always_ stored in the `children` of some virtual node. To access a Merkle
+/// tree node for a file, one needs to map from the repository relative filepath and a commit
+/// hash to the right virtual node hash, call [`get_children`] on it, then iterate to the
+/// file's corresponding [`MerkleTreeNode`].
+///
+/// Dyn-compatible: callers can store use this as `dyn MerkleReader`.
+pub trait MerkleReader: Send + Sync {
+    /// True if there is some node with the given hash. False otherwise.
+    /// An error is returned if there is some other failure in the Merkle tree's underlying storage layer.
+    fn exists(&self, hash: &MerkleHash) -> Result<bool, OxenError>;
+
+    /// Retrieve the node record for the given hash, if it exists. None means no such node exists.
+    /// Note that a file node's [`MerkleHash`] will result in None: file nodes are stored as _children_.
+    /// An error is returned if there is some other failure in the Merkle tree's underlying storage layer.
+    fn get_node(&self, hash: &MerkleHash) -> Result<Option<MerkleEntry>, OxenError>;
+
+    /// Retrieve the children of the node for the given hash, if it exists and if it is a directory node.
+    /// If the node represents a file, then an empty list is always returned.
+    /// An error is returned if there is some other failure in the Merkle tree's underlying storage layer.
+    fn get_children(
+        &self,
+        hash: &MerkleHash,
+    ) -> Result<Vec<(MerkleHash, MerkleTreeNode)>, OxenError>;
+
+    /// Load a [`MerkleTreeNode`] with full node info and 1-level (aka direct) children for any non-file node.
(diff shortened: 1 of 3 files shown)


malcolmgreaves:
NOTE: Stacked PR! Must merge https://github.com/Oxen-AI/Oxen/pull/523 before merging.

crates/lib/src/model/merkle_tree/merkle_reader.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// If the node represents a file, then an empty list is always returned.
+    /// An error is returned if there is some other failure in the Merkle tree's underlying storage layer.
+    fn get_children(
+        &self,
+        hash: &MerkleHash,
+    ) -> Result<Vec<(MerkleHash, MerkleTreeNode)>, OxenError>;
+}
+
+/// Metadata returned when reading a single node.
+pub struct MerkleNodeRecord {
```
CleanCut: This struct is redundant. We already have `struct MerkleTreeNode` that contains the same information. You can get the "`dtype`"-equivalent from `MerkleTreeNode.node.node_type()`.

crates/lib/src/model/merkle_tree/merkle_reader.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn exists(&self, hash: &MerkleHash) -> Result<bool, OxenError>;
+
+    /// Retrieve the node record for the given hash, if it exists. None means no such node exists.
+    /// An error is returned if there is some other failure in the Merkle tree's underlying storage layer.
+    fn get_node(&self, hash: &MerkleHash) -> Result<Option<MerkleNodeRecord>, OxenError>;
+
+    /// Retrieve the children of the node for the given hash, if it exists and if it is a directory node.
+    /// If the node represents a file, then an empty list is always returned.
+    /// An error is returned if there is some other failure in the Merkle tree's underlying storage layer.
+    fn get_children(
```
CleanCut, line 31: Check the comment here for accuracy. There are nodes other than directories that have children.

crates/lib/src/model/merkle_tree/merkle_reader.rs
```diff
@@ -0,0 +1,73 @@
...
+        &self,
+        hash: &MerkleHash,
+    ) -> Result<Vec<(MerkleHash, MerkleTreeNode)>, OxenError>;
```
CleanCut, line 34: This should probably return the same inner node type as `get_node`

crates/lib/src/model/merkle_tree/merkle_writer.rs
```diff
@@ -0,0 +1,84 @@
...
+/// guarantee is that all node and child information must be persisted to the store.
+///
+/// Object-safe: lives behind `Box<dyn NodeWriteSession + '_>`. `finish` takes
+/// `self: Box<Self>` for the same reason as [`MerkleWriteSession::finish`].
+pub trait NodeWriteSession {
+    /// The hash of the node being written in this session.
+    fn node_id(&self) -> &MerkleHash;
+
+    /// Add a child to the current node.
+    fn add_child(&mut self, child: &dyn TMerkleTreeNode) -> Result<(), OxenError>;
```
CleanCut, line 78: `child` will need to be some other type here. Probably something like `MerkleTreeNode`.  `TMerkleTreeNode` is a serialization trait through which you can only access raw bytes.

CleanCut (approved):
👍🏻 This implementation can work.

🎁 There are some opportunities for simplification that you should consider:

- Just like with the `VersionStore` trait, the trait design here can be simplified from 5 traits to just one: `MerkleStore`
  - A single trait design can still allow for implementations to be eager, buffer writes in-memory, and provide transaction semantics if necessary.
  - The trait implementation can internally handle transaction semantics within its methods, rather than producing nested types representing various sessions that the user must create/use/finalize manually.
  - Inner transactions/locking can be handled with private fields on the implementation struct, and happen within `create_node`, `add_child`, (etc.) implementations.
  - The current design allows you to drop the structs without calling `finalize()`. It would be easier to avoid problems if whatever `finalize()` is supposed to do is folded into the internal transaction implementation.
- `MerkleNodeRecord` is redundant (see comment below)

PR #512, opened 2026-05-01
FileBackend for Merkle{Reader,Writer}

malcolmgreaves:
Implements the current custom file format based Merkle tree node storage
and retrieval logic for the `MerkleReader` and `MerkleWriter` traits as
the new `FileBackend` type. The implementation internally uses `MerkleDbError`,
but this is transformed at the trait boundary into an `OxenError`.

The write sessions that `FileBackend` produces are `Box`'d because they are
required to be object-safe (aka dyn-compatible).

Also refactors `MerkleNodeDB::to_node`'s logic into a new function on the
enum `EMerkleTreeNode::from_type_and_bytes`

Additionally, `MerkleNodeDB`'s `open_read_write`, `write_node`, and `add_child`
switch from generic `<N: TMerkleTreeNode>` to `&dyn TMerkleTreeNode` so they're
callable through the dyn-compatible `NodeWriteSession`.

Finally, fixes a bug in `MerkleNodeDb::open_read_only`: it was previously
incorrectly setting the `node_id: MerkleHash` to `0`. The function now requires
the right hash from the path to be provided and uses that.

### DIFF
--- a/crates/lib/src/core/db/merkle_node/file_backend.rs
+++ b/crates/lib/src/core/db/merkle_node/file_backend.rs
@@ -1,0 +1,298 @@
+use std::path::{Path, PathBuf};
+
+use crate::core::db::merkle_node::merkle_node_db::{MerkleDbError, MerkleNodeDB};
+use crate::error::OxenError;
+use crate::model::merkle_tree::merkle_reader::{MerkleEntry, MerkleReader};
+use crate::model::merkle_tree::merkle_writer::{
+    MerkleWriteSession, MerkleWriter, NodeWriteSession,
+};
+use crate::model::merkle_tree::node::MerkleTreeNode;
+use crate::model::{LocalRepository, MerkleHash, TMerkleTreeNode};
+
+/// File-based Merkle node store backend. Implements the [`MerkleReader`] and
+/// [`MerkleWriter`] traits.
+///
+/// Borrows the path to a local repository (via the [`LocalRepository`]'s [`PathBuf`]) so it
+/// can delegate straight to [`MerkleNodeDB`]'s existing repository-based methods without any
+/// modification.
+#[derive(Debug)]
+pub struct FileBackend {
+    /// Location to the repository's root in the filesystem. Must be an absolute path.
+    pub(crate) repo_path: PathBuf,
+}
+
+impl FileBackend {
+    pub fn new(repo: &LocalRepository) -> Self {
+        Self {
+            repo_path: repo.path.clone(),
+        }
+    }
+}
+
+/// Merkle reader implementation for the [`FileBackend`].
+///
+/// NOTE: Uses MerkleDbError internally, but is unfortunately forced to convert into
+/// an [`OxenError`] at the trait boundary. It is a bug if any of these methods do
+/// not return a wrapped `MerkleDbError`.
+impl MerkleReader for FileBackend {
+    /// Checks if a node with the given `hash` exists in the store.
+    ///
+    /// Alias for [`MerkleNodeDB::exists`].
+    fn exists(&self, hash: &MerkleHash) -> Result<bool, OxenError> {
+        Ok(MerkleNodeDB::exists(&self.repo_path, hash))
+    }
+
+    /// Retrieves the node with the given `hash` from the store. `None` means no such node exists.
+    ///
+    /// Alias for [`MerkleNodeDB::open_read_only`].
+    fn get_node(&self, hash: &MerkleHash) -> Result<Option<MerkleEntry>, OxenError> {
+        if !MerkleNodeDB::exists(&self.repo_path, hash) {
+            return Ok(None);
+        }
+        let db = MerkleNodeDB::open_read_only(&self.repo_path, hash)?;
+        let node = db.node()?;
+        Ok(Some(MerkleEntry {
+            node,
+            parent_id: db.parent_id,
+        }))
+    }
+
+    /// Retrieves the children of the node with the given `hash` from the store.
+    /// An empty vec means that either the node is a not a directory or virtual node or it is one
+    /// but has no files.
+    ///
+    /// Alias for [`MerkleNodeDB::open_read_only`] & a `.map()` call on it.
+    fn get_children(
+        &self,
+        hash: &MerkleHash,
+    ) -> Result<Vec<(MerkleHash, MerkleTreeNode)>, OxenError> {
+        if !MerkleNodeDB::exists(&self.repo_path, hash) {
+            return Ok(Vec::with_capacity(0));
+        }
+        let mut db = MerkleNodeDB::open_read_only(&self.repo_path, hash)?;
+        let children = db.map()?;
+        Ok(children)
+    }
+
+    /// Load a [`MerkleTreeNode`] with full node info and 1-level (aka direct) children for any non-file node.
+    /// More efficient than doing a [`get_node`] and a [`get_children`] call.
+    fn read_full_node(&self, hash: &MerkleHash) -> Result<Option<MerkleTreeNode>, OxenError> {
+        if !MerkleNodeDB::exists(&self.repo_path, hash) {
+            return Ok(None);
+        }
+        let mut db = MerkleNodeDB::open_read_only(&self.repo_path, hash)?;
+        let node = db.node()?;
+        let children = db.map()?;
+        Ok(Some(MerkleTreeNode {
+            hash: *hash,
+            node,
+            parent_id: db.parent_id,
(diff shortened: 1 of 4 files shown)


crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+        let db = MerkleNodeDB::open_read_only(self.repo, hash)?;
+        let node = db.node()?;
+        Ok(Some(MerkleNodeRecord::new(
+            db.node_id,
+            db.dtype,
+            db.parent_id,
+            node,
+            db.num_children(),
+        )))
```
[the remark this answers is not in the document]
malcolmgreaves: Fixed this bug!

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -0,0 +1,301 @@
...
+    ///
+    /// Alias for [`MerkleNodeDB::open_read_only`] & a `.map()` call on it.
+    fn get_children(
+        &self,
+        hash: &MerkleHash,
+    ) -> Result<Vec<(MerkleHash, MerkleTreeNode)>, OxenError> {
+        let mut db = MerkleNodeDB::open_read_only(self.repo, hash)?;
+        let children = db.map()?;
+        Ok(children)
+    }
```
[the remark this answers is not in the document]
malcolmgreaves: Fixed.

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+}
+
+/// Ensure that the `node` and `children` file handles are flushed and closed when dropped.
+impl Drop for FileNodeSession {
+    fn drop(&mut self) {
+        self.idempotent_finish()
+            .expect("Did not explicitly call finish() and encountered an error.");
+    }
```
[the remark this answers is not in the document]
malcolmgreaves: The user is supposed to call finish. If they don't, we do a best-effort here. Silently failing is a bad idea.

malcolmgreaves:
NOTE: Stacked PR. Must merge https://github.com/Oxen-AI/Oxen/pull/510 before merging.

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -0,0 +1,311 @@
...
+}
```
malcolmgreaves: I am strongly considering removing this. It doesn't make _conceptual_ sense to finish on drop. 

Originally, I was thinking "make sure finish() is called if it is forgotten," but this _is not_ that. We'll finish even on an error, which we don't want on a philosophical / conceptual design level. 

The file backend writes eagerly and there's no "undo," so the writes will _probably_ still be there. This finish is ensuring the file handles are flushed and closed. But arguably we don't want to do this if we're erroring out 🤔

CleanCut: I think you're right that removing this is the cleaner call.

⚠️ If we do decide to keep it, we'd probably want to swap the `expect` for something non-panicking. Anything that can panic inside a `Drop` impl will abort the process (segfault) if it fires during unwinding, which is a nasty failure mode to leave around.

crates/lib/src/core/db/merkle_node/merkle_node_db.rs
```diff
@@ -126,16 +125,20 @@ struct MerkleNodeLookup {
 }
 
 impl MerkleNodeLookup {
-    pub fn load(node_table_file: &mut File) -> Result<Self, MerkleDbError> {
+    pub(crate) fn load(node_table_file: &mut File) -> Result<Self, MerkleDbError> {
```
CleanCut: Nit: `pub` and `pub(crate)` on a method of a private struct are effectively a no-op (the struct's visibility bounds it). Not worth fixing on its own.

```suggestion
    fn load(node_table_file: &mut File) -> Result<Self, MerkleDbError> {
```

crates/lib/src/model/merkle_tree/node.rs
```diff
@@ -72,4 +72,22 @@ impl EMerkleTreeNode {
...
+        data: &[u8],
+    ) -> Result<Self, rmp_serde::decode::Error> {
+        match dtype {
+            MerkleTreeNodeType::Commit => {
+                Ok(EMerkleTreeNode::Commit(CommitNode::deserialize(data)?))
+            }
+            MerkleTreeNodeType::Dir => Ok(EMerkleTreeNode::Directory(DirNode::deserialize(data)?)),
+            MerkleTreeNodeType::File => Ok(EMerkleTreeNode::File(FileNode::deserialize(data)?)),
+            MerkleTreeNodeType::VNode => Ok(EMerkleTreeNode::VNode(VNode::deserialize(data)?)),
+            MerkleTreeNodeType::FileChunk => Ok(EMerkleTreeNode::FileChunk(
```
CleanCut, line 88: 🗒️  Note: I think the `FileChunk` variant should be removed (in general, not in this PR). I couldn't find any usage of it, and conceptually it feels like an implementation detail of how a `File` is stored behind the `VersionStore` trait rather than a node in the merkle tree. If a file chunk were part of the merkle tree, the `File` hash (its parent) would shift depending on how we picked chunk boundaries, which seems undesirable.

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Retrieves the children of the node with the given `hash` from the store.
+    /// An empty vec means that either the node is a not a directory or virtual node or it is one
+    /// but has no files.
+    ///
+    /// Alias for [`MerkleNodeDB::open_read_only`] & a `.map()` call on it.
+    fn get_children(
+        &self,
+        hash: &MerkleHash,
+    ) -> Result<Vec<(MerkleHash, MerkleTreeNode)>, OxenError> {
```
CleanCut, line 68: The signature of this method (and the trait) could probably be simplified. `MerkleTreeNode` already contains its own `hash` field, so passing it alongside feels redundant.

```suggestion
    ) -> Result<Vec<MerkleTreeNode>, OxenError> {
```

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -0,0 +1,326 @@
...
+            children: children.into_iter().map(|(_, c)| c).collect(),
+        }))
+    }
+}
+
+/// Merkle writer implementation for the [`FileBackend`].
+///
+/// NOTE: Uses MerkleDbError internally, but is unfortunately forced to convert into
+/// an [`OxenError`] at the trait boundary. It is a bug if any of these methods do
+/// not return a wrapped `MerkleDbError`.
```
CleanCut, line 99: If you want `MerkleDbError` to cover all the error variants across the merkle implementations, you could return it from all these functions instead.

crates/lib/src/core/db/merkle_node/merkle_node_db.rs
```diff
@@ -144,27 +147,27 @@ impl MerkleNodeLookup {
         cursor.read_exact(&mut buffer)?;
         let node_data_type = u8::from_le_bytes(buffer);
         // log::debug!(
-        //     "MerkleNodeLookup.load() data_type: {:?}",
+        //     "MerkleNodeLookup.deserialize() data_type: {:?}",
         //     MerkleTreeNodeType::from_u8(node_data_type)
         // );
```
CleanCut, line 152: Nit: There's lots of commented-out logging here. It's probably worth either uncommenting or deleting it.

CleanCut (approved):
⚠️ It would be great to make a decision on which way to go on the `Drop` implementation (see comment below).

Otherwise, just minor stuff to consider.

crates/lib/src/core/db/merkle_node/file_backend.rs
malcolmgreaves: This is just how the existing code needs to interact with the Merkle store. This whole effort is a refactor, so I'm minimizing changes to existing code as much as possible.

crates/lib/src/model/merkle_tree/node.rs
malcolmgreaves: I'm not 100% sure on this, but I believe it exists so that we can roll out block level deduplication one day.

crates/lib/src/core/db/merkle_node/file_backend.rs
malcolmgreaves: Glad for the sanity check -- I'm removing `Drop` 🫡

PR #513, opened 2026-05-01
Refactor oxen's Merkle tree operations: use MerkleStore

malcolmgreaves:
**Refactor**
Change all reading & writing of Merkle tree nodes to use the new traits:
- Use `MerkleReader::exists` calls in `commit_merkle_tree.rs`
- Use `MerkleReader::get_{node,children}` in `merkle_tree_node.rs`
- Use `MerkleWriter`'s write sessions in:
    + `create_empty_commit` & `create_initial_commit` (`commits.rs`)
    + `update_metadata` (`entries.rs`)
    + `commit_dir_entires`, `write_commit_entires`, `r_create_dir_node` (`commit_writer.rs`)
    + `run_on_commit`, `rewrite_nodes` (migration `..add_child_counts_to_nodes.rs`)
    + `write_tree`, `p_write_tree` (`tree.rs`)

Access to a `MerkleStore` for these operations is provided via the `LocalRepository`.

**LocalRepository**
Adds method `LocalRepository::merkle_store() -> Box<dyn MerkleStore + '_>`. Provides the
repository's Merkle tree node storage implementation. The return type is dynamic: it
allows for selecting a compatible `MerkleStore` at runtime.

**`OxenError` Updates**
Converts 2 string errors into:
- `MerkleNodeNotFound(HexHash)`: no such commit/dir/vnode in lookup
- `DisallowedNodeWrite(std::any::type_name_of_val)`

### DIFF
--- a/crates/lib/src/command/migrate/m20250111083535_add_child_counts_to_nodes.rs
+++ b/crates/lib/src/command/migrate/m20250111083535_add_child_counts_to_nodes.rs
@@ -3,10 +3,10 @@ use std::path::Path;
 use super::Migrate;
 
 use crate::config::RepositoryConfig;
-use crate::core::db::merkle_node::MerkleNodeDB;
 use crate::core::versions::MinOxenVersion;
 use crate::error::OxenError;
 use crate::model::merkle_tree::merkle_tree_node_cache;
+use crate::model::merkle_tree::merkle_writer::MerkleWriteSession;
 use crate::model::merkle_tree::node::vnode::VNodeOpts;
 use crate::model::merkle_tree::node::{
     CommitNode, DirNode, EMerkleTreeNode, MerkleTreeNode, VNode,
@@ -135,14 +135,27 @@ fn run_on_commit(repository: &LocalRepository, commit: &Commit) -> Result<(), Ox
     dir_node_opts.num_entries = num_children as u64;
     let dir_node = DirNode::new(&new_repo, dir_node_opts)?;
 
-    // Write a new commit db
+    // Write a new commit node via a session on the old repo.
     let commit_node = CommitNode::from_commit(commit.clone());
-    let mut root_commit_db =
-        MerkleNodeDB::open_read_write(&old_repo.path, &commit_node, root_node.parent_id)?;
-    root_commit_db.add_child(&dir_node)?;
+    let old_store = old_repo.merkle_store();
+    let new_store = new_repo.merkle_store();
+    let old_session = old_store.begin()?;
+    let new_session = new_store.begin()?;
+    let mut root_commit_ns = old_session.create_node(&commit_node, root_node.parent_id)?;
+    root_commit_ns.add_child(&dir_node)?;
+    root_commit_ns.finish()?;
 
     let current_path = Path::new("");
-    rewrite_nodes(&old_repo, &new_repo, &root_node, current_path)?;
+    rewrite_nodes(
+        &old_repo,
+        &new_repo,
+        &*old_session,
+        &*new_session,
+        &root_node,
+        current_path,
+    )?;
+    new_session.finish()?;
+    old_session.finish()?;
 
     // println!("new tree for commit {}", commit);
     // repositories::tree::print_tree(&new_repo, commit)?;
@@ -158,9 +171,12 @@ fn run_on_commit(repository: &LocalRepository, commit: &Commit) -> Result<(), Ox
 
 // Forgive me if you are reading this for reference, we don't have great writers for the
 // merkle tree yet - so there is a lot of duplicate logic with `commit_writer.rs`
+#[allow(clippy::too_many_arguments)]
 fn rewrite_nodes(
     old_repo: &LocalRepository,
     new_repo: &LocalRepository,
+    old_session: &dyn MerkleWriteSession,
+    new_session: &dyn MerkleWriteSession,
     node: &MerkleTreeNode,
     current_dir: &Path,
 ) -> Result<(), OxenError> {
@@ -169,18 +185,9 @@ fn rewrite_nodes(
             EMerkleTreeNode::Directory(dir) => {
                 // Load all the children of children (files and folders)
                 // Then redistribute into buckets...
-                // and then just use the MerkleNodeDB to write the nodes
-                // to the new tree
+                // and then just use the session to write the nodes to the new tree
                 let dir_children = repositories::tree::list_files_and_folders(child)?;
                 let current_dir = current_dir.join(dir.name());
-
-                // log::debug!(
-                //     "rewrite_nodes {} children on current_dir {:?} DIRECTORY {} {}",
-                //     dir_children.len(),
-                //     current_dir,
-                //     dir.hash(),
-                //     dir
-                // );
 
                 let total_children = dir_children.len();
                 let vnode_size = old_repo.vnode_size();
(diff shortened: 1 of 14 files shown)


crates/lib/src/command/migrate/m20250111083535_add_child_counts_to_nodes.rs
```diff
@@ -190,7 +197,7 @@ fn rewrite_nodes(
                 let mut dir_node_opts = dir.get_opts();
                 dir_node_opts.num_entries = total_children as u64;
                 let dir = DirNode::new(new_repo, dir_node_opts)?;
-                let mut dir_db = MerkleNodeDB::open_read_write(old_repo, &dir, node.parent_id)?;
+                let mut dir_ns = old_session.create_node(&dir, node.parent_id)?;
```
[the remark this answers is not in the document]
malcolmgreaves: Incorrect.

crates/lib/src/repositories/commits/commit_writer.rs
```diff
@@ -794,25 +811,27 @@ fn write_commit_entries(
...
         commit_id,
-        &mut Some(dir_db),
+        Some(&mut *dir_ns),
         dir_hash_db,
         dir_hashes,
         entries,
         root_path,
         &mut total_written,
     )?;
+    dir_ns.finish()?;
```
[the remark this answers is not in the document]
malcolmgreaves: Not quite true. It does the flush, but the node writes are eager anyways. It's an implementation detail if the writes would be visible or not before this refactor. This refactor makes it consistent by ensuring the writes are flushed on drop. In practice, if there's some error encountered during this (before and after the refactor), we'd have a Merkle tree in some inconsistent state. We don't have any idea what state that'd be.

crates/lib/src/repositories/commits/commit_writer.rs
```diff
@@ -872,75 +892,56 @@ fn r_create_dir_node(
...
                         else {
-                            // log::debug!(
-                            //     "r_create_dir_node could not read old dir node {:?}",
-                            //     node.hash
-                            // );
+                            log::trace!(
+                                "r_create_dir_node could not read old dir node {:?}",
+                                node.hash(),
+                            );
                             continue;
```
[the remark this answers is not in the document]
malcolmgreaves: This is a refactor: existing behavior must be preserved.

crates/lib/src/repositories/tree.rs
```diff
@@ -1038,34 +1042,36 @@ pub fn write_tree(repo: &LocalRepository, node: &MerkleTreeNode) -> Result<(), O
...
-                db.add_child(vnode)?;
-                p_write_tree(repo, child, vnode)?;
+                ns.add_child(vnode)?;
+                p_write_tree(session, child, vnode)?;
             }
             EMerkleTreeNode::Directory(dir_node) => {
-                db.add_child(dir_node)?;
-                p_write_tree(repo, child, dir_node)?;
+                ns.add_child(dir_node)?;
+                p_write_tree(session, child, dir_node)?;
```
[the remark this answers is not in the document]
malcolmgreaves: Refactor is preserving existing behavior.

malcolmgreaves:
NOTE: Stacked PR! Must merge https://github.com/Oxen-AI/Oxen/pull/512 before merging.

crates/lib/src/storage/version_store.rs
```diff
@@ -148,7 +148,8 @@ impl Default for StorageConfig {
 
 /// Trait defining operations for version file storage backends
 #[async_trait]
-pub trait VersionStore: Debug + Send + Sync + 'static {
+pub trait VersionStore: Debug + Send + Sync {
+    // + 'static {
```
CleanCut: ⚠️  What's the motivation for dropping the `'static` bound on `VersionStore`? Relaxing trait bounds tends to be a one-way door, so I'd want to understand the need before we ship it. Otherwise it's probably worth restoring. The code still compiled fine when I added `'static` back in.

However, if it's going to be a pain to re-stack your PRs, I can re-add the bound in a separate PR.

```suggestion
pub trait VersionStore: Debug + Send + Sync + 'static {
```

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -23,6 +25,10 @@ use utoipa::ToSchema;
 static MTIME_TOLERANCE_CACHE: LazyLock<Mutex<HashMap<PathBuf, Duration>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 
+// TODO: A `LocalRepository` shouldn't require serialization.
```
CleanCut, line 37: Agreed! I'm fixing this (soon!) in the S3 project.

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -40,10 +46,15 @@ pub struct LocalRepository {
...
-    // Skip this field during serialization/deserialization
+    // Skip these fields during serialization/deserialization as it relates to
+    // state for the entire repository.
     #[serde(skip)]
     #[schema(ignore)]
     version_store: Option<Arc<dyn VersionStore>>,
+    // TODO: once the serialization requirement is fixed, then we can store and reuse the merkle_store
+    // #[serde(skip)]
+    // #[schema(ignore)]
+    // merkle_store: Arc<Box<dyn MerkleStore>>,
```
CleanCut: The inner `Box` is redundant here. `Arc` is already a heap allocation, and it adds shared ownership on top. We can drop the `Box`.

```suggestion
    // merkle_store: Arc<dyn MerkleStore>,
```

crates/lib/src/error.rs
```diff
@@ -228,6 +225,15 @@ pub enum OxenError {
     #[error("No changes to commit")]
     NoChanges,
 
+    #[error("No such commit, dir, or vnode Merkle tree node with hash (hex): {0}")]
+    MerkleNodeNotFound(HexHash),
```
CleanCut, line 248: What do you think about wrapping a MerkleHash here?

crates/lib/src/error.rs
```diff
@@ -228,6 +225,15 @@ pub enum OxenError {
...
+
+    #[error(
+        "Unsupported node type adding to a child file. Only accept Commit, Directory, File, or VNode. Found: {0}"
+    )]
+    /// Contains the name of the incompatible type as reported by [`std::any::type_name_of_val`].
+    DisallowedNodeWrite(&'static str),
```
CleanCut, line 254: Hey, I'm wondering if this is effectively acting like a `Basic` error here? It looks like the places it's used have a wildcard branch when matching on an enum. What do you think about removing the wildcard and matching each variant explicitly instead, so we wouldn't need to deal with an error at all?

CleanCut (approved):

crates/lib/src/error.rs
malcolmgreaves: This is a neat thing I added a little while ago! `HexHash` is a restriction on a `String`. It's only ever possible to create it by having a `MerkleHash`. There's `.into()` conversions between the two.

Why didn't I go with `MerkleHash` directly? I'm opting for the `HexHash` because it's primarily being used for the error message. The `.into()` conversion means we can recover the `MerkleHash` without any possibility of an error if we need to. (We'd have to open up the visibility a bit too ofc, but adding more visibility is always backwards compatible, so we can do that at anytime we need to!)

crates/lib/src/error.rs
malcolmgreaves: 🤔 I don't quite follow. Making a named variant to describe a specific error condition is the design goal. Perhaps this explanation will help?

The intention is that `DisallowedNodeWrite` is a unique error condition that happens because of a runtime constraint on how the tree writing code exists today. If you look at `v_latest/entires.rs` and `repositories/tree.rs`, you can see how it's being used. It's replacing a basic string error.

There's a bunch of existing variants of `OxenError` variants that internally wrap a string. The difference between those and `OxenError::Basic` is the existence of a _specific_ variant _for_ the _specific_ error condition.

crates/lib/src/storage/version_store.rs
malcolmgreaves: Whoops! Thanks for catching this -- not sure why this is there. Reverting.

PR #514, opened 2026-05-01
MerkleTransport trait for client <> server

malcolmgreaves:
Adds new traits for defining how `MerkleStore`s can encode Merkle tree
nodes to send and receive between the `oxen-cli` client and `oxen-server`.
The `MerklePacker` trait collects a subset of nodes and allows a writer to
encode them.

The `MerkleUnpacker` trait uses a provided reader to decode Merkle tree
nodes and update the underlying physical store. Provided that they have the
same `Error` type, combining these two traits is a `MerkleTransport`. A
blanket `impl` for any such unified traits is included here too.

Also defines a `TransportableMerkleStore`, which is a `MerkleStore` + a
`MerkleTransport`. A blanket impl. is provided to unify, so long as the
type's implementation's `Error`s are all the same.


malcolmgreaves:
NOTE: Stacked PR! Must merge https://github.com/Oxen-AI/Oxen/pull/513 before merging.

CleanCut (approved):

PR #515, opened 2026-05-01
Implement Merkle{Packer,Unpacker} for FileBackend

malcolmgreaves:
**`MerklePacker` & `MerkleUnpacker`**
Implemented the new `MerkleTransport` traits for packing up
and unpacking Merkle tree nodes as they're sent on the wire
from the oxen client to server. Preserves the existing tar-gz
formats specific to the FileBackend's physical store layout.

Supports the two unique tar-gz paths:
- upload: only captures `{prefix}/{suffix}/{node,children}`
- everything else: `tree/nodes/{prefix}/{suffix}/{node,children}`
The `PackOptions` enum controls this specific behavior.

**Progress Bar in Bytes**
Changes the progress bar for uploads & downloads to use bytes/s
instead of nodes/s. Nodes are not evenly sized, so progress will
look like it stops when handling a large file. This now hooks
into the reader & writer types that the trait uses.

**Errors**
`MerkleDbError` gains transport-side variants for tar-archive parsing
(`CannotReadMerkle`, `UnsupportedTarEntry`, `PathTraversal`,
`InvalidTarStructure`, `InvalidNodeIdHex`, `MissingNodeDir`,
`MissingTreeNodesDir`, `FsTransport`) plus a corresponding `tokio-util`
features bump (`io`, `io-util`) for async I/O bridging.

**AtomicTempFileSync**
A synchronous analouge to the `async`-friendly `AtomicTempFile`. Used
in the `MerkleUnpacker::unpack` implementation so that unpacking works
efficiently on virtual and networked file systems. Access to it is
only through `liboxen::util::fs::atomic_write_from_reader_sync`.

**Testing**
Tests cover round-trips, path-traversal rejection, malformed entries,
empty tarballs, and on-disk byte parity with the existing `compress_*` /
`unpack_async_tar_archive` paths.

### DIFF
--- a/crates/lib/src/core/db/merkle_node/file_backend.rs
+++ b/crates/lib/src/core/db/merkle_node/file_backend.rs
@@ -1,16 +1,28 @@
-use std::path::{Path, PathBuf};
+use std::collections::HashSet;
+use std::io::{Read, Write};
+use std::path::{Component, Path, PathBuf};
 
-use crate::core::db::merkle_node::merkle_node_db::{MerkleDbError, MerkleNodeDB};
+use flate2::Compression;
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
+use tar::Archive;
+
+use crate::constants::{NODES_DIR, OXEN_HIDDEN_DIR, TREE_DIR};
+use crate::core::db::merkle_node::merkle_node_db::{MerkleDbError, MerkleNodeDB, node_db_path};
 use crate::error::OxenError;
 use crate::model::merkle_tree::merkle_reader::{MerkleEntry, MerkleReader};
+use crate::model::merkle_tree::merkle_transport::{
+    MerklePacker, MerkleUnpacker, PackOptions, UnpackOptions,
+};
 use crate::model::merkle_tree::merkle_writer::{
     MerkleWriteSession, MerkleWriter, NodeWriteSession,
 };
 use crate::model::merkle_tree::node::MerkleTreeNode;
 use crate::model::{LocalRepository, MerkleHash, TMerkleTreeNode};
+use crate::util;
 
-/// File-based Merkle node store backend. Implements the [`MerkleReader`] and
-/// [`MerkleWriter`] traits.
+/// File-based Merkle node store backend. Implements the [`MerkleReader`],
+/// [`MerkleWriter`], [`MerklePacker`], and [`MerkleUnpacker`] traits.
 ///
 /// Borrows the path to a local repository (via the [`LocalRepository`]'s [`PathBuf`]) so it
 /// can delegate straight to [`MerkleNodeDB`]'s existing repository-based methods without any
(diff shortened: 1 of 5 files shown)


malcolmgreaves:
NOTE: Stacked PR! Must merge https://github.com/Oxen-AI/Oxen/pull/514 before merging.

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -181,12 +198,352 @@ impl NodeWriteSession for FileNodeSession {
...
+        reader: &mut dyn Read,
+        opts: UnpackOptions,
+    ) -> Result<HashSet<MerkleHash>, OxenError> {
+        let overwrite_existing = matches!(opts, UnpackOptions::Overwrite);
+        let oxen_hidden = self.repo_path.join(OXEN_HIDDEN_DIR);
+        if self.is_vfs {
+            let tmp = TempDir::new()?;
+            let hashes = extract_tar_under(reader, tmp.path(), overwrite_existing)
+                .map_err(OxenError::from)?;
+            util::fs::copy_dir_all(tmp.path(), &oxen_hidden)?;
```
CleanCut: ⚠️ The default temp dir is sometimes on a different filesystem (like our production server). This can end up forcing a real copy and 2 x peak storage usage.

⚠️ If you ensure that the temp dir is on the filesystem (for example, having it under the dir root you want the final files on would work) then you can use a file move/rename operation to move each file into place and avoid a copy.

CleanCut (approved):

crates/lib/src/core/db/merkle_node/file_backend.rs
```diff
@@ -181,12 +198,352 @@ impl NodeWriteSession for FileNodeSession {
...
+    let mut archive = Archive::new(decoder);
+    let entries = archive.entries().map_err(MerkleDbError::CannotReadMerkle)?;
+
+    let tree_nodes_prefix = Path::new(TREE_DIR).join(NODES_DIR);
+
+    for entry in entries {
+        let Ok(mut file) = entry else {
+            log::error!("Could not unpack file in merkle tar archive");
+            // TODO: raise this error to the caller instead!?
+            continue;
```
CleanCut, line 362: ⚠️ Yeah, your hunch in the comment is spot on — returning the error to the caller feels like the right call here, since silent failures can be a real pain to track down.

crates/lib/src/core/db/merkle_node/file_backend.rs
malcolmgreaves: Using `AtomicTempFile` right? Thanks for the suggestion, let me look at using that here.

CleanCut: No problem! Quick clarification: `AtomicTempFile` (from #545) is single-file, sibling tempfile + atomic rename over one target path. What you'd want here is more like a `TempDir` rooted on the same filesystem as `.oxen/` (e.g. directly inside it) so the unpacked files can be moved into place with a rename instead of a deep copy. Same underlying idea, just at directory granularity. Maybe mimic `AtomicTempFile`'s design, but with a directory?

malcolmgreaves: I had to implement a synchronous analogue so it can work in the trait. I'm now just using it in all cases so there's no more need for `is_vfs` in the `FileBackend`'s fields 👏

crates/lib/src/core/db/merkle_node/file_backend.rs
malcolmgreaves: IIRC I put this comment in because this was something odd I spotted while porting over the refactor. My plan here is to merge this refactor in without changing behavior, then issue a quick follow up to make this `return Err` here. I think erroring out is the right thing to do, but I want to make sure I avoid any sort of external behavior changes in the refactor.

crates/lib/src/core/db/merkle_node/file_backend.rs
CleanCut: Sounds good. 👍🏻 

I've been changing my thoughts about async vs sync anyway, after looking into the tokio::fs implementation (it's...not great).

malcolmgreaves: I can't recall exactly, but isn't there an issue with every OS not necessarily supporting a true async sys call for reading and writing files? There's `io_uring` in Linux but no equivalent for mac os, right?

CleanCut: Yes, it's pretty bad. Apparently only Windows has an actual async interface _that tokio uses_. The `io_uring` interface in Linux is still experimental and disabled by default in `tokio`.

So, every time we do an "async" filesystem call through tokio, it delegates it out to a blocking thread pool. So, all the overhead of async, plus the overhead of switching context to a new thread, then the sync operation, and then the overhead getting back to the other thread again. 🤦🏻‍♂️

PR #516, opened 2026-05-01
Refactor to use Merkle{Packer,Unpacker} traits throughout

malcolmgreaves:
Refactors oxen's Merkle tree node transport between clients
and servers to use the new `MerklePacker` and `MerkleUnpacker`
traits. The `LocalRepository::merkle_store()` method has been
updated to return a `impl TransportableMerkleStore + '_`. And
the private `merkle_store_dispatch::StoreEnum` has been updated
to include dispatch to the `MerkleTransport`'s methods.

Updated codepaths:
- `repositories::tree::compress_tree`, `compress_nodes`, `compress_node`,
  `compress_commits`, and `unpack_nodes` route through the trait, deleting
  the inline tar/gz/file-system blocks.
- `api::client::tree::create_nodes` packs via `merkle_store().pack_nodes`
  with `PackOptions::LegacyClientPush`; `node_download_request` unpacks via
  `merkle_store().unpack` with `UnpackOptions::Overwrite`. The async ↔ sync
  boundary is bridged with `SyncIoBridge` inside `spawn_blocking`.
- `push_progress` and `sync_progress` add a `pack_bytes` length field, used
  to extend the bar before the pack/upload kicks off so the bar has a known
  end and a meaningful ETA.
- `file_backend.rs` test sites switch from `FileBackend::new(&repo)` to
  `repo.merkle_store()`, exercising the trait.

**MerklePacker::raw_byte_count**
Updates the packing trait to include `raw_byte_count`, which is used
as an upper bound on the total number of bytes that will be transfered if
`pack_nodes` is called on a set of Merkle hashes. This is used by user-facing
progress bars in `api::client::tree::create_nodes`: the number of raw bytes
written (_pre compression_) is incremented in the progress bar on each write.

### DIFF
--- a/crates/lib/src/api/client/tree.rs
+++ b/crates/lib/src/api/client/tree.rs
@@ -1,19 +1,17 @@
-use async_compression::futures::bufread::GzipDecoder;
-use async_tar::Archive;
-use flate2::Compression;
-use flate2::write::GzEncoder;
 use futures_util::TryStreamExt;
 use std::collections::HashSet;
+use std::io::Write;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time;
-use tempfile::TempDir;
+use tokio_util::io::{ReaderStream, StreamReader, SyncIoBridge};
 
+use crate::api;
 use crate::api::client;
-use crate::constants::{NODES_DIR, OXEN_HIDDEN_DIR, TREE_DIR};
 use crate::core::progress::push_progress::PushProgress;
 use crate::core::v_latest::index::CommitMerkleTree;
 use crate::error::OxenError;
+use crate::model::merkle_tree::merkle_transport::{PackOptions, UnpackOptions};
 use crate::model::merkle_tree::node::MerkleTreeNode;
 use crate::model::{LocalRepository, MerkleHash, RemoteRepository};
 use crate::opts::download_tree_opts::DownloadTreeOpts;
@@ -21,7 +19,6 @@ use crate::opts::fetch_opts::FetchOpts;
 use crate::view::tree::MerkleHashResponse;
 use crate::view::tree::merkle_hashes::MerkleHashes;
 use crate::view::{MerkleHashesResponse, StatusMessage};
-use crate::{api, util};
 
 /// Check if a node exists in the remote repository merkle tree by hash
 #[tracing::instrument(skip_all)]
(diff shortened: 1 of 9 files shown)


malcolmgreaves:
NOTE: Stacked PR! Must merge https://github.com/Oxen-AI/Oxen/pull/515 before merging.

crates/lib/src/api/client/tree.rs
```diff
@@ -50,61 +47,72 @@ pub async fn has_node(
...
-    log::debug!("create_nodes finished tar");
-    let buffer: Vec<u8> = tar.into_inner()?.finish()?;
+    let n = nodes.len();
+    progress.set_message(format!("Pushing {n} nodes"));
+
+    // Extend the progress bar's total length by an uncompressed-bytes estimate of the
+    // tarball so the upload phase has a known end and a meaningful ETA. Random-ish
+    // merkle hash bytes compress to ~1.0×, so the uncompressed estimate is a tight
+    // upper bound on the bytes that will actually flow over the wire.
+    let estimated_upload_bytes = file_backend::pack_nodes_byte_estimate(local_repo, &nodes);
```
CleanCut: Looks like this slipped past the interfaces and is calling the file backend directly. Let's route it through one of the traits.

crates/lib/src/api/client/tree.rs
```diff
@@ -50,61 +47,72 @@ pub async fn has_node(
...
+        repo.merkle_store()
+            .pack_nodes(&nodes, PackOptions::LegacyClientPush, &mut sync_writer)?;
+        Ok(())
+    });
+
+    // Tick `progress` per chunk so the user sees upload progress moving.
+    let progress_for_stream = Arc::clone(progress);
+    let body_stream = ReaderStream::new(async_reader).inspect_ok(move |chunk| {
+        progress_for_stream.add_bytes(chunk.len() as u64);
+    });
```
CleanCut: Nit: the progress bar is going to stop short of 100%, which might look a little confusing to users.

The reason is that "random-ish merkle hash bytes" are only a portion of the payload. If we drill down into the actual node definitions, there's also a bunch of integers and UTF-8 text, both of which compress really well.

A couple of options here:
1. the simple fix would be to add a progress.finish() at the end so it jumps to 100%, or
2. a more complete fix would be to measure progress in terms of just the uncompressed bytes — set the total to the uncompressed total, and increment by uncompressed bytes as they get fed in to be packed.

CleanCut (approved):
Couple of small things worth a look, but nothing that needs to hold this up.

crates/lib/src/api/client/tree.rs
malcolmgreaves: Great catch! 🙌 Thank you for spotting this. It's a bug and an oversight in the `MerklePacker` trait. I'm adding this as a new trait method and moving this implementation there. The LMDB backend will need an update as it can't reuse this.

crates/lib/src/api/client/tree.rs
malcolmgreaves: Great call out. For (2), please check me on this -- this is inverting the counting from updating as we read the tar gz bytes to updating as we write them. So we'd need to wrap the `Write` impl. so we could read the bytes right at write time, correct?

malcolmgreaves: I added `fn raw_byte_count` to `MerklePacker` so each backend can provide the on-disk size of the stored Merkle nodes. And I took your advice to change the progress bar to count the uncompressed bytes right as we `write` them. I also added in a `.finish()` to the bar when we done for good measure.

Thank you for this suggestion!

PR #519, opened 2026-05-04
Branch count and default branch in repository API

subygan:


crates/lib/src/view/repository.rs
```diff
@@ -73,6 +75,9 @@ pub struct RepositoryDataTypesView {
     pub data_types: Vec<DataTypeCount>,
     pub min_version: Option<String>,
     pub is_empty: bool,
+    pub branch_count: usize,
```
malcolmgreaves, line 79:
```suggestion
    #[serde(default)]
    pub branch_count: usize,
```

Seeing as there's a `default` for the other field you added, I think this is just a happy little accident. But just to explain the reasoning behind this requested change -- 

One thing to keep in mind is that we have to have some sort of upgrade path for old clients interacting with new versions of the server.

If we introduce a new field, an old client won't send it. So we need some way to deserialize it.

Is `0` an ok value to default to here if it's not specified? Or would `None` be safer?

crates/server/src/controllers/repositories.rs
```diff
@@ -86,23 +88,37 @@ pub async fn show(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHttpE
     let repository = get_repo(&app_data.path, &namespace, &name)?;
     let mut size: u64 = 0;
     let mut data_types: Vec<DataTypeCount> = vec![];
+    let mut default_resource: Option<ParsedResourceView> = None;
```
malcolmgreaves, line 91:
```suggestion
```
I don't believe this needs to be mutable. It's only ever being assigned once here as far as I can tell. Is there something I'm missing?

crates/server/src/controllers/repositories.rs
```diff
@@ -86,23 +88,37 @@ pub async fn show(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHttpE
...
+                .iter()
+                .map(|(data_type, count)| DataTypeCount {
+                    data_type: data_type.to_string(),
+                    count: *count as usize,
+                })
+                .collect();
+        }
+
+        let branch = repositories::branches::get_by_name(&repository, DEFAULT_BRANCH_NAME).ok();
+        default_resource = Some(ParsedResourceView::from(ParsedResource {
```
malcolmgreaves, line 110:
```suggestion
        let default_resource = Some(ParsedResourceView::from(ParsedResource {
```

crates/server/src/controllers/repositories.rs
```diff
@@ -86,23 +88,37 @@ pub async fn show(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHttpE
...
+            size = dir_node.num_bytes();
+            data_types = dir_node
+                .data_type_counts()
+                .iter()
+                .map(|(data_type, count)| DataTypeCount {
+                    data_type: data_type.to_string(),
+                    count: *count as usize,
+                })
+                .collect();
+        }
```
malcolmgreaves, line 107: (**Not review blocking**) This is something pre-existing, so no need to take care of this here. But it'd be nice to get rid of all unnecessary `mut` in the codebase. So, if you can, if you're editing something that does this, I'd be great to refactor it into an assignment with expressions.

For instance, here there's no need for `size` and `data_types` to be mutable because they are only ever assigned once:

```suggestion
  let (size, data_types) = {
        // If we have a commit on the main branch, we can get the size and data types from the commit
        if let Ok(Some(commit)) = repositories::revisions::get(&repository, DEFAULT_BRANCH_NAME)
            && let Some(dir_node) =
                repositories::entries::get_directory(&repository, &commit, PathBuf::from(""))?
        {
            (
                dir_node.num_bytes(),
                dir_node
                    .data_type_counts()
                    .iter()
                    .map(|(data_type, count)| DataTypeCount {
                        data_type: data_type.to_string(),
                        count: *count as usize,
                    })
                    .collect(),
            )
        } else {
            (0, vec![])
        }
    };
```

malcolmgreaves (commented):
Nice addition! One blocker is providing a `#[serde(default)]` for the new field. A few other suggestions too.

crates/lib/src/view/repository.rs
jcelliott: Wouldn't this only get sent from the server? Haven't traced it, but I'm guessing we don't send this from the client. If that's true it would only be old clients _receiving_ it, in which case they should just ignore it, correct?

subygan: yup. it'll only be old clients that are receiving the new fields. and can he safely ignored.

malcolmgreaves: (**Only review blocking comment**)

crates/server/src/controllers/repositories.rs
subygan: adding the `let` makes `default_resource` shadow the one in the outerscope and gets collected.

There's a way to not have too many mut, but requires a restructure of the whole function.
I'll get to this in a different PR

crates/server/src/controllers/repositories.rs
subygan: same as previous comment

malcolmgreaves (approved):

PR #523, opened 2026-05-04
Refactor MerkleNodeDb to use repo's Path directly

malcolmgreaves:
The `MerkleNodeDb` only uses the `LocalRepository`'s path. All functions
have changed to borrow the path instead of an entire `LocalRepository` instance.

Additionally, since this type is only used in `liboxen` and it is not an
explicit public API of oxen, it and its associated functions have been
changed to the `pub(crate)` visibility level. This identified some dead
code, which, along with some commented-out code, has been removed.


CleanCut (approved):
❤️ Excellent!

PR #526, opened 2026-05-06
Add Merkle storage enum to RepositoryConfig + merkle_store field

malcolmgreaves:
Adds a new `MerkleStoreKind` enum to the `RepositoryConfig`, which dictates
the type of physical storage that the `LocalRepository` uses for storing
and retrieving Merkle tree nodes. Defaults to the existing custom file format
(aka `FileBackend`) so it is backwards compatible with all existing repositories.

Makes the `TransportableMerkleStore` a field of the `LocalRepository` so that
this is only loaded once at construction time and reused. Also added the
`MerkleStoreKind` uses at load time as another field. Since there's
`Serialization` & `Deserialization` derives for `LocalRepository`, the field
has to be `Option` to provide a default value since it cannot be serialized.
This leads to a brittle design where a `LocalRepository` can be made without
being correctly initialized, hence the new `MerkleStoreNotInitialized` error
variant for `OxenError`. A follow up refactor is necessary to remove these
unnecessary serializaton derives and migrate the HTTP responses to only use
`*View` structs (i.e. `WorkspaceView`, ParsedResourceView`, etc.)

As a result of this initialization problem, `LocalRepository::merkle_store`
has been changed to return a `Result.` It returns a `&dyn MerkleStore` as
the borrow on the `LocalRepository` means we can return a borrow on the
inner state without needing to increment the refcount.

Additionally, uses of the `MerkleTransport` trait from the `LocalRepository`
now go through a new method: `merkle_transport()`. This borrows the internal
`merkle_store` `Arc` and coerces the fat pointer to a `&dyn MerkleTransport`.
The original `merkle_store()` method has been changed to do the same thing,
returning a `&dyn MerkleStore`.

Finally, removes the `set_vfs` method. Merkle store backends need a consistent
`vfs` value and allowing mutation to make this go out of sync. Any places
that needed to use `set_vfs` have been updated to use a modified constructor
that sets the right `vfs` value.

### DIFF
--- a/crates/lib/src/config/repository_config.rs
+++ b/crates/lib/src/config/repository_config.rs
@@ -1,11 +1,30 @@
 use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
+use utoipa::ToSchema;
 
 use crate::constants::DEFAULT_VNODE_SIZE;
 use crate::error::OxenError;
 use crate::model::{LocalRepository, Remote};
 use crate::storage::StorageConfig;
 use crate::util;
+
+/// A sort of registry for known [`MerkleStore`] implementations that can be used by [`LocalRepository`].
+/// This enum serves as a configuration option in a repository's `config.toml`
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, ToSchema)]
+// TODO: remove Serialize + Deserialize derives. These are only necessary because `LocalRepository`
+//       requires them. Those bounds will eventually be dropped.
+#[serde(rename_all = "lowercase")]
+pub enum MerkleStoreKind {
+    /// The [`FileBackend`] store.
+    File,
+}
+
+/// The default is the original custom file format based Merkle tree node storage.
+impl Default for MerkleStoreKind {
+    fn default() -> Self {
+        Self::File
+    }
+}
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct RepositoryConfig {
@@ -28,6 +47,9 @@ pub struct RepositoryConfig {
     /// Currently used only for remote mode
     pub workspace_name: Option<String>,
     pub workspaces: Option<Vec<String>>,
+    /// The type of Merkle tree node storage that the repository uses.
+    #[serde(default)]
+    pub merkle_store_kind: MerkleStoreKind,
 }
 
 impl Default for RepositoryConfig {
@@ -44,6 +66,7 @@ impl Default for RepositoryConfig {
             remote_mode: None,
             workspace_name: None,
             workspaces: None,
+            merkle_store_kind: MerkleStoreKind::default(),
         }
     }
 }
--- a/crates/lib/src/model/repository/local_repository.rs
+++ b/crates/lib/src/model/repository/local_repository.rs
@@ -1,11 +1,12 @@
 use crate::config::RepositoryConfig;
+use crate::config::repository_config::MerkleStoreKind;
 use crate::constants::SHALLOW_FLAG;
 use crate::constants::{self, DEFAULT_VNODE_SIZE, MIN_OXEN_VERSION};
 use crate::core::db::merkle_node::file_backend::FileBackend;
 use crate::core::versions::MinOxenVersion;
 use crate::error::OxenError;
-use crate::model::merkle_tree::TransportableMerkleStore;
 use crate::model::merkle_tree::node::FileNode;
+use crate::model::merkle_tree::{MerkleStore, MerkleTransport, TransportableMerkleStore};
 use crate::model::{MetadataEntry, Remote, RemoteRepository};
 use crate::storage::{NoopVersionStore, StorageConfig, VersionStore, create_version_store};
 use crate::util;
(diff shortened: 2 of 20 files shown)


malcolmgreaves:
NOTE: Stacked PR. Must merge https://github.com/Oxen-AI/Oxen/pull/516 before merging.

crates/lib/src/api/client/internal_types.rs
```diff
@@ -4,6 +4,6 @@ use crate::model::LocalRepository;
 
 #[derive(Debug, Clone)]
 pub(crate) enum LocalOrBase {
-    Local(LocalRepository),
+    Local(Box<LocalRepository>),
```
malcolmgreaves, line 7: Started to get a little large -- `clippy` warned:
```
warning: large size difference between variants
 --> crates/lib/src/api/client/internal_types.rs:6:1
  |
6 | / pub(crate) enum LocalOrBase {
7 | |     Local(LocalRepository),
  | |     ---------------------- the largest variant contains at least 232 bytes
8 | |     Base(PathBuf),
  | |     ------------- the second-largest variant contains at least 24 bytes
9 | | }
  | |_^ the entire enum is at least 232 bytes
````

File: crates/lib/src/model/repository/local_repository.rs [hunk 2]
```diff
@@ -92,21 +102,36 @@ impl LocalRepository {
...
+                initialized_config = RepositoryConfig::from_file(config_path)?;
+                &initialized_config
+            }
+        };
+
+        let store = match config.merkle_store_kind {
+            MerkleStoreKind::File => Box::new(FileBackend {
+                repo_path,
+                is_vfs: config.vfs.unwrap_or(false),
+            }),
```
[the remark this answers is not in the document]
malcolmgreaves: Fixed by removing `LocalRepository::set_vfs`.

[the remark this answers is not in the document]
malcolmgreaves: -1000 for a mutating state design when it should be a parse config then instantiate driven design.

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -179,10 +222,12 @@ impl LocalRepository {
         path: impl AsRef<Path>,
         min_version: impl AsRef<str>,
         storage_config: Option<StorageConfig>,
+        is_vfs: bool,
     ) -> Result<LocalRepository, OxenError> {
         let path = path.as_ref().to_path_buf();
         let storage_config = storage_config.unwrap_or_default();
         let version_store = create_version_store(&path, &storage_config)?;
+        let (m_store, merkle_store_kind) = Self::load_merkle_store(path.clone(), None)?;
```
CleanCut: `is_vfs` isn't getting plumbed through to `load_merkle_store`, but I'm not sure it matters?

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -218,13 +266,20 @@ impl LocalRepository {
...
-    pub fn from_remote(repo: RemoteRepository, path: &Path) -> Result<LocalRepository, OxenError> {
+    pub fn from_remote(
+        repo: RemoteRepository,
+        path: &Path,
+        is_vfs: bool,
+    ) -> Result<LocalRepository, OxenError> {
         let path = path.to_owned();
         let storage_config = StorageConfig::default();
         let version_store = create_version_store(&path, &storage_config)?;
+        let (m_store, merkle_store_kind) = Self::load_merkle_store(path.clone(), None)?;
```
CleanCut: `is_vfs` isn't getting plumbed through to `load_merkle_store`, but I'm not sure it matters?

crates/lib/src/config/repository_config.rs
```diff
@@ -1,12 +1,29 @@
...
 use crate::util;
 
+/// A sort of registry for known [`MerkleStore`] implementations that can be used by [`LocalRepository`].
+/// This enum serves as a configuration option in a repository's `config.toml`
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, ToSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum MerkleStoreKind {
+    /// The [`FileBackend`] store.
+    File,
+}
```
CleanCut, line 20: Quick note: `MerkleStoreKind` and the other serialization extras can probably disappear once I remove the serialization of `LocalRepository`. Maybe worth adding a "todo" so it doesn't get forgotten?

CleanCut (approved):
Meta question: could we maybe get rid of the "vfs" feature? There's a lot of logic plumbed through to support it, but a few things make me wonder:

- It can only be set by manually editing the config toml
- All it really does is unpack the merkle tarball to the system `/tmp` and then copy it over, rather than unpacking directly to the destination. It doesn't touch any other write paths, not  even version file tarball unpacking. So it's a little hard to imagine it being load-bearing.
- In #181 where it first showed up, it's framed as a bug fix, though the actual problem and why this would address it aren't really discussed in the PR. Could it have been an AI assumption that slipped through?

**Update:** I found a [thread with a customer](https://oxenai.slack.com/archives/C098C3W2403/p1763768102647999?thread_ts=1763114989.668779&cid=C098C3W2403) where setting this flag via `oxen clone --vfs` is claimed to have fixed a problem. But the customer also reported immediately running into another problem. 🤷🏻‍♂️ Sounds like we'd need to do some testing with NFS to really figure out if this is necessary or not. If it _is_, then we should probably aute-detect (or auto-fallback) to this behavior. If it isn't, it'd be nice to remove the logic. In any case, it's all stuff to figure out in the future.

malcolmgreaves:
Good question on VFS. It needs to be plumbed through for the LMDB backend because LMDB doesn't support VFS. (So if it's set and a user tries to use that backend, it will error out at creation time).

I think the whole feature should be reexamined and we should potentially redesign how we support and work with VFSes. The `vfs` config option's history is a bit organic and it's probably time to take another look from first principles on what problems we're trying to solve with it.

PR #531, opened 2026-05-07
Reorder tree traversal during writing

malcolmgreaves:
Reorders the tree traversal when writing a new Merkle tree during
a commit from depth-first to a layerwise depth-first approach. Before,
`r_create_dir_node` would start writing the `node` & `children` files
for a directory and keep open file handles while it descended into each
child. It would only close file handles once it recursed up the call stack.
Since each recursive descent created a _new_ node write session (for the
node that was being recursed on), recursion meant that the current stack
frame's session was not being modified by a recursive call. Any modification
to `node` or `children` files was occuring only within a single stack frame.

In the new approach, a directory's vnodes are staged in a `Vec` in the same
stack frame. They're directly written into the `children` file (and their
offsets stored in the `node` file) in one iteration. Any files in the directory
(via a vnode) are included in the `node` and `children` files in this same pass.
The dir node's write session is finished and the collected vnodes are recursed
one at a time.

One performance benefit of this change is that there is only a constant number
of open file handles for the `FileBackend`: a `node` and `children` file per
call to `r_create_dir_node`. The file handles are no longer persisted through
recursive calls, meaning that the number of open file handles no longer grows
linearly with respect to the Merkle tree depth.

Regression tests have been added to show that the `node` and `children` file
contents are byte identicial to before this change.

### DIFF
--- a/crates/lib/src/repositories/commits/commit_writer.rs
+++ b/crates/lib/src/repositories/commits/commit_writer.rs
@@ -795,19 +795,19 @@ fn write_commit_entries(
         root_path.to_str().unwrap(),
         &dir_node.hash().to_string(),
     )?;
-    let mut dir_ns = session.create_node(&dir_node, Some(commit_id))?;
+
     r_create_dir_node(
         repo,
         session,
         commit_id,
-        Some(&mut *dir_ns),
+        commit_id,
+        &dir_node,
+        &root_path,
         dir_hash_db,
         dir_hashes,
         entries,
-        root_path,
         &mut total_written,
     )?;
-    dir_ns.finish()?;
 
     // The dir_hash_db was pre-populated from the previous commit, so
     // removed directories still have stale entries that must be deleted;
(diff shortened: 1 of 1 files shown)


malcolmgreaves:
NOTE: Stacked PR!  https://github.com/Oxen-AI/Oxen/pull/526 must be merged beforehand.

crates/lib/src/repositories/commits/commit_writer.rs
```diff
@@ -839,77 +839,89 @@ fn cache_invalidate_dir_hash_db<'a>(
...
-                            dir_hashes,
-                            entries,
-                            &dir_path,
-                            total_written,
-                        )?;
-                        child_ns.finish()?;
-
-                        dir_node
+                        subdirs_to_recurse.push((n.clone(), sub_path.clone(), v.id));
+                        n
```
CleanCut: Nit: We're cloning every directory node just to return it and call `.hash()` to get the hash. Instead, return the hash everywhere.

```suggestion
                        let hash = n.hash();
                        subdirs_to_recurse.push((n, sub_path.clone(), v.id));
                        hash
```

CleanCut (approved):
> The file handles are no longer persisted through
recursive calls, meaning that the number of open file handles no longer grows
linearly with respect to the Merkle tree depth.

Cool! Maybe we can remove the `ulimit` hack from `bin/test-rust`, then!

PR #533, opened 2026-05-07
Display drift in shared-subtree files after pull/merge

CleanCut:
The merge walker used to short-circuit on dir/vnode subtrees whose hash matched between base and merge target. Files inside those subtrees were never visited, so on-disk drift (e.g., from local changes or an earlier interrupted operation that left files at the wrong content) silently survived every pull/merge. `oxen pull` or `oxen merge` would succeed and the user would only notice via `oxen status` afterwards.

This PR updates the code to track this case and emit output to the user so they are aware of it. Basically a "Hey, we did your pull/merge, and these locally-changed files you have were not affected."

Closes ENG-999


CleanCut:
This is a "nice-to-have", but it adds effort to pull/merge operations that scales `O(n)` with the number of files in the working directory, which is quite significant for repos with, e.g., a million files. We can rely on `oxen status` or `oxen restore` instead to surface this information.

PR #534, opened 2026-05-07
Recreate tracked empty directories in `oxen restore <dir>`

CleanCut:
Oxen tracks directories as first-class citizens, so an empty directory created via `oxen add <empty-dir>` (or left behind by `oxen rm` of the only file in a directory) is a real entry in the merkle tree. Once such a directory is missing on disk (e.g. a user deletes the empty directory), `oxen status` correctly reports it as `removed`, but `restore_dir` only iterated file nodes, so missing empty directories were being ignored by `oxen status` and the user was stuck with a permanently dirty status and no recovery path.

`dir_entries_with_paths` now returns both file and directory entries. `restore_dir` creates every directory in the subtree before the file loop, so empty directories at any depth are materialized on disk. `mkdir` failures accumulate into the same per-path failures vector as file-restore failures.

I also improved the `restore_dir` progress bar: the message used `{path:?}` which printed quoted paths and rendered as `Restoring Directory: ""` for `oxen restore .`. It now reads `Restoring .` or `Restoring some/path`, and the bar's total counts both directories and files so a purely-empty directory tree no longer shows `0/0`.

I added a regression test to make sure this stays fixed.

I also added a section in `.claude/CLAUDE.md` noting Oxen's first-class-directory model so future sessions don't apply Git intuition here.

Closes ENG-1003

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -6,6 +6,16 @@ This file provides guidance to AI agents when working with code in this reposito
 
 Oxen is a fast, unstructured data version control system written in Rust. It's designed to version large machine learning datasets efficiently and provides both a CLI tool and server implementation.
 
+# How Oxen Differs from Git
+
+Oxen's user-facing model is close to Git's (commits, branches, push, pull, status, restore), but a few non-obvious differences trip up Git intuition. Don't apply Git assumptions blindly. Expand this section when you discover a new difference.
+
+- **Empty directories are first-class, tracked content.** Git ignores empty directories; Oxen treats each directory as an entry in the merkle tree on its own. Consequences:
+  + `oxen add <empty-dir>` followed by commit puts the empty dir in the tree.
+  + `oxen rm <only-file-in-dir>` followed by commit leaves the now-empty parent directory tracked as an intentionally-empty entry. By design, not a bug.
+  + If a tracked directory is missing on disk, `oxen status` correctly reports it as `removed`, and `oxen restore <dir>` should recreate it on disk (mkdir, even if it contains no files).
+  + Files placed in a tracked-but-empty directory are still untracked relative to HEAD; `oxen clean -f` will delete them. That is correct behavior.
+
 # Project Organization
 
 The Cargo workspace lives at the repository root, with crates under `crates/`:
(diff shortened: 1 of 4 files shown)


jcelliott (approved):

.claude/CLAUDE.md
```diff
@@ -6,6 +6,16 @@ This file provides guidance to AI agents when working with code in this reposito
...
+# How Oxen Differs from Git
+
+Oxen's user-facing model is close to Git's (commits, branches, push, pull, status, restore), but a few non-obvious differences trip up Git intuition. Don't apply Git assumptions blindly. Expand this section when you discover a new difference.
+
+- **Empty directories are first-class, tracked content.** Git ignores empty directories; Oxen treats each directory as an entry in the merkle tree on its own. Consequences:
+  + `oxen add <empty-dir>` followed by commit puts the empty dir in the tree.
+  + `oxen rm <only-file-in-dir>` followed by commit leaves the now-empty parent directory tracked as an intentionally-empty entry. By design, not a bug.
+  + If a tracked directory is missing on disk, `oxen status` correctly reports it as `removed`, and `oxen restore <dir>` should recreate it on disk (mkdir, even if it contains no files).
+  + Files placed in a tracked-but-empty directory are still untracked relative to HEAD; `oxen clean -f` will delete them. That is correct behavior.
+
```
malcolmgreaves, line 18: This kind of implementation information shouldn't be in a CLAUDE.md file @CleanCut . It's better for the LLM to read the source code to understand actual behavior. And it's unlikely that someone will remember to read through this file and change it when the implementation changes.

From [Anthropic's best practices guide for CLAUDE.md files](https://code.claude.com/docs/en/best-practices):

> CLAUDE.md is a special file that Claude reads at the start of every conversation. Include Bash commands, code style, and workflow rules. This gives Claude persistent context it can’t infer from code alone.

> ✅ Include | ❌ Exclude
> -- | --
> Bash commands Claude can’t guess | Anything Claude can figure out by reading code

The more this file grows, the more it starts to dominate the user-specified prompt. There's nothing special about this file per-se. It's not a "memory" or "special instruction" for claude. It it is just the 2nd thing that claude adds to its prompt. The full prompt is:
1. claude system prompt (_we don't know this, it's secret and proprietary_)
2. this CLAUDE.md file
3. what you write in your first message to claude

At best, claude will disregard this kind of thing when it reads the source code. That will still take up some part of its context and require the attention heads to apply near-zero weight to these tokens. That later part will be harder to do, so it will end up probably confusing the LLM if there's any sort of divergence.

malcolmgreaves: Cleaning this up here: https://github.com/Oxen-AI/Oxen/pull/539

PR #537, opened 2026-05-07
LMDB based `MerkleStore` implementation

malcolmgreaves:
Adds the `heed` crate to provide access LMDB (Lightning Memory-Mapped
Database). Creates a new `MerkleStore` implementation using LMDB as
`LmdbBackend` under the new `core::db::merkle_node::lmdb` package in
`liboxen`. Extensive new tests have been added to ensure that the
memory layout of LMDB values is consistent and that LMDB operations
work as expected.

**LMDB Store Design**
The `LmdbBackend` uses two tables to store all Merkle tree nodes:
1. `merkle_tree_nodes`: u128 -> ~EMerkleTreeNode
2. `merkle_links`: u128 -> parent(u128) + []children(u128)

(1) stores the actual Merkle tree node struct. It has the type and
the msgpack serialized bytes for the `EMerkleTreeNode`. To maintain
backwards compatability, the `EMerkleTreeNode`'s serialized representation
is used as-is and _not_ modified. (Modification would require a migration).

(2) stores the connections that dictate the structure of the Merkle tree.
Each node maps to a `LmdbLink`, which is an optional parent connection
and a list of the node's children. Each of these is are `MerkleHash`es:
they're stored as 16 byte `u128` values.

**`zerocopy` uses**
The `zerocopy` dependency has been added as the `LmdbBackend` offers full
zero-copy support for read operations. These are implemented using methods
on the `LmdbBackend` struct itself. The `MerkleReader` operations require
owned data, so these views have to be copied to comply with the trait design.
However, this opens the door in the future to iterating on the trait to
return borrows on the underlying data.

Each internal table has its own zerocopy view: `LmdbNodeRef` for `LmdbNode`
and `LmdbLinkRef` for `LmdbLink`. The borrows last as long as the lifetime
for the read transaction because they are direct views into LMDB's internal
memory-mapped pages.

**`MerkleReader` implementation**
The `LmdbBackend` actually stores `FileNode` and `FileChunkNode` Merkle tree
nodes in its store directly. This diverges from the `FileBackend`, where,
for better file access patterns and to reduce inode pressure, file nodes are
only stored in the `children` file and require parsing the lookup table from
the `node` file.

To ensure that `LmdbBackend` adheres to the constraints of `MerkleReader`,
the `get_node` and `exists` methods treat file nodes as not being present.

However, the `LmdbBackend` struct provides `full_exists` & `full_get_node`
which work correctly on actually stored file and file chunk nodes.

**`MerkleWriter` implementation**
LMDB encourages the use of short-lived transactions. Writing into LMDB
directly buffers data in memory (via memory-mapped pages). Closing a transaction
requires an fsync, which is an expensive syscall. The writer implementation
explicitly buffers written nodes and children via a `Cell<Vec<.>>`. The enclosing
write session's `finish` performs the actual write to LMDB. Node that the node
write session _does not_ actually ensure that writes are persisted to LMDB,
as this would incur a performance penalty via fsync.

### DIFF
--- a/crates/lib/src/core/db/merkle_node/lmdb/lmdb_backend.rs
+++ b/crates/lib/src/core/db/merkle_node/lmdb/lmdb_backend.rs
@@ -1,0 +1,228 @@
+use std::path::{Path, PathBuf};
+
+use heed::byteorder::LE;
+use heed::types::{Bytes, DecodeIgnore, U128};
+use heed::{AnyTls, Database, Env, EnvOpenOptions, WithTls};
+
+use crate::constants::OXEN_HIDDEN_DIR;
+use crate::core::db::merkle_node::lmdb::LmdbError;
+use crate::core::db::merkle_node::lmdb::value_structs::{LmdbLink, LmdbNode};
+use crate::error::OxenError;
+use crate::model::MerkleHash;
+
+/// Keys are merkle hashes, which are `u128` values.
+/// MUST USE LITTLE ENDIAN (LE) so the byte layout matches
+/// `ContentHash::as_bytes` / `LocationHash::as_bytes`.
+pub(super) type KeyLmdb = U128<LE>;
+
+/// Values are serialized as bytes. Access requires deserialization into structs.
+pub(super) type ValueLmdb = Bytes;
+
+/// LMDB table name: storing merkle tree node byte
+const DB_NODES: &str = "merkle_tree_nodes";
+/// LMDB table name: storing parent & child relationships per-node
+const DB_LINKS: &str = "merkle_links";
+
+/// Merkle tree node storage that is backed by LMDB.
+///
+/// NOTE: DO NOT USE ON A VIRTUAL FILE SYSTEM !!
+pub struct LmdbBackend {
+    /// The filesystem location of the local repository.
+    pub(super) repo_root: PathBuf,
+    /// The LMDB environment that contains the [`Database`] fields.
+    pub(super) lmdb_env: Env<WithTls>, // note: WithTls makes this !Send. Use AnyTls if need to send between threads.
+
+    /// Stores every kind of merkle tree node: any concrete [`MerkleTreeNode`].
+    /// This includes files nodes! Note that there is no other data but the Merkle
+    /// tree node type (u8) and the msgpack-serialized bytes of the actual node.
+    ///
+    /// These values deserialize into a [`LmdbNode`]. The  zerocopy view on these
+    /// values is the [`LmdbNodeRef`].
+    ///
+    /// In the LMDB environment, this has name of the [DB_NODES] constant.
+    pub(super) merkle_tree_nodes: Database<KeyLmdb, ValueLmdb>,
+
+    /// Stores the parent and children connections for each Merkle tree node.
+    ///
+    /// These values deserialize into a [`LmdbLink`]. The zerocopy view on these
+    /// values is the [`LmdbLinkRef`].
+    ///
+    /// In the LMDB environment, this has name of the [DB_LINKS] constant.
+    pub(super) merkle_links: Database<KeyLmdb, ValueLmdb>,
+}
+
+impl LmdbBackend {
+    pub fn new(repo_root: PathBuf, options: EnvOpenOptions) -> Result<Self, OxenError> {
+        let options = {
+            let mut options = options;
+            options.max_dbs(2);
+            log::debug!("Config for LMDB backend: {options:?}");
+            options
+        };
+
+        let lmdb_env = {
+            let db_location = lmdb_dir_location(&repo_root);
+            log::debug!("Opening LMDB backend at: {}", db_location.display());
+            // SAFETY: LMDB uses a memory mapped file. If this file is modified in or out of process,
+            //         it can cause undefined behavior. There is nothing else in the Oxen codebase that
+            //         modifies this mmap'd file. Nothing else access `db_location` except for this code.
+            //         This uses LMDB's own opening code, which does lock the file to check for and
+            //         prevent concurrent access. The fact that it is a "hidden" file helps somewhat: it will
+            //         be unlikely for someone or something to accidentily stumble upon it and mess things up.
+            unsafe { options.open(db_location).map_err(LmdbError::Access)? }
+        };
+
+        // create these two key-value tables
+        let mut wtxn = lmdb_env.write_txn().map_err(LmdbError::Access)?;
+        let merkle_tree_nodes = lmdb_env
+            .create_database::<KeyLmdb, ValueLmdb>(&mut wtxn, Some(DB_NODES))
+            .map_err(LmdbError::Write)?;
+        let merkle_links = lmdb_env
+            .create_database::<KeyLmdb, ValueLmdb>(&mut wtxn, Some(DB_LINKS))
+            .map_err(LmdbError::Write)?;
+        wtxn.commit().map_err(LmdbError::Write)?;
+        log::debug!(
+            "Ensure tables '{DB_NODES}' (Merkle nodes) and '{DB_LINKS}' (tree connections) exist"
+        );
+
+        Ok(Self {
+            repo_root,
+            lmdb_env,
+            merkle_tree_nodes,
+            merkle_links,
+        })
+    }
+
+    /// Private helper for creating a new LMDB write transaction.
+    /// Returns [`LmdbError::Access`] on error.
+    pub(in crate::core::db::merkle_node::lmdb) fn write_txn<'a>(
+        lmdb_env: &'a heed::Env<heed::WithTls>,
+    ) -> Result<heed::RwTxn<'a>, LmdbError> {
+        lmdb_env.write_txn().map_err(LmdbError::Access)
+    }
+
+    /// Private helper for creating a new LMDB read transaction.
+    /// Returns [`LmdbError::Access`] on error.
+    pub(in crate::core::db::merkle_node::lmdb) fn read_txn<'a>(
+        &'a self,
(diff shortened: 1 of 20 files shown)


malcolmgreaves:
NOTE: Stacked PR! https://github.com/Oxen-AI/Oxen/pull/531 must be merged first!

malcolmgreaves:
NOTE: Stacked PR! https://github.com/Oxen-AI/Oxen/pull/531 must be merged first!!

crates/lib/src/core/db/merkle_node/lmdb/writer.rs
```diff
@@ -0,0 +1,497 @@
...
+}
+
+impl NodeWriteSession for LmdbNodeWriteSession {
+    /// Hash of the node currently being written.
+    fn node_id(&self) -> &MerkleHash {
+        &self.node_hash
+    }
+
+    /// Serialize this child and queue for writing.
+    fn add_child(&mut self, child: &dyn TMerkleTreeNode) -> Result<(), OxenError> {
```
CleanCut, line 167: ⚠️ What sets the child's parent?

crates/lib/src/core/db/merkle_node/lmdb/reader.rs
```diff
@@ -0,0 +1,250 @@
...
+        match node_ref.kind()? {
+            File | FileChunk => Ok(false),
+            Commit | Dir | VNode => Ok(true),
+        }
+    }
+
+    /// Retrieves the node with the given `hash` from the store. `None` means no such node exists.
+    ///
+    /// NOTE: to comply with [`MerkleReader::get_node`]'s semantics, this method
+    /// has to consider present file nodes as not existing.
```
CleanCut, line 40: I'm trying to wrap my head around this. What is the reason `get_node` returns `None` even when a file node exists? It's a bit confusing on first read. If the file implementation is the source of this behavior, would it be worth changing it in this PR so it is more intuitive?

crates/lib/src/core/db/merkle_node/lmdb/lmdb_backend.rs
```diff
@@ -0,0 +1,222 @@
...
+const DB_LINKS: &str = "merkle_links";
+
+/// Merkle tree node storage that is backed by LMDB.
+///
+/// NOTE: DO NOT USE ON A VIRTUAL FILE SYSTEM !!
+pub struct LmdbBackend {
+    /// The filesystem location of the local repository.
+    pub(super) repo_root: PathBuf,
+    /// The LMDB environment that contains the [`Database`] fields.
+    pub(super) lmdb_env: Env<WithTls>, // note: WithTls makes this !Send. Use AnyTls if need to send between threads.
```
CleanCut, line 33: Nit: A small thought on the comment — `Env<WithTls>` doesn't seem to make `LmdbBackend` `!Send`, since the `impl MerkleReader for LmdbBackend` below compiles fine even with `pub trait MerkleReader: Send + Sync`. Could it be that you were thinking of the transactions created from `Env<WithTls>`? Those are `!Send`, though I'm not totally sure we need to mention it.

crates/lib/src/core/db/merkle_node/lmdb/lmdb_backend.rs
```diff
@@ -0,0 +1,222 @@
...
+            let mut options = options;
+            options.max_dbs(2);
+            log::debug!("Config for LMDB backend: {options:?}");
+            options
+        };
+
+        let lmdb_env = {
+            let db_location = lmdb_dir_location(&repo_root);
+            log::debug!("Opening LMDB backend at: {}", db_location.display());
+            unsafe { options.open(db_location).map_err(LmdbError::Access)? }
```
CleanCut, line 72: What do you think about adding a [Safety comment](https://std-dev-guide.rust-lang.org/policy/safety-comments.html) wherever we reach for unsafe code? It would help future readers (and us) understand the invariants.

crates/lib/src/core/db/merkle_node/lmdb/writer.rs
```diff
@@ -0,0 +1,497 @@
...
+/// onto the queue when their `finish` is called. This session's [`Self::finish`] opens
+/// a single [`heed::RwTxn`], drains the queue, and commits — so either every queued write
+/// is persisted or none is.
+///
+/// **Callers MUST call `finish()` on every node session and on the outer write session.**
+/// There is intentionally no [`Drop`] guard: any session that goes out of scope without
+/// an explicit `finish()` silently loses its buffered state. The [`super::super::file_backend`]
+/// `FileNodeSession` has a different model (it writes eagerly inside each `add_child` /
+/// `create_node`) so its `Drop` only flushes file handles — LMDB's queue-based model
+/// requires explicit commits, which is the trait's contract.
```
CleanCut: Since calling `finish()` is a hard requirement on this backend, what would you think about making it a requirement of the trait itself, and dropping the `Drop` impl on the file backend that calls `finish()` for you? That way the contract is uniform across implementations.

One thought on how to keep "forgot to call `finish()`" from slipping in: we could add a `bool` field that tracks whether `finish()` has been called, paired with test-only `Drop` impls on both backends (gated by `#[cfg(test)]`) that panic if the flag is still `false`. Then as long as our tests exercise the trait reasonably well, anything that forgets to call `finish()` would surface as a failing test rather than going unnoticed.

CleanCut (approved):
⚠️ Double-check that a child's parent gets set when you call `add_child`

crates/lib/src/core/db/merkle_node/lmdb/reader.rs
malcolmgreaves: It's unintuitive! But it's here to fit how the existing oxen code expects this to work.

First, to address -- 
> would it be worth changing it in this PR so it is more intuitive?

Yes. 100%. What this _should_ do is have `get_node` return you a node full stop. This is why I put in a `get_node_full` method for `LmdbBackend`. When the time comes, we'll just delete and rename. (I also did this to document the "sane" way of doing this :) ).

Now, why is it this way? _I asked myself this **a ton** while working on this._

Because the rest of oxen is somewhat hardocded under the premise that Merkle tree nodes are stored in these `node` and `children` files that live under a directory (the `{hash prefix}/{hash suffix}/` dir).

If there was a `node` file for every actual file that's versioned, we'd have _way_ too many files. Both in terms of managing (We might hit into some OS limits at the large end of repos too!) And in terms of access times.

So the original file backend for the Merkle tree made a design decision: `FileNode`s are actually only ever stored in `children` files. `node` files only store a `DirNode`, a `CommitNode`, or a `VNode`. There's only a `create_node` on these subtypes. Every time we see a file (or `FileChunkNode`) we are writing it in some `add_children` call. Similarly, the code only calls `get_node` on non-files. Files are looked up by calling `get_children` on its directory vnode.

Files are always accessed by doing:
1. get the directory
2. figure out the vnode for the file
2. do `get_node` on this
3. get the offset table (_from the `node` file_)
4. seek to the position of the desired file in the `children` file
5. read & deserialize the file node from this position

What I really think the code _should_ do is just call `get_node`. There shouldn't be any need to call `get_children` to get a file node IMO!

Doing this right now would be a really big ask though. I'd have to audit the entire codebase and make sure I track down all of these calls and change them correctly. That's a bit too much scope creep for this initial large refactor + LMDB implementation 😅 Something that's very much on my radar though!

crates/lib/src/core/db/merkle_node/lmdb/lmdb_backend.rs
malcolmgreaves: > Parameter defining that read transactions are opened with Thread Local Storage (TLS) and cannot be sent between threads `!Send`. It is often faster to open TLS-backed transactions.

https://docs.rs/heed/latest/heed/enum.WithTls.html

crates/lib/src/core/db/merkle_node/lmdb/lmdb_backend.rs
malcolmgreaves: Very happy to have that here! It should be here. Thanks for flagging this!

crates/lib/src/core/db/merkle_node/lmdb/writer.rs
malcolmgreaves: > making it a requirement of the trait itself, and dropping the Drop impl on the file backend that calls finish() for you?

Yup! I agree with this. Let me double check -- maybe this comment is left over and wrong and needs to be changed, or maybe I forgot to remove this. I should have updated `FileBackend`'s implementation earlier to not have a `impl Drop`. The `LmdbBackend` should not have one.

malcolmgreaves: re: checking if `finish()` is called. I think I see what you're suggesting. Though, IIUC, a test that wanted to show a failure and show that things were correctly not finished would make this a harder to check, right? E.g. if we wanted to show that we can back out of an LMDB write session and not commit, we'd have to make sure we didn't panic in the test code in that situation.

malcolmgreaves: Comment was outdated so I updated it!

crates/lib/src/core/db/merkle_node/lmdb/writer.rs
malcolmgreaves: It's set by creating one of these sessions. It's created via `session.create_node(parent) -> NodeWriteSession`. It goes `MerkleWriteSession` -> (many) `NodeWriteSession`s. This `add_child` is on the `NodeWriteSession`. The `finish()` here gets the parent ID out of self as it pushes the `PendingWrites` into the queue.

PR #540, opened 2026-05-08
S3: Rename storage backend identifier to `kind` and rewrite TOML schema (step 13.2)

CleanCut:
This PR simplifies the per-repository config file options for storage.

- `type` in the config (`type_` in the code) was ignored. Renamed to `kind` so that the name could be the same in the code and config. Since it was ignored before, this is completely safe. `kind = s3` will be implemented in a future PR.
- `storage.settings.path` setting was being used before to specify a custom location for the versions storage. Extra care was taken to continue accepting this value, and rewriting it into the new value `storage.versions_path`
- Any write to the config by oxen will write out the new config format. Interactions that prompt writing the config format include moving the repo's location on disk and `oxen config {--set-remote|--remove-remote|--storage-backend}`.

Before:

```toml
[storage]
type = "local"

[storage.settings]
path = "/mnt/nfs/..."
```

After:

```toml
[storage]
kind = "local"
versions_path = "/mnt/nfs/..."
```

Closes ENG-918

### DIFF
--- a/crates/lib/src/config/repository_config.rs
+++ b/crates/lib/src/config/repository_config.rs
@@ -70,3 +70,138 @@ impl RepositoryConfig {
         self.vnode_size.unwrap_or(DEFAULT_VNODE_SIZE)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::PathBuf;
+
+    fn parse(toml_str: &str) -> RepositoryConfig {
+        toml::from_str(toml_str).expect("test fixture must parse")
+    }
+
+    #[test]
+    fn parses_canonical_storage_section() {
+        let toml = r#"
+            remotes = []
+
+            [storage]
+            kind = "local"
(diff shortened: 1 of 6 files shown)


jcelliott:
@CleanCut what's your plan for the shape of the S3-specific settings? I _think_ the original intent of having a specific settings object separate from the type/kind was to allow that shape to be defined by the selected storage backend. I don't have any preference here, just curious what the direction is.

CleanCut:
@jcelliott My idea was that since there's only a single setting for `kind = local`, we'd just have everything in `[storage]`, and the backends would only look at the settings they care about.

We could split into `[storage.local]` for local and `[storage.s3]` for s3, etc. if it gets more complicated. Or if you feel better about that direction we can switch now. Just let me know!

jcelliott:
@CleanCut that makes sense to me. No point adding complexity unless we need it.

jcelliott (approved):
Looks good. Tested locally with a new clone as well as updating a remote and saw the config update. Push and pull still work.

PR #542, opened 2026-05-08
S3: Delete `StorageOpts`, use `StorageConfig` directly (step 13.3)

CleanCut:
We were allowing the client to set and change version files storage config _on the server_ from the client. This removes that ability, other than the ability to select `local` or `s3` at remote repo creation time (and the server's config will then use the appropriate local or s3 settings, or return an error if the kind is not configured).

This also removes the ability to use cli commands to alter local storage config settings, though the ability to manually set them in the config file is still present. My reasoning for this was (1) until now, these settings were mostly vestigial and unused, and (2) changing the config on all but an empty repo requires knowledge of how to do surgery on the backend, which is wildly unsafe. If we allow settings these settings at all in the future, we ought to do it only in `oxen init`. I'm skeptical giving the client the ability to use `s3` for version file storage even makes sense, given we provide "partial repo" functionality on the client.

- **_Most_** of this PR is just deleting stuff.
- The rest of the PR is pretty much just:
  - Use the `StorageConfig` struct instead of the removed/redundant `StorageOpts`
  - I added [a `StorageKind` enum](https://github.com/Oxen-AI/Oxen/pull/542/changes#diff-d9543740dd6fc2962deb38882527e8c531511eb2a21e8ede5cc5bb9a2de64592R77-R106) to use instead of "local" and "s3" strings

<details><summary>More details, if you want a technical deep dive...</summary>
<br>
`StorageOpts` was a holdover that bundled CLI inputs, wire payloads, and runtime parameters into one shape. Now that `StorageConfig` exists (step 13.2), the runtime can consume it directly.

Before:

    { "storage_opts": { "kind": "s3", "s3_opts": { "bucket": "...", "prefix": "..." } } }

After:

    { "storage_kind": "s3" }

`enum StorageKind { Local, S3 }` with `Default = Local`, serde `rename_all = "lowercase"`, plus `Display`, `FromStr`, and `ToSchema` impls. Replaces the stringly-typed `kind: String` on `StorageConfig` and the `Option<String>` on `RepoNew`. The `VersionStore` trait's `storage_type(&self) -> &str` becomes `storage_kind(&self) -> StorageKind`, so the call site in `LocalRepository::save()` matches exhaustively on the enum.

- `crates/lib/src/opts/storage_opts.rs`, `local_storage_opts.rs`, `s3_opts.rs`
- `command::config::set_version_store`
- `CloneOpts.storage_opts` field — cloned working trees use the local default

`--storage-backend-path` and `--storage-backend-bucket` removed from every command. `--storage-backend` removed from `init`, `clone`, and `config` — local working trees are always local; users wanting a custom `versions_path` edit the repo's `.oxen/config.toml` directly. `--storage-backend <kind>` stays on `create-remote` (the client picks among configured server backends; the server may reject if the requested kind isn't enabled). Python `create_repo` / `create` lose the `storage_backend_path` / `storage_backend_bucket` kwargs.

- `OxenError::UnsupportedStorageKind(String)` — surfaces when `StorageKind::FromStr` receives something other than `"local"` or `"s3"`.
- `OxenError::S3BackendNotImplemented` — `create_version_store` returns this for `StorageKind::S3` until step 13.5 wires up the server-side admin config.
</details>

Closes ENG-919


subygan (approved):
super nice Cleanup!

PR #545, opened 2026-05-12
Repo Integrity: atomic file write helpers (step 1.1)

CleanCut:
This PR adds two utility functions to write files atomically to the local filesystem.

Some of our most recent repository corruption (both client- and server-side) has occurred when operations were interrupted (i.e. the client or server is killed) while writing files to local storage on disk. It's extremely difficult (and slow) to constantly try to detect and recover from corrupted files every time we interact with storage-on-disk, so we need to do better in preventing the situation in the first place. These functions should facilitate that by making it so that we have high confidence that if a file exists, it contains the content we intended to write to it. These helpers accomplish that by writing to a temp file, and then renaming it over the target . A crash before the rename leaves the prior contents intact (and a temp file that can be cleaned up later). A crash after the rename leaves the new contents.

```rust
util::fs::atomic_write_to_path(target, contents)   // for known-small files
util::fs::atomic_write_from_reader(target, reader) // for large or streamed files
```

I also migrated `OxenError::file_create_error` and `file_rename_error` off of using `OxenError::basic_str` under the hood. Instead, they use new specific variants:

```rust
FileCreate(PathBuf, #[source] io::Error)
FileRename { src, dst, #[source] source: io::Error }
```


crates/lib/src/util/fs.rs
```diff
@@ -198,6 +199,183 @@ pub fn write_to_path(path: impl AsRef<Path>, value: impl AsRef<str>) -> Result<(
...
+                .map_err(|err| OxenError::file_create_error(parent, err))?;
+        }
+        let temp_dir = parent.unwrap_or_else(|| Path::new("."));
+
+        let target_name = target.file_name().ok_or_else(|| {
+            OxenError::file_create_error(
+                target,
+                std::io::Error::other("target path has no filename component"),
+            )
+        })?;
```
malcolmgreaves: Minor, but you could check and error out before successfully creating the parent dir if it doesn't exist.

crates/lib/src/util/fs.rs
```diff
@@ -198,6 +199,183 @@ pub fn write_to_path(path: impl AsRef<Path>, value: impl AsRef<str>) -> Result<(
...
+
+        tokio::fs::rename(temp_path, &self.target)
+            .await
+            .map_err(|err| OxenError::file_rename_error(temp_path, &self.target, err))?;
+
+        if let Some(parent) = self.target.parent().filter(|p| !p.as_os_str().is_empty())
+            && let Ok(dir) = tokio::fs::File::open(parent).await
+            && let Err(err) = dir.sync_all().await
+        {
+            log::warn!("AtomicTempFile::commit: parent fsync failed for {parent:?}: {err}");
```
malcolmgreaves: If there's an error in the parent's open, I am pretty sure this will not log the warning. When the 2nd clause (`tokio::fs::File::open`) returns `Err` instead of `Ok`, the evaluation will both short-circuit and it treat the whole expression as `false`.

Did you want this to log a warning if `dir.sync_all()` didn't run? It seems like the intention is to warn if the directory fsync doesn't work as expected. Presumably not being able to run it is also a warning?

crates/lib/src/util/fs.rs
```diff
@@ -198,6 +199,183 @@ pub fn write_to_path(path: impl AsRef<Path>, value: impl AsRef<str>) -> Result<(
...
+                 after rename — concurrent recreation? Continuing."
+            ),
+            Err(err) if err.kind() == std::io::ErrorKind::NotFound => {}
+            Err(err) => {
+                log::error!(
+                    "AtomicTempFile::commit: cleanup unlink failed for {temp_path:?}: {err}"
+                );
+                return Err(err.into());
+            }
+        }
```
malcolmgreaves, line 342: Very smart idea! 👏

crates/lib/src/util/fs.rs
```diff
@@ -198,6 +199,183 @@ pub fn write_to_path(path: impl AsRef<Path>, value: impl AsRef<str>) -> Result<(
...
+
+        Ok(())
+    }
+}
+
```
malcolmgreaves, line 347: Both of these functions and the entire `AtomicTempFile` idea are wonderful 💯

malcolmgreaves (approved):
Great design on `AtomicTempFile` and the `pub` parts only exposing the complete write path!

crates/lib/src/util/fs.rs
CleanCut: Nice catch! ✅ Swapped the order.

crates/lib/src/util/fs.rs
CleanCut: Good point! ✅ I refactored the code so we warn in each case.

PR #546, opened 2026-05-12
Option to use LMDB Merkle tree store in repos

malcolmgreaves:
**`oxen init --merkle-store`**
Wires up `oxen init` with a new `--merkle-store` parameter. It defaults to
use `"file"` for the `FileBackend`. But it can accept `"lmdb"`, which will
configure the repository to use the new `LmdbBackend`.

**`MerkleTransport for LmdbBackend`**
Implements `MerkleTransport` for `LmdbBackend` using the same tar archive
wire format that the `FileBackend` client <> server communication uses. A
migration will be required in order to support a more efficient Merkle tree
node transport protocol that's more suited to LMDB.

### DIFF
--- a/crates/lib/src/config/repository_config.rs
+++ b/crates/lib/src/config/repository_config.rs
@@ -1,5 +1,7 @@
 use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
+use strum::{Display, EnumIter, EnumString, IntoStaticStr, VariantArray, VariantNames};
+use thiserror::Error;
 use utoipa::ToSchema;
 
 use crate::constants::DEFAULT_VNODE_SIZE;
@@ -8,15 +10,51 @@ use crate::model::{LocalRepository, Remote};
 use crate::storage::StorageConfig;
 use crate::util;
 
+#[derive(Debug, Error)]
+pub enum RepoConfigError {
+    #[error("[RepositoryConfig] Failed to read config: {0}")]
+    Read(Box<OxenError>),
+
+    #[error("[RepositoryConfig] TOML read error: {0}")]
+    TomlDe(#[from] toml::de::Error),
+
+    #[error("[RepositoryConfig] TOML write error: {0}")]
+    TomlSer(#[from] toml::ser::Error),
+
+    #[error("[RepositoryConfig] Failed to write config: {0}")]
+    Write(Box<OxenError>),
+
+    #[error("[RepositoryConfig] Unsupported Merkle store kind: {kind}. Expected one of {tokens:?}.", kind=.0, tokens=<MerkleStoreKind as VariantNames>::VARIANTS)]
+    UnknownMerkeKind(#[from] strum::ParseError),
+}
+
 /// A sort of registry for known [`MerkleStore`] implementations that can be used by [`LocalRepository`].
 /// This enum serves as a configuration option in a repository's `config.toml`
-#[derive(Serialize, Deserialize, Debug, Clone, Copy, ToSchema)]
+#[derive(
+    Serialize,
+    Deserialize,
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    ToSchema,
+    EnumString,
+    EnumIter,
+    VariantNames,
+    VariantArray,
+    Display,
+    IntoStaticStr,
+)]
 // TODO: remove Serialize + Deserialize derives. These are only necessary because `LocalRepository`
 //       requires them. Those bounds will eventually be dropped.
 #[serde(rename_all = "lowercase")]
+#[strum(serialize_all = "lowercase")] // WARNING!! must mirror serde's `rename_all`
 pub enum MerkleStoreKind {
     /// The [`FileBackend`] store.
     File,
+    /// The [`LmdbBackend`] store.
+    Lmdb,
 }
 
 /// The default is the original custom file format based Merkle tree node storage.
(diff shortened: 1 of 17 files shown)


malcolmgreaves:
NOTE: This is a stacked PR! https://github.com/Oxen-AI/Oxen/pull/537 must be merged first!!

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -114,32 +119,40 @@ impl LocalRepository {
...
+                if is_vfs {
+                    return Err(OxenError::MerkleStoreLmdbNotSupportedOnVfs);
+                }
+                let env_dir = lmdb_dir_location(&repo_path);
+                util::fs::create_dir_all(&env_dir)?;
+                let mut options = heed::EnvOpenOptions::new();
+                // 1 GiB ceiling — large enough for typical Merkle trees,
+                // small enough to keep sparse-file disk usage in check. Can
+                // be revisited if/when we have repos that exceed it.
+                options.map_size(1024 * 1024 * 1024);
```
CleanCut, line 159: What would you think about bumping this up to 1 TiB? Since it's just virtual address space rather than actual disk usage, the sparse file should only consume as much as we've actually written — at least on any modern filesystem, which feels like a safe assumption for our target customers.

For some context on the upper end of scale: per @gschoeni, [Activision's repo has around 15.8 million files](https://oxenai.slack.com/archives/C03EBLD4B7W/p1765479707015199). Claude estimates that a typically-shaped repo fits roughly 5 million files' node records per gigabyte of lmdb storage (give or take, depending mostly on filename length). So bumping this to 1 TiB would give us headroom for somewhere around 5 billion files before anyone needs to revisit this setting.

crates/lib/src/config/repository_config.rs
```diff
@@ -24,6 +26,38 @@ impl Default for MerkleStoreKind {
...
+impl std::str::FromStr for MerkleStoreKind {
+    type Err = OxenError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "file" => Ok(MerkleStoreKind::File),
+            "lmdb" => Ok(MerkleStoreKind::Lmdb),
+            other => Err(OxenError::basic_str(format!(
+                "Unsupported Merkle store kind: {other:?}. Expected one of {:?}.",
+                Self::all_tokens()
```
CleanCut: Adding a new `Basic` error? 😉

crates/lib/src/core/db/merkle_node/lmdb/unpack.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Hashes we observed at any path component. Returned to the caller even
+    /// if the entry's bytes were skipped (mirrors the file backend's reporting
+    /// of every parsed hash from the tarball).
+    seen_hashes: HashSet<MerkleHash>,
+}
+
+/// Read every `node` / `children` entry out of the tar archive into memory,
+/// keyed by hash. Path validation (path-traversal, allowed entry types,
+/// well-formed hash dirs) matches [`super::super::file_backend::FileBackend::unpack`].
+fn collect_tar_entries(reader: &mut dyn Read) -> Result<BufferedTarPayloads, OxenError> {
```
CleanCut, line 72: ⚠️ If I'm reading this right, the whole merkle tarball gets untarred into memory in one go — which could end up being pretty large at our scale. What would you think about streaming the entries straight to the backing store, the way the client code does?

crates/lib/src/test.rs
```diff
@@ -303,6 +303,80 @@ where
...
+    )?;
+    let version_store = repo.version_store();
+    version_store.init().await?;
+
+    log::info!(">>>>> run_empty_local_repo_test_with_merkle_store_async running test");
+    let result = match test(repo).await {
+        Ok(_) => true,
+        Err(err) => {
+            eprintln!("Error running test. Err: {err}");
+            false
```
CleanCut: What if the test called maybe_cleanup_repo and then returned this error? That way its info would land in the official test failure output, which would make this a lot easier to debug if it trips in CI.

crates/lib/src/model/repository/local_repository.rs
```diff
@@ -192,11 +210,20 @@ impl LocalRepository {
...
+    pub fn new_with_merkle_store_kind(
+        path: impl AsRef<Path>,
+        storage_config: Option<StorageConfig>,
+        merkle_store_kind: MerkleStoreKind,
     ) -> Result<LocalRepository, OxenError> {
         let path = path.as_ref().to_path_buf();
         let storage_config = storage_config.unwrap_or_default();
         let version_store = create_version_store(&path, &storage_config)?;
-        let (m_store, merkle_store_kind) = Self::load_merkle_store(path.clone(), None)?;
+        let m_store = Self::load_merkle_store(path.clone(), merkle_store_kind, false)?;
```
CleanCut, line 234: I think this call might be skipping the config's value here. Is that the intended behavior, or should it be reading from the config?

```suggestion
        let m_store = Self::load_merkle_store(path.clone(), merkle_store_kind, config.vfs.unwrap_or(false))?;
```

CleanCut (approved):
> Migrate up & down
> Adds a new oxen migration command to allow existing FileBackend repos
> to migrate to the LmdbBackend. Also implements a down migration, allowing
> LMDB-using repositories to switch back to the custom file format implementation.

~I don't see any migration stuff at all (???). How do we convert to LMDB on the server and client?~ Update: I see that has been moved to #551

crates/lib/src/config/repository_config.rs
malcolmgreaves: Whoops, thank you for flagging. Will fix.

crates/lib/src/model/repository/local_repository.rs
malcolmgreaves: Taking a look at implementing this.

crates/lib/src/core/db/merkle_node/lmdb/unpack.rs
malcolmgreaves: Nice performance call-out. Let me take a look at refactoring for streaming!

crates/lib/src/model/repository/local_repository.rs
malcolmgreaves: There's no `config: RepositoryConfig` instance here. In the earlier version no config implied `is_vfs` was `false`.

CleanCut: It's on [line 228](https://github.com/Oxen-AI/Oxen/pull/546/changes#diff-54874de6772880d4d055a3dbb4d744047c2ea20231e90c3d5111a848205a82c1R228), currently-named `storage_config`.

crates/lib/src/test.rs
malcolmgreaves: Yeah, the failing closures here can make for some not fun to read stack traces. I think if it was a macro then it would lead to a better stack trace. It should show up in-lined to the actual failing test IIUC. Let me try that transformation.

CleanCut: I was just trying to say return the error instead of converting it to `false`!

I was not in any way suggesting macros. Macros are notoriously opaque and difficult to deal with when there are errors. 😬

crates/lib/src/model/repository/local_repository.rs
malcolmgreaves: I think this was something that changed with the earlier PR that introduced the new `StorageConfig` type:
```rust
pub struct StorageConfig {
    pub kind: StorageKind,
    pub versions_path: Option<PathBuf>,
}
```
There's no VFS bool parameter on this type. Is there supposed to be one?

CleanCut: You're right. I definitely got my wires crossed somewhere. Sorry about that!

crates/lib/src/core/db/merkle_node/lmdb/unpack.rs
malcolmgreaves: I'm following up on this with: https://github.com/Oxen-AI/Oxen/pull/562

PR #548, opened 2026-05-12
Document error handling patterns for the codebase

malcolmgreaves:
Adds a new `docs/` top-level directory to contain developer documentation for the
entire Oxen codebase. The first entry documents the goals for consitent error
handling in the Rust code. Links to the README's contributing section link here too.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -119,6 +119,8 @@ oxen push origin main               # Push to remote
 - Make sure there's an `OxenError` variant for every error type. Be liberal in wrapping other modules error types, or other specific error types, in a new variant. Use a `Box<>` wrapper for it and have a `#[from]` to derive.
 - `OxenError` is the top-level type for everything. If you need to unify different error types into one, use `OxenError`. These kinds of functions should return `Result<T, OxenError>`
 - Implement proper error propagation through the `?` operator.
+- Never write code that uses `OxenError::Basic` or `OxenError::InternalError`. Do not use their constructor methods either (`OxenError::basic_str` and `OxenError::internal_error`, respectively). These variants are deprecated and will be removed. As a general principle, never encode an error as a string: it throws away valuable structured information about the error, which makes it impossible for a caller to handle the error programmatically. Instead, make a structured error variant of an appropriate error `enum` that describes the error. Include a `#[error("...")]` on the variant to provide a helpful user-facing error message describing the problem and, if applicable, a possible command or procedure the user can perform to fix the error.
+- If making logically related code that all share the same kind of errors, then strongly consider making a unique error `enum` _for that code only_. If that code is called by other code that has a different error enum, then define an explicit `impl From<SourceError> for TargetError` conversion to convert. If that makes the code messy, then make a conversion into `OxenError` and upgrade the calling function to return `OxenError` instead.
 
 # Making Changes
 
--- a/docs/error_handling.md
+++ b/docs/error_handling.md
@@ -1,0 +1,50 @@
+# Error Handling in Oxen
+
+This document defines and describes the standards for error handling in the Oxen codebase.
+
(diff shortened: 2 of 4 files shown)


.claude/CLAUDE.md
```diff
@@ -119,6 +119,8 @@ oxen push origin main               # Push to remote
...
+- Never write code that uses `OxenError::Basic` or `OxenError::InternalError`. Do not use their constructor methods either (`OxenError::basic_str` and `OxenError::internal_error`, respectively). These variants are deprecated and will be removed. As a general principle, never encode an error as a string: it throws away valuable structured information about the error, which makes it impossible for a caller to handle the error programatically. Instead, make a structured error variant of an appropriate error `enum` that describes the error. Include a `#[error("...")]` on the variant to provide a helpful user-facing error message describing the problem and, if applicable, a possible command or procedure the user can perform to fix the error.
+- If making logically related code that all share the same kind of errors, then strongly consider making a unique error `enum` _for that code only_. If that code is called by other code that has a different error enum, then define a `From<.> impl` to convert. If that makes the code messy, then make a conversion into `OxenError` and upgrade the calling function to return `OxenError` instead.
```
CleanCut: One thing to flag: I think this counts as a shift in policy, so it'd probably be good to get a thumbs-up from @jcelliott or @gschoeni too before landing it.

docs/error_handling.md
```diff
@@ -0,0 +1,50 @@
...
+   provide guidance or instructions the user can follow that will rectify the issue.
+
+5. Errors should be defined as close to the using code as possible. Strongly perfer making error `enum`s
+   that are specific to a module, package, or crate. For example, wrappers and helpers on networking
+   code should use a streamlined locally defined error `enum` that is specific to _networking_ errors.
+   Calling code should rely on `From` implementations to convert from one logically related set of code's
+   error type into an appropriate unifying container error type.
+
+6. For `liboxen`, the top-level unifying error type is `OxenError`. All library code must have some
+   conversion(s) that allow a more specific error type to be wrapped as an `OxenError`.
```
CleanCut: I want to make sure I'm reading these two requirements right, because they feel like they might be pulling in opposite directions. To paraphrase what I'm seeing:

- Make specific errors close to where we use them, so we can use separate error types.
- Convert all specific errors to an OxenError variant, so we can use just one error type.

Am I missing something that ties them together?

CleanCut (approved):
It feels like there may be two different approaches at play at once here, but we can work with it.

Agreed on the rabbit's point: spelling and the like are worth cleaning up.

.claude/CLAUDE.md
malcolmgreaves: Thanks for the call out. This is something we've been working towards for a few weeks now with Josh.

docs/error_handling.md
malcolmgreaves: Yes, it is indeed a balance. It's hard to provide a one-size fits all rule, so it will take a good deal of judgement and best-effort at keeping both arms in balance. 

To explain a bit further, `OxenError` is the unifying error type for all of `liboxen`. The intention is to mirror things like `MerkleDbError` having a `#[from]` variant in `OxenError`.

The end result is more of Oxen's business logic can be encoded in the type system. This makes code easier to read and lets the compiler catch mistakes. e.g. if we see `-> Result<(), MerkleDbError>` it immediately narrows down what could be going wrong. For callers, there's less possibilities to handle when matching on `Err`. And it prevents us from accidentally returning an unrelated variant.

Happy to answer any other clarifications or questions!

.claude/CLAUDE.md
CleanCut: Acknowledged 👍🏻

docs/error_handling.md
CleanCut: Okay 👍🏻

PR #551, opened 2026-05-13
`oxen migrate` between file and LMDB Merkle backends

malcolmgreaves:
Adds a new `oxen migrate` command: `switch_merkle_store_to_lmdb`. This `up` migration
allows existing `FileBackend` repos to migrate to the `LmdbBackend`. Also implements a
`down`, migration, allowing LMDB-using repositories to switch back to the custom file
format implementation.

This is an **optional** migration. It is never required to run, so users must supply
`--run-optional` in order to perform the migration if it is applicable. If the
migration isn't applicable (e.g. `up` but it's already using LMDB, or `down` but
it is already using the `FileBackend`), then it is a no-op with an explanatory message.

Also, updates `lmdb_backend_options` to automatically round-up the desired mmap size
to the nearest multiple of the OS's page size. LMDB requires its mmap file to be a
multiple of the page size.

### DIFF
--- a/crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
+++ b/crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
@@ -1,0 +1,530 @@
+use std::path::Path;
+
+use bytesize::ByteSize;
+
+use super::{Direction, Migrate};
+
+use crate::config::RepositoryConfig;
+use crate::config::repository_config::MerkleStoreKind;
+use crate::constants::{NODES_DIR, OXEN_HIDDEN_DIR, TREE_DIR};
+use crate::core::db::merkle_node::LmdbBackend;
+use crate::core::db::merkle_node::file_backend::FileBackend;
+use crate::core::db::merkle_node::lmdb::{lmdb_backend_options, lmdb_dir_location};
+use crate::error::OxenError;
+use crate::model::merkle_tree::MerkleReader;
+use crate::model::merkle_tree::merkle_writer::{MerkleWriteSession, MerkleWriter};
+use crate::model::merkle_tree::node::EMerkleTreeNode;
+use crate::model::{LocalRepository, MerkleHash};
+use crate::util::progress_bar::{ProgressBarType, oxen_progress_bar};
+use crate::{repositories, util};
+
+/// Map-size hint for the destination LMDB env when migrating into LMDB. LMDB
+/// only consumes what's actually written (the file is sparse), so this is a
+/// ceiling rather than an allocation. 64 GiB comfortably accommodates the
+/// largest Merkle trees we've seen in practice; users can grow it later by
+/// re-opening the env with a larger map.
+const LMDB_MIGRATION_MAP_SIZE_BYTES: ByteSize = ByteSize::gib(64);
+
+/// Transcodes the repository's Merkle tree node store between the file-based backend
+/// ([`FileBackend`]) and the LMDB backend ([`LmdbBackend`]).
+///
+/// - `up`: file → LMDB. Reads every node out of `.oxen/tree/nodes/...`, writes it into
+///   a freshly-opened LMDB env at `.oxen/lmdb_merkle_tree_store/`, flips
+///   `RepositoryConfig::merkle_store_kind` from `File` to `Lmdb`, then removes the
+///   old tree node directory.
+/// - `down`: LMDB → file. The mirror image: reads from LMDB, writes the file-based
+///   layout, flips the config back to `File`, and removes the LMDB env directory.
+///
+/// Both directions are idempotent: re-running on an already-migrated repo is a no-op
+/// guarded by the per-repo helpers' explicit kind check.
+pub struct SwitchMerkleStoreToLmdbMigration;
+
+impl Migrate for SwitchMerkleStoreToLmdbMigration {
+    fn name(&self) -> &'static str {
+        "switch_merkle_store_to_lmdb"
+    }
+
+    fn description(&self) -> &'static str {
+        "Transcode the repository's Merkle tree node store from the file-based \
+         backend to LMDB. `down` reverts to the file-based backend."
+    }
+
+    /// Migrate from [`FileBackend`] to [`LmdbBackend`].
+    fn up(&self, repo: LocalRepository) -> Result<(), OxenError> {
+        if !SwitchMerkleStoreToLmdbMigration.is_applicable(Direction::Up, &repo)? {
+            log::info!(
+                "File → LMDB migration not applicable for repo {:?}",
+                repo.path
+            );
+            println!(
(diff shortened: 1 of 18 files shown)


malcolmgreaves:
NOTE: Stacked PR!! https://github.com/Oxen-AI/Oxen/pull/546 must be merged first!!

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
```diff
@@ -0,0 +1,502 @@
...
+                        util::fs::canonicalize(&repo_path),
+                        err
+                    );
+                }
+            }
+        }
+        bar.inc(1);
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
CleanCut: Building on this, what would you think about also collecting and surfacing:

- the total count of successes and failures
- the path and error for each failure

That would make it a lot easier to diagnose what went wrong when a batch partially fails.

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
```diff
@@ -0,0 +1,502 @@
...
+                            "Could not clean up old file-tree dir for repo {:?}\nErr: {}",
+                            repo_path,
+                            err
+                        );
+                    }
+                }
+                Err(err) => {
+                    log::error!(
+                        "Could not migrate merkle store (file → LMDB) for repo {:?}\nErr: {}",
+                        util::fs::canonicalize(&repo_path),
```
CleanCut: I think this is meant to be displaying the path here rather than a `Result`.  Does that match what you had in mind?

We could simplify to either:

```suggestion
                        repo_path,
```

or

```suggestion
                        util::fs::canonicalize(&repo_path)?,
```

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
```diff
@@ -0,0 +1,502 @@
...
+            run_down_on_all_repos(path)?;
+        } else {
+            run_down_on_one_repo_at(path)?;
+        }
+        Ok(())
+    }
+
+    fn is_needed(&self, repo: &LocalRepository) -> Result<bool, OxenError> {
+        Ok(repo.merkle_store_kind() == MerkleStoreKind::File)
+    }
```
CleanCut: ⚠️ Heads up — I think as written this would have the effect of forcing all users to upgrade to the LMDB implementation, since most commands immediately error out otherwise. That would conflict with [what was said yesterday in slack](https://oxenai.slack.com/archives/C0AJ635QE3D/p1778796228228139?thread_ts=1778632654.279279&cid=C0AJ635QE3D):

> Yup -- the FileBackend isn't going anywhere. It's still the default and it's functionality is unchanged. So for nex they can keep using it without any issues.

So in practice, this logic would end up making LMDB the only option for clients.

The simple workaround of having `is_needed` return `false` doesn't quite get us there either. That would prevent the migration from running at all.

A couple of paths we could consider:

- If we want to support both backends long-term, it might be worth introducing a new concept (maybe "conversions"?) — similar to migrations, but independent and giving the permanent ability to switch back and forth.
- If we're planning to move everyone to lmdb eventually, we could adjust migrations to allow optional upgrades now, where they don't bug the user while they're still optional.

Curious which direction you're leaning.

CleanCut (commented):
Once we've worked through the forced-upgrade piece, I'd be happy to take another pass at this. Just ping me when you'd like the re-review.

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
malcolmgreaves: Thanks for flagging this! This is a bug -- it should not have been "if it's = to file." It's optional, so I'm changing it to return `Ok(false)` always.

CleanCut: > The simple workaround of having `is_needed` return `false` doesn't quite get us there either. That would prevent the migration from running at all.

> I'm changing it to return `Ok(false)` always.


As far as I can tell, we can't do the migration _at all_ now. 😬

malcolmgreaves: Ah darn, thank you for pointing that out. I should have read your first message more carefully. 🤦

I'll work on a new concept for this to use this here. I'll put that in another PR and merge that first. On this:
> If we want to support both backends long-term, it might be worth introducing a new concept (maybe "conversions"?) — similar to migrations, but independent and giving the permanent ability to switch back and forth.
> If we're planning to move everyone to lmdb eventually, we could adjust migrations to allow optional upgrades now, where they don't bug the user while they're still optional.

I think the former is a bit more flexible and it's what we want to do right now. I think if we do this it covers the 2nd case too.

AFAIK we don't have any guarantees of any specific migration sticking around. It's ok to delete a migration if we want to support something else. So, the way I see it, when we drop support for the file backend, we can also delete this "optional" migration and replace it with a migration that always upgrades to lmdb.

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
malcolmgreaves: I think it'll print `Ok(...)` in the success case, but this isn't very clean looking. I don't think I need to canonicalize here so I'll use `repo_path` directly.

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
CleanCut: Ya, a semi-permanent "conversions" between supported stuff would be a good feature to add.

malcolmgreaves:
NOTE: Stacked PR! https://github.com/Oxen-AI/Oxen/pull/567 MUST BE MERGED **BEFORE** THIS PR.

crates/lib/src/command/migrate/m20260512000000_switch_merkle_store_to_lmdb.rs
malcolmgreaves: Fixed! Implemented optional migrations in a new PR: https://github.com/Oxen-AI/Oxen/pull/567 . This one is now stacked on it.

malcolmgreaves:
Also @CleanCut added the accumulation of errors from running all up & down migrations.

CleanCut (approved):
Items to consider:
- You may have a Windows-specific bug to fix
- If the migration fails with an error, the partially written conversion still exists on disk in an unknown state. Should we clean up the partial conversion upon error?
- It might be good to run the tests on a fixture that involves more than a single commit

PR #552, opened 2026-05-15
Fix intermittent macOS CI failure: rustup-init invoked instead of cargo

malcolmgreaves:
1. Don't use the login shell with `bash` in the macOS CI tests.
2. Don't cache binaries in the `rust-cache`.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -27,7 +27,18 @@ jobs:
             shell: bash -leo pipefail {0}
           - os: macos-latest
             platform: macos
-            shell: bash -leo pipefail {0}
+            # Non-login shell (no `-l`) on macOS. A login shell sources /etc/profile
+            # and ~/.profile, which on macOS runs /usr/libexec/path_helper and
+            # Homebrew's shellenv hooks. Those can reorder PATH so that `cargo`
+            # resolves to the Homebrew `rustup` formula's keg, where `cargo` is a
+            # symlink to an env-script wrapper that does
+            # `exec "/path/to/libexec/bin/rustup-init" "$@"`. The wrapper makes
+            # argv[0] end in "rustup-init", so rustup's argv[0]-based dispatcher
+            # picks setup_mode (the installer) instead of proxy_mode and rejects
+            # `build` as an "unexpected argument". Linux is left as a login shell
+            # because it isn't affected (no path_helper) and changing it would
+            # broaden the blast radius unnecessarily.
+            shell: bash -eo pipefail {0}
           - os: windows-x64-8cpu-32ram-300ssd
             platform: windows
             shell: powershell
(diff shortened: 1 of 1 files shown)


malcolmgreaves:
I didn't really know what was going on with these intermittent failures on mac os's `cargo build` command in the Rust + Python tests. I pointed opus 4.7 max at the failed CI logs with the additional debug logging enabled and let it churn. It came up with these changes and accompanying explanation:

```
The macOS jobs in ci_test.yml have been intermittently failing at
`cargo build --workspace` with:

    error: error: unexpected argument 'build' found

    Usage: rustup-init[EXE] [OPTIONS]

This is rustup running in setup_mode (the installer) instead of
proxy_mode (forwarding cargo invocations). It's been impossible to
reproduce locally, and a CI re-run "fixes" it eventually — both clues
pointed at shared state across runs.

Analysis
--------

Two compounding causes:

1. The workflow's macOS shell was `bash -leo pipefail` — a *login*
   shell. Login mode sources /etc/profile and ~/.profile; on macOS that
   runs /usr/libexec/path_helper and Homebrew's shellenv hooks, which
   reorder PATH. By contrast, the steps inside
   `actions-rust-lang/setup-rust-toolchain@v1` use
   `bash --noprofile --norc -e -o pipefail`, and their cargo/rustup
   calls in the same job succeeded — so the difference was the shell.

   The macos-latest runner image installs Rust via `brew install
   rustup-init` (now the `rustup` Homebrew formula). That formula uses
   `bin.env_script_all_files`, so the proxy `cargo` in the brew keg is
   a symlink to a wrapper script that does
   `exec "/path/to/libexec/bin/rustup-init" "$@"`. When the login-shell
   PATH reorder puts the keg ahead of ~/.cargo/bin, `cargo build`
   resolves to that wrapper, the wrapper passes the literal
   `…/libexec/bin/rustup-init` as argv[0], and rustup's argv[0]-based
   dispatcher (file_stem starts with "rustup-init" → setup_mode) hands
   the args to the installer parser, which rejects `build` as
   unexpected.

2. `Swatinem/rust-cache@v2` defaults to `cache-bin: true`, which caches
   ~/.cargo/bin. Since rustup ≥1.28 (this run used 1.29.0), the proxy
   binaries (cargo, rustc, …) are *symlinks* to the rustup regular
   file. The action's cleanBin step skips symlinks (dirent.isFile()
   is false for them) but can delete the rustup target itself, since
   rustup isn't tracked in .crates2.json. A poisoned cache from a main
   run is then restored across all subsequent PR runs (save-if is
   restricted to main), which is exactly the "intermittent, fixes
   itself eventually" symptom — the cache stays bad until something
   changes the cache key.

The login shell is the trigger; the bad cache makes it reliable.

Fix
---

Two surgical workflow changes:

* Drop `-l` from the macOS shell in both jobs (`rust-test`,
  `python-test`). The runner image places brew, cargo, python, uv,
  pytest on the default non-login PATH via /etc/paths.d, so nothing
  in the workflow needs login-shell init. Linux is intentionally left
  alone — it isn't failing (no path_helper) and changing it broadens
  the blast radius unnecessarily.

* Add `cache-bin: false` to both `Swatinem/rust-cache@v2` blocks. This
  closes the cache-poisoning pathway. cargo-nextest is the only
  binary we'd otherwise cache and it's already installed by
  taiki-e/install-action, so cost is nil.

Inline comments in the workflow file capture the full reasoning so
future readers don't have to rediscover it.
```

CleanCut (approved):
Let's give it a try. We can't see the effect on build times until this is run in `main` so it caches result and then have something else load the cache. 🤞🏻 Hopefully we fix the problems without adding build time.

PR #553, opened 2026-05-15
Upgrade to astral-sh/setup-uv@v8

CleanCut:
Fix CI warnings like this:

image

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -120,7 +120,7 @@ jobs:
           workspaces: "."
 
       - name: Install uv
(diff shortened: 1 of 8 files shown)


malcolmgreaves:
Is this the same version that is used in the install pre-reqs script?

CleanCut:
> Is this the same version that is used in the install pre-reqs script?

This PR bumps the action's own version tag (the code GitHub runs in CI). The action itself installs the latest `uv` both before and after this bump. The `install-prereqs` script also installs the latest `uv`. So if the question is whether they end up with the same `uv` installed: Yes, they do.

Happy to dig in further if you were asking about something else.

malcolmgreaves (approved):

PR #558, opened 2026-05-18
S3 (13.5): Stop serializing LocalRepository, Workspace, ParsedResource

CleanCut:
These three internal models were carrying Serialize/Deserialize/ToSchema only because `ParsedResource` was embedded in `MetadataEntry`'s wire shape. Moving `MetadataEntry.resource` (and `DiffEntry`'s resource fields, and `ParseResourceResponse.resource`) to `ParsedResourceView` lets the internal types drop those derives — and lets `NoopVersionStore` (a placeholder that existed only to satisfy `LocalRepository`'s `Deserialize)` go away too.

@malcolmgreaves This unblocks switching `LocalRepository.merkle_store` from `Option<Arc<...>>` to `Arc<...>`, and removing `LocalRepository.merkle_store_kind`.

The untagged `EMetadataEntry` enum used to distinguish its two variants by the `Workspace`-vs-`WorkspaceView` shape nested inside `resource.workspace`. Both variants now share `ParsedResourceView,` so that structural discriminator is gone. Instead, I set `deny_unknown_fields` on `MetadataEntry` plus `skip_serializing_if=None` on `WorkspaceMetadataEntry.changes`. This makes the presence of `changes` on the wire the new discriminator between which enum variant to deserialize to.

Wire shape change: `MetadataEntry.resource.workspace` now serializes as `{id, name, commit}` instead of the full `Workspace`-with-`LocalRepository` shape that our clients didn't actually use.

Breakage for old clients is extremely minimal. As far as I can tell, the only surface this breaks for old clients is if you use the Python bindings to call `RemoteRepo(name, host, revision=SOME_ID` where `SOME_ID` is a workspace ID.


crates/lib/src/model/workspace.rs
```diff
@@ -16,7 +16,10 @@ pub struct WorkspaceConfig {
     pub workspace_id: Option<String>,
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+/// In-process model. Not part of any API wire shape — use `WorkspaceView`
```
malcolmgreaves:
```suggestion
/// In-process model for workspace state.
/// Not part of any API wire shape — use [`WorkspaceView`] for communicating workspaces between client & server.
```
Or something similar.

crates/lib/src/model/workspace.rs
```diff
@@ -16,7 +16,10 @@ pub struct WorkspaceConfig {
...
+/// (`{id, name, commit}`) when crossing the wire; the embedded `LocalRepository`s here
+/// would bloat the response and leak server-internal storage state.
```
malcolmgreaves:
```suggestion
```
I think this comment makes more sense on a `WorkspaceView` struct. The first sentence directing users to this view type is good!

crates/lib/src/repositories/diffs.rs
```diff
@@ -412,15 +412,15 @@ pub async fn diff_path(
...
                     path: head_path.as_ref().to_path_buf(),
-                    ..ParsedResource::default() //TODO: Fill in other fields as well
+                    ..ParsedResourceView::default() //TODO: Fill in other fields as well
                 }),
-                base_resource: Some(ParsedResource {
+                base_resource: Some(ParsedResourceView {
                     commit: Some(base_commit.clone()),
                     path: base_path.as_ref().to_path_buf(),
-                    ..ParsedResource::default() //TODO: Fill in other fields as well
+                    ..ParsedResourceView::default() //TODO: Fill in other fields as well
```
malcolmgreaves, line 423: Are there other fields that should be filled-in here? Or is these line comments something that can be removed now?

malcolmgreaves (approved):
LGTM. It's a nice cleanup to use the `*View` structs and drop the needless serialization. Before we merge and release, let's just hammer out the specific details for the upgrade path for these breaking changes. Maybe we need an announcement or to include this in its own specific update. In general, I'd like for us to do more breaking changes as we're pre 1.x. But I'd like us to get into a good habit of communicating breaking changes early and announcing publicly so there's some lead time for users to upgrade.

crates/lib/src/model/workspace.rs
CleanCut: ✅ Done

crates/lib/src/model/workspace.rs
CleanCut: ✅ Done -- I added a doc comment similar to this to the `WorkspaceView` struct.

crates/lib/src/repositories/diffs.rs
CleanCut: ✅ Removed the out-of-date todo comment.

CleanCut:
> Before we merge and release, let's just hammer out the specific details for the upgrade path for these breaking changes. Maybe we need an announcement or to include this in its own specific update. In general, I'd like for us to do more breaking changes as we're pre 1.x. But I'd like us to get into a good habit of communicating breaking changes early and announcing publicly so there's some lead time for users to upgrade.

There was a discussion about this [in slack](https://oxenai.slack.com/archives/C0AJ635QE3D/p1779209949428699?thread_ts=1779122737.415919&cid=C0AJ635QE3D) and @jcelliott indicated that the minor breaking change in this PR is unlikely to affect the small number of Python users we have and that we should follow up on announcement/version improvements later.

I created linear issues ENG-1037 and ENG-1038 as followups.

PR #559, opened 2026-05-18
Remove lmdb experiment

CleanCut:
Pure cleanup! Now that we've got an actual lmdb implementation, we don't need the experiment! 🎉


CHaiz15 (approved):
🚮

PR #564, opened 2026-05-20
S3 (13.6): Add server `StoragePolicy` and config file

CleanCut:
Adds the server-side storage policy that validates caller-requested storage backends on repo creation, plus the TOML config-file machinery admins use to populate it. Closes ENG-1010; supersedes ENG-1033.

- `config::Config` is the top-level deserialize target — currently a tiny wrapper around `StoragePolicy`. Carried verbatim on `OxenAppData.config`, so future top-level keys (auth, telemetry, …) are reachable as `app_data.config.<key>` from any handler — one edit, no plumbing.
- `load_server_config()` reads + parses the file; validation errors surface as `ServerError::ConfigParse` at startup (the validating `TryFrom` fires during `toml::from_str`).
- `config::storage_policy::StoragePolicy` — the admin's "what backends will this server host?" answer. If the file is absent, or the section of the file is absent, then we get a default config with only "local" version storage configured -- which is the current behavior anyway.
- We don't yet support actually setting the S3 values in the config file (it will fail with a "not implemented" error). We'll be adding that in another PR.
- We'll be adding admin docs in another PR as well.

The server config file format looks like this:

```toml
[storage]
backends = ["local", "s3"]   # first element is the server default
s3_bucket = "my-oxen-bucket"
```


crates/server/src/config/storage_policy.rs
```diff
@@ -0,0 +1,321 @@
...
+
+use crate::errors::OxenHttpError;
+
+/// Server-side storage policy.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(try_from = "StoragePolicyRaw")]
+pub struct StoragePolicy {
+    /// Non-empty; index 0 is the server's default. No duplicate kinds.
+    backends: Vec<StorageKind>,
+    /// Must be non-empty when `StorageKind::S3 ∈ backends`. When S3 is not enabled,
```
malcolmgreaves: > "When this enum value is in the set of backends" 

What does this mean? I don't follow this comment.

> Must be non-empty 

If this can be empty then it should be `Optional<String>` to denote these two states.

crates/server/src/config/storage_policy.rs
```diff
@@ -0,0 +1,321 @@
...
+
+#[derive(Deserialize)]
+struct StoragePolicyRaw {
+    backends: Vec<StorageKind>,
+    #[serde(default)]
+    s3_bucket: String,
+}
+
+impl TryFrom<StoragePolicyRaw> for StoragePolicy {
+    type Error = OxenError;
```
malcolmgreaves: This looks like a good opportunity to define a storage policy error enum and have a `#[from]` in `OxenError`.

crates/server/src/app_data.rs
```diff
@@ -1,19 +1,23 @@
...
-            path: self.path.clone(),
+            path,
+            config: Config::default(),
         }
     }
+
+    /// Construct with an explicit server config.
+    pub fn with_config(path: PathBuf, config: Config) -> OxenAppData {
+        OxenAppData { path, config }
+    }
```
malcolmgreaves:
```suggestion
```
Does this need a new `pub fn` if it's only used once? Why not use the struct initializer directly?

crates/server/src/config.rs
```diff
@@ -0,0 +1,64 @@
+//! Server-side configuration.
+//!
+//! - [`Config`] — the TOML deserialize target for `--config`.
+//! - [`storage::StoragePolicy`] — the
```
malcolmgreaves: Looks like this comment is incomplete?

crates/server/src/config/storage_policy.rs
```diff
@@ -0,0 +1,321 @@
...
+use serde::Deserialize;
+
+use crate::errors::OxenHttpError;
+
+/// Server-side storage policy.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(try_from = "StoragePolicyRaw")]
+pub struct StoragePolicy {
+    /// Non-empty; index 0 is the server's default. No duplicate kinds.
+    backends: Vec<StorageKind>,
```
malcolmgreaves: So it's valid to have repeated `StorageKind` values?

crates/server/src/config/storage_policy.rs
malcolmgreaves: After looking through this PR, it seems like the desired use is to have a configuration file where local and s3(bucket) can be listed. This struct relies on some complex logic with specific values at runtime being ok while other values at runtime are invalid. There's not a lot of type-level support that can catch these errors.

Instead, I'd encourage a redesign that makes these invalid states impossible to represent.

For instance, perhaps:
```rust
pub struct StoragePolicy {
  backends: Vec<StorageBackend>,
}

pub enum StorageBackend {
   Local,
   S3 { bucket: String },
}
```

This should work for the following toml:
```toml
  [[backends]]
  type = "Local"

  [[backends]]
  type = "S3"
  bucket = "hello"
```

Though this has the same issue as the repeated storage backends of the original `struct` definition. Though, presumably we should be able to handle different S3 buckets for different repositories? (Perhaps that's not an explicitly supported feature!)

malcolmgreaves: Taking a step back -- this seems like a lot of general functionality for something that is maybe being overcomplicated. There's always the Local storage backend. There's only optional support for an S3 backend in this PR if IIUC.

Why not have a single `s3_bucket: Option<String>` parameter in the configuration then? 🤔 

The storage backends for any `StoragePolicy` then is a function:
```rust
fn backends(&self) -> Vec<StorageKind> {
    if let Some(config) = self.config {
        vec![StorageKind::Local, StorageKind::S3]
    } else {
        vec![StorageKind::Local]
    }
}
```

crates/server/src/config.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod storage_policy;
+
+use crate::config::storage_policy::StoragePolicy;
+use serde::Deserialize;
+
+/// TOML deserialization target for the server's `--config` flag.
+///
+/// Future top-level keys (auth options, telemetry settings, etc.) land here without restructuring.
+#[derive(Debug, Clone, Default, Deserialize)]
+pub struct Config {
```
malcolmgreaves, line 15: This looks like it could be `pub(crate)`, right?

Do the other new things introduced here need to be `pub` or does `pub(crate)` suffice?

crates/server/src/main.rs
```diff
@@ -502,22 +546,25 @@ fn init_metrics() -> Result<Option<MetricsGuard>, ServerError> {
     Ok(None)
 }
 
+/// CLI/env-derived flags collected at startup, distinct from the TOML-loaded
+/// [`config::Config`] (which carries settings from disk).
```
malcolmgreaves, line 553: Should these CLI settings be unified into this new config file? It seems like this PR is introducing a new concept for configuring the server, but it seems like it's leaving out all of the existing configuration settings. Any explicit decisions to not include existing options in `Config` here?

malcolmgreaves (commented):
A handful of comments. I'm unsure about the `backends` and `bucket` setup for `StoragePolicy`. And I'm unsure why its introducing a config file that doesn't refactor existing configuration settings into it. Happy to discuss more. Perhaps there's a bigger picture that I'm not seeing that puts this PR into more context.

crates/server/src/config/storage_policy.rs
```diff
@@ -0,0 +1,321 @@
...
+            }
+        }
+        // Orphan bucket (set when S3 isn't in backends) is silently allowed so admins
+        // toggling backends on/off during config iteration don't trip a validation error.
+        if raw.backends.contains(&StorageKind::S3) && raw.s3_bucket.is_empty() {
+            return Err(OxenError::StoragePolicyEmptyS3Bucket);
+        }
+        Ok(Self {
+            backends: raw.backends,
+            s3_bucket: raw.s3_bucket,
```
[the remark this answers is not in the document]
CleanCut: We'll address validation when we begin using the value.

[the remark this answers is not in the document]
CleanCut: We'll deal with any semantically-bad value later when we attempt to establish a connection to S3.

crates/server/src/config/storage_policy.rs
CleanCut: No, there's no way to create repeated values. The validating `TryFrom` rejects duplicates during parsing, so by the time you have a `StoragePolicy` the list is guaranteed unique.

crates/server/src/config/storage_policy.rs
CleanCut: > After looking through this PR, it seems like the desired use is to have a configuration file where local and s3(bucket) can be listed. This struct relies on some complex logic with specific values at runtime being ok while other values at runtime are invalid. There's not a lot of type-level support that can catch these errors.
> 
> Instead, I'd encourage a redesign that makes these invalid states impossible to represent.

I totally agree that making invalid states unrepresentable is the right goal. That was my goal here too.

I did start with a design closer to what you're describing, but landed on this shape because it keeps the user-facing config really simple (the TOML below) while still making invalid states impossible to represent at the type level after construction. The validation is just [these ~10 lines](https://github.com/Oxen-AI/Oxen/pull/564/changes#diff-4dc40b6dc6570ee89b1319cc177eea3c42e1331ab970ee85ad1dc4e52330f8b7R44-R57) in the `TryFrom`, and once you have a `StoragePolicy` value the invariants hold.

```toml
[storage]
backends = ["local", "s3"]   # first element is the server default
s3_bucket = "my-oxen-bucket"
```

> There's always the Local storage backend. There's only optional support for an S3 backend in this PR if IIUC

Good catch on the current state. In this PR, yes, `Local` is always present. The end goal is to support `S3`-only deployments too (no `Local` backend at all), so the policy type is designed to allow that even though the change will made in a later PR.

crates/server/src/config/storage_policy.rs
CleanCut: I'd lean toward holding off until we need to inspect or branch on the errors. Right now the errors only flow upward into `OxenError`, so a dedicated enum + `#[from]` would mostly be ceremony. Happy to revisit the moment we have a caller that wants to do something specific with them.

crates/server/src/app_data.rs
CleanCut: ✅ Switched to the inline form.

crates/server/src/config.rs
CleanCut: Ha, yep. I left that one hanging. I'll finish it up. ✅ Done.

crates/server/src/config.rs
CleanCut: Fair question. I'd prefer to leave them as `pub`. With no external library consumers of the server crate, `pub(crate)` is a cosmetic distinction here, and I'd rather not spend cycles tightening visibility that nothing depends on.

crates/server/src/main.rs
CleanCut: Good question! Eventually, yes. Right now, no. I'm intentionally scoping this PR to a new config file with only the functionality we need ourselves. Unifying configs, CLI flags, and env vars is a larger project I'd like us to take on down the road, just not in this change.

crates/server/src/config.rs
```diff
@@ -0,0 +1,64 @@
+//! Server-side configuration.
+//!
+//! - [`Config`] — the TOML deserialize target for `--config`.
+//! - [`storage::StoragePolicy`] — the version store storage policy for the server.
```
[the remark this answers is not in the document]
CleanCut: ✅ Fixed

crates/server/src/config/storage_policy.rs
malcolmgreaves: 🤔 This code is only defined in the server crate, so it can't be used in `liboxen`. Should `liboxen` be the crate to define this code's errors?

It's simpler if this code defines its own error conditions: it keeps `OxenError` smaller, and callers of this don't have to potentially wade through unrelated `OxenError` variants. And since it's server code, the server main's error enum can have a `#[from]` defined here to make the conversion easy as well.

Here's the requested change: https://github.com/Oxen-AI/Oxen/pull/570

CleanCut: Fair point. Merged #570.

crates/server/src/config/storage_policy.rs
malcolmgreaves: That's a nice easy syntax for the config file!

I'm glad there's alignment on the end goal! I'm glad that the `TryFrom` route provides runtime safety for unrepresentable states. I do prefer designs that can define code in such a way that the state is actually not representable in a well-typed program.

With this struct definition, it's possible to add _new_ code that creates one outside of this `TryFrom` route incorrectly _without_ changing the existing `struct` definition. Ideally, the data structure definition would prevent this from ocurring.

What's a way that this could be defined so that if it's in memory it has to always be valid?

malcolmgreaves: > What's a way that this could be defined so that if it's in memory it has to always be valid?

If this isn't coming easily and/or quickly right now, I'm ok to keep this design as-is as long as we're willing to revisit and change this when we start building out the server configuration file later.

CleanCut: Could you walk me through the scenario you have in mind? The fields on `StoragePolicy` are private, and the only external construction paths I can find are `TryFrom<StoragePolicyRaw>` (which validates) and `Default` (valid by construction). Is there a route I'm missing?

CleanCut: Sidenote: I'm open to re-evaluating and potentially rewriting everything when we add in the cli argument and env variable handling. We could end up with very different needs when we combine everything.

malcolmgreaves: Let's say someone in the future wants a `StoragePolicy` directly, so they accidentally make a new function in the file:

```rust
fn this_shouldnt_be_representable() -> StoragePolicy {
  StoragePolicy { backends: vec![StorageKind::Local], s3_bucket: "uh-oh".to_string() }
 }
``` 

This code compiles, but it has an s3 bucket without a `StorageKind::S3` in the `backends` list. Alternative bug is having `S3` in `backends` but an empty `s3_bucket`: `StoragePolicy { backends: vec![StorageKind::Local, StorageKind::S3], s3_bucket: "".to_string() }`. Both invalid states compile.

malcolmgreaves: Ideally, we have code that makes an invalid state a compilation error.

malcolmgreaves: IIUC there's 2 potential valid storage policy states:
1. local storage only
2. local storage and S3 storage

This could be:
```rust
enum StoragePolicy {
  Local(LocalStorage),
  Both(LocalStorage, S3Storage),
}
```

Where `TryFrom` parses the config file contents into one of these two variants.

As support:
```rust
struct LocalStorage {
  storage_directory: PathBuf
}

struct S3Storage {
  s3_bucket: String,
}
```

We could maybe later add a third situation -- s3 only storage:
```rust
  S3(S3Storage),
```

To enable setups where 100% of the stored files are in S3.

CleanCut: Those literals would compile if placed inside the module, that's a fair technical point. But it's a theoretical concern: it needs someone to write a literal inside the module that contradicts the field doc-comments, survive review, and slip past tests.

But I can see that you care a lot about this, so I'll push up a change to use a design that isn't possible to misrepresent even inside the module, and I'll make it parse the existing config format into it. Then we've got the best of all worlds.

CleanCut: ✅ Done

malcolmgreaves: Thanks for discussing!

malcolmgreaves (approved):

PR #565, opened 2026-05-20
Per-process `LmdbBackend` cache with `Weak` and `Arc`

malcolmgreaves:
Creates a new `static` cache of opened LMDB environments in `lmdb::cache`. Internally,
it maps  maps each repository's root path (absolute) to a `Weak<LmdbBackend>` reference.

All creation management and access of `LmdbBackend` in oxen now goes through the cache
access function, `lmdb::cache::get_or_open`. This function hands out an `Arc<LmdbBackend>`,
allowing callers to reuse the same instance. When accessing, the internal `Weak` reference
is checked for outstanding holders of its `Arc`. The common path is to return a refcount-
bumped `Arc`. First time creation will return the only `Arc<LmdbBackend>`.

However, if the original caller ever drops this `Arc` and there are no other holders of
it, then the underyling `LmdbBackend` and its LMDB env is also dropped. Subsequent access
via `get_or_open` will re-create the LMDB environment, which is safe since the previous
LMDB env was dropped.

Accessing the cache checks for tombstoned `Weak` references and removes them. This keeps
the cache from growing unbounded and lets LMDB envs drop when they're no longer needed.

The `LmdbBackend::new` constructor is now restricted from `pub` to the `lmdb` module only.

### DIFF
--- a/crates/lib/src/core/db/merkle_node/lmdb/cache.rs
+++ b/crates/lib/src/core/db/merkle_node/lmdb/cache.rs
@@ -1,0 +1,227 @@
+//! Process-wide cache of open [`LmdbBackend`] instances, keyed by the repository root path.
+//! Use the [`get_or_open`] function to manage creation and access to all [`LmdbBackend`]s.
+//!
+//! LMDB rejects opening the same database directory twice in one process (`mdb_env_open`
+//! returns `MDB_BUSY`). Since a [`LocalRepository`] contains a Merkle tree store, it
+//! may end up holding an [`Arc`] to an [`LmdbBackend`]. Creating another [`LocalRepoistory`]
+//! at the same location while holding onto a previously created one will create a runtime `panic!`.
+//!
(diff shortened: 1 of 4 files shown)


crates/lib/src/core/db/merkle_node/lmdb/cache.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    // check if we have created an LMDB backend for this repository
+    // or if the one we created is still alive
+    if let Some(weak) = guard.get(&repo_root)
+        && let Some(strong) = weak.upgrade()
+    {
+        log::debug!("LMDB backend cache hit: {:?}", repo_root);
+        return Ok(strong);
+    }
+
```
malcolmgreaves, line 105: - change to `RwLock`
- make above acquire a read lock and exit early if possible
- on `else`, re-acquire write lock and re-check, then insert

jcelliott (approved):

PR #568, opened 2026-05-20
Cleanup Migration and CLI

malcolmgreaves:
Initializes the migrations map (ALL_MIGRATIONS) once and reuses. Uses `lazy_static!` and
adds `Sync` and` Send` markers to the `Migrate` trait. Updated callsites.

Lowered visibility in CLI crate: helpers and cli_error modules are not pub. Their members
are `pub(crate)` now. Made all `RunCmd::name()` implementations inline their `&'static str`.
Also made `Migrate::{name,description}()` inline as well.

### DIFF
--- a/crates/cli/src/helpers.rs
+++ b/crates/cli/src/helpers.rs
@@ -111,3 +111,19 @@ pub fn check_repo_migration_needed(repo: &LocalRepository) -> Result<(), OxenErr
         "Error: Migration required".to_string().into(),
     ))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use liboxen::test;
+
+    /// The LMDB merkle-store migration is optional: even on a file-backend repo
+    /// (where it's *applicable*), it must not appear in the auto-block list, or
+    /// every normal CLI operation would refuse to proceed. `run_empty_local_repo_test`
+    /// creates a file-backend repo by default — exactly the case we want to assert
+    /// stays unblocked.
+    #[test]
+    fn test_check_repo_migration_needed_ignores_optional_migrations() -> Result<(), OxenError> {
+        test::run_empty_local_repo_test(|repo| check_repo_migration_needed(&repo))
+    }
+}
--- a/crates/cli/src/helpers.rs
+++ b/crates/cli/src/helpers.rs
@@ -11,7 +11,7 @@ use colored::Colorize;
 use std::collections::HashMap;
 use std::str::FromStr;
 
-pub fn get_scheme_and_host_or_default() -> Result<(String, String), OxenError> {
+pub(crate) fn get_scheme_and_host_or_default() -> Result<(String, String), OxenError> {
     let config = AuthConfig::get_or_create()?;
     let mut default_host = (
         constants::DEFAULT_SCHEME.to_string(),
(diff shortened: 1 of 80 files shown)


malcolmgreaves:
NOTE: STACKED PR! https://github.com/Oxen-AI/Oxen/pull/551 MUST BE MERGED **FIRST**.

crates/cli/src/helpers.rs
```diff
@@ -58,7 +58,7 @@ pub async fn check_remote_version(
     Ok(())
 }
 
-pub async fn check_remote_version_blocking(
+pub(crate) async fn check_remote_version_blocking(
```
CleanCut, line 61: For the cli and server crates, I don't think we intend either to be used as a library by our customers, which means `pub` and `pub(crate)` don't make a meaningful difference here. Let me know if there's a consumer I'm missing, but otherwise we can probably focus our efforts elsewhere.

crates/lib/src/command/migrate.rs
```diff
@@ -54,24 +55,26 @@ pub trait Migrate {
...
+        map.insert(
+            AddWorkspaceNameIndexMigration.name().to_string(),
+            Box::new(AddWorkspaceNameIndexMigration),
+        );
+        map.insert(
+            SwitchMerkleStoreToLmdbMigration.name().to_string(),
+            Box::new(SwitchMerkleStoreToLmdbMigration),
+        );
+        map
+    };
```
CleanCut, line 79: Could you walk me through what this change is buying us? My read of the diff is that we're moving the same work from compile time to runtime, with a little added complexity. I'm happy to be corrected if I'm missing the upside.

CleanCut (approved):
My honest read is that I'm not seeing a clear upside in this PR, though the diff doesn't introduce any harm either. A few specific notes:

- Every `name()` method on implementations of `RunCmd` and `Migrate` got `#[inline(always)]`, but those methods are usually called through trait objects (`&dyn RunCmd`, `Box<dyn Migrate>`). Trait-object calls go through vtable dispatch, which can't be inlined, so the attribute is a no-op on these call sites. Beyond not helping, the attributes could give future readers the wrong mental model about whether inlining is happening here.
  - The same applies to `TFileNode::name()` / `TDirNode::name()` — those are also dispatched dynamically in most consumers.
- For the cli and server crates, I don't think we intend either to be used as a library by our customers, which means changing from `pub` to `pub(crate)` doesn't make a meaningful difference here. Let me know if there's a consumer I'm missing, but otherwise we can probably focus our efforts elsewhere.
- Could you walk me through what the change from a hard-coded function to a lazy-static is buying us? My read of the diff is that we're moving the same work from compile time to runtime, with a little added complexity. I'm happy to be corrected if I'm missing the upside.
- The CI and Lint problems should be resolved in one of the parent PRs, so I expect they'll clear up once you rebase.

crates/lib/src/command/migrate.rs
malcolmgreaves: The idea was to make this map once and reuse it, since once it's created it's never modified. 

But, I actually make a simplification in the first PR of this stack to turn it into a simple 2-element array. As I thought about it, I realized that the overhead of creating a hash table & hashing the keys is more than doing linear search on 2 elements. (Creating the hash is O(length of key) + constant time. Then O(1) lookup. Comparing a string against 2 other strings character-by-character is O(2 * min(string, max(migration name)). So longer in theory, but max migration name is bounded, so 2x of that is still < some constant amount of time.

PR #569, opened 2026-05-20
Enable `oxen-server` + `oxen {clone,fork}` to use LMDB Merkle stores

malcolmgreaves:
**`oxen-server --merkle-store`**
Enables `oxen-server` to use LMDB Merkle tree stores on new repositories via the
server's new `--merkle-tree` flag on startup. This defaults to `file` for existing
behavior. This option controls the `MerkleStore` backend that `oxen-server` uses
when a new repository is created. Note that existing repositories can be migrated
to either the file or LMDB backend using the migration route on the server.

**`oxen-clone --merkle-store`**
Also updates `oxen clone` to select an explicit Merkle tree store kind on clone.
Has the same `--merkle-store` flag. Ensures that the cloned repository is using
that Merkle tree store locally. Servers and clients can use different Merkle
tree stores and still be completely compatible.

**`oxen fork`**
Updates the repository forking code to properly handle copying LMDB's data file.
Uses the weak LmdbBackend cache to get the source fork's LMDB env and uses `heed`'s
`copy_to_path` function (accessed via `LmdbBackend::copy_on_disk`). Fork's recursive
file copying skips over all LMDB files.

**`MerkleStoreLmdbNotSupportedOnVfs`**
Finally, includes a minor refactor to the server repository creation code. Unifies
their custom HTTP error response handling of OxenError and makes them treat
`MerkleStoreLmdbNotSupportedOnVfs` as a HTTP Bad Request. This error is because
LMDB is not officially supported on virtual filesystems.

**`RepoNew`**
Moves `RepoNew` from `liboxen::models` to a new sub-package `liboxen::api::requests`.
Documents this struct as being the HTTP payload for a client requesting creation of
a remote repository on `oxen-server`.

### DIFF
--- a/crates/lib/src/repositories/fork.rs
+++ b/crates/lib/src/repositories/fork.rs
@@ -1,6 +1,9 @@
+use crate::config::RepositoryConfig;
+use crate::config::repository_config::MerkleStoreKind;
 use crate::constants::{OXEN_HIDDEN_DIR, WORKSPACES_DIR};
+use crate::core::db::merkle_node::lmdb::{OXEN_LMDB_MERKLE_DIR, get_or_open, lmdb_dir_location};
 use crate::error::OxenError;
-use crate::util::fs as oxen_fs;
+use crate::util::fs::{self as oxen_fs, config_filepath};
 use crate::view::fork::{ForkStartResponse, ForkStatus, ForkStatusFile, ForkStatusResponse};
 use std::fs;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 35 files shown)


malcolmgreaves:
NOTE: STACKED PR! https://github.com/Oxen-AI/Oxen/pull/565 MUST BE MERGED **FIRST**.

malcolmgreaves:
review notes, requested changes:
- don't allow `oxen` client to control merkle backend
- new `oxen-server` flag to determine which merkle backend it uses when creating all _new_ repositories


notes for production testing:
- have default still be file
- create new repo
- use migration to go from file -> lmdb
- test

crates/server/src/controllers/repositories.rs
```diff
@@ -308,9 +308,26 @@ pub async fn create(
...
 ) -> actix_web::Result<HttpResponse, OxenHttpError> {
-    data.storage_kind = Some(app_data.config.storage.resolve(data.storage_kind)?);
+    let data = {
+        let mut data = data;
+
+        if let Some(ref m_kind) = data.merkle_store_kind
+            && m_kind != &app_data.merkle_store_kind
+        {
+            log::error!(
+                "Repository creation endpoint does not allow callers to set Merkle store. \
```
CleanCut: What's the intended role of `merkle_store_kind` on the request side, given the server overrides it? Is the field there for forward-compatibility, or could it be dropped from the request type so clients don't set it?

crates/lib/src/repositories/fork.rs
```diff
@@ -53,43 +56,77 @@ pub fn start_fork(
...
-            &original_path,
-            &new_path,
-            &new_path,
-            total_items,
-            &mut copied_items,
-        ) {
-            Ok(()) => {
+
+        // if LMDB, then perform a special copy of its database file
+        if lmdb_dir_location(&original_path).is_dir() {
```
CleanCut: Would it make more sense to check the config for which merkle store kind rather than checking the existence of a directory that's tied to a specific implementation?

```suggestion
        if matches!(og_config.merkle_store_kind, MerkleStoreKind::Lmdb) {
```

CleanCut (approved):

crates/server/src/controllers/repositories.rs
malcolmgreaves: https://oxenai.slack.com/archives/C0AJ635QE3D/p1779906426528119?thread_ts=1779257742.953079&cid=C0AJ635QE3D

malcolmgreaves: Talked about this post standup w/ Josh and a few others -- here the conclusions from that conversation: 

(1) The  `RepoNew` is supposed to be the payload format for a client making a request to the server to create a new repository. So I'm adding some documentation to the struct to make that intent clear. And for organization, I'm moving it to a new package: `liboxen::api::requests::repo_new`. `requests` will serve as a place to organize things that are strictly request payloads between `oxen` <> `oxen-server`.

(2) Audit the codebase to switch any local-use of `RepoNew` to create a `LocalRepository` is switched to some other constructor.

(3) `RepoNew` will not have `merkle_store_kind`. Clients cannot request this. It's a server setting configured at startup.

PR #578, opened 2026-05-26
Remove CLI-only errors from OxenError, have CLI use anyhow::Error

malcolmgreaves:
Removes all CLI-only error variants from `OxenError`. `OxenError` is the `liboxen` unifying
error type. It should never have any coupling with either `oxen-cli` nor `oxen-server.
All CLI-specific variants have been deleted.

Instead, all `oxen` programs -- via the `RunCmd` trait -- now use `anyhow::Error` as the
error return type. `anyhow` is intended for use in applications, so it fits the CLI.
`anyhow` has been added to `oxen-cli`'s dependencies.

Most of the CLI errors are only needed for their string formatting: there is never another
caller that uses most of the errors (they're mostly `OxenError::Basic` variants anyways).
These isolated string errors have been changed to use `anyhow::Error` directly.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -21,6 +21,7 @@ otel = ["liboxen/otel"]
 production = ["metrics", "otel", "default"]
 
 [dependencies]
+anyhow = { workspace = true }
 async-trait = { workspace = true }
 bytesize = { workspace = true }
 chrono = { workspace = true }
--- a/crates/cli/src/cmd/download.rs
+++ b/crates/cli/src/cmd/download.rs
@@ -5,7 +5,6 @@ use clap::{Arg, Command};
 
 use liboxen::api;
 use liboxen::constants::{DEFAULT_BRANCH_NAME, DEFAULT_HOST, DEFAULT_SCHEME};
-use liboxen::error::OxenError;
 use std::path::PathBuf;
 
 use liboxen::repositories;
@@ -59,14 +58,14 @@ impl RunCmd for DownloadCmd {
         )
     }
 
-    async fn run(&self, args: &clap::ArgMatches) -> Result<(), OxenError> {
+    async fn run(&self, args: &clap::ArgMatches) -> Result<(), anyhow::Error> {
         // Parse args
         let id = args
             .get_one::<String>("ID")
             .expect("Must supply a repository id");
         // Check that the id format is namespace/repo-name
         if id.chars().filter(|&c| c == '/').count() != 1 {
-            return Err(OxenError::basic_str(
+            return Err(anyhow::anyhow!(
                 "Invalid repository ID format. Must be namespace/repo-name",
             ));
         }
@@ -76,7 +75,7 @@ impl RunCmd for DownloadCmd {
             .map(PathBuf::from)
             .collect();
         if paths.is_empty() {
-            return Err(OxenError::basic_str("Must supply a path to download."));
+            return Err(anyhow::anyhow!("Must supply a path to download."));
         }
         let dst = args
             .get_one::<String>("output")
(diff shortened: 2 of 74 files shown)


crates/cli/Cargo.toml
```diff
@@ -21,13 +21,15 @@ otel = ["liboxen/otel"]
...
 [dependencies]
+anyhow = { workspace = true }
 async-trait = { workspace = true }
 bytesize = { workspace = true }
 chrono = { workspace = true }
 clap = { workspace = true }
 colored = { workspace = true }
 dialoguer = { workspace = true }
 glob = { workspace = true }
+lazy_static = { workspace = true }
```
CleanCut: This doesn't appear to be used anywhere in this PR.

crates/cli/src/cmd/schemas/add.rs
```diff
@@ -175,9 +175,7 @@ impl SchemasAddCmd {
 
             Ok(serde_json::to_string(&json)?)
         } else {
-            Err(OxenError::basic_str(format!(
-                "Invalid render type: {render_type}"
-            )))
+            Err(anyhow::format_err!("Invalid render type: {render_type}"))
```
CleanCut: Note: Anywhere we manually do `anyhow::format_err!` or `anyhow::Error::msg()` to create a new error, we could use `anyhow!(...)` instead.

crates/cli/src/cmd/schemas/add.rs
```diff
@@ -143,13 +143,13 @@ impl SchemasAddCmd {
...
         metadata: impl AsRef<str>,
-    ) -> Result<(), OxenError> {
+    ) -> Result<(), anyhow::Error> {
         let metadata: serde_json::Value = serde_json::from_str(metadata.as_ref()).map_err(|e| {
-            OxenError::basic_str(format!(
+            anyhow::format_err!(
                 "Metadata must be valid JSON: '{}'\n{}",
                 metadata.as_ref(),
                 e
-            ))
```
CleanCut: Note: Anywhere we manually interpolate an error into an anyhow error, we could use anyhow's `context` or `with_context` methods instead.

CleanCut (approved):
> Removes all CLI-only error variants from OxenError. 

It doesn't look like liboxen's `error.rs` was touched at all. I think this part of the cleanup was missed.

crates/cli/Cargo.toml
CleanCut: Looks like this was fixed between the time I wrote the comment and the time I submitted the review.

crates/cli/src/cmd/download.rs
```diff
@@ -59,14 +58,14 @@ impl RunCmd for DownloadCmd {
...
         // Parse args
         let id = args
             .get_one::<String>("ID")
             .expect("Must supply a repository id");
         // Check that the id format is namespace/repo-name
         if id.chars().filter(|&c| c == '/').count() != 1 {
-            return Err(OxenError::basic_str(
+            return Err(anyhow::Error::msg(
                 "Invalid repository ID format. Must be namespace/repo-name",
             ));
```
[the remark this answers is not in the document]
malcolmgreaves: This is a refactor, keeping existing behavior as-is.

crates/cli/src/cmd/schemas/add.rs
malcolmgreaves: Thanks for the suggestion, I'll change to use this!

crates/cli/src/cmd/schemas/add.rs
malcolmgreaves: Changing to use the macro instead, thanks for the suggestion!

PR #595, opened 2026-05-29
Tolerate invalid JSON in JSON-typed columns on data frame export

jcelliott:
Workspace data-frame export built `SELECT * EXCLUDE (<oxen cols>)` and wrapped it in `COPY ... (FORMAT JSON)`. Legacy rows can hold raw non-JSON text in columns DuckDB typed as JSON, so the re-serialize re-parses them and aborts the whole commit with "Malformed JSON".

Replace the `SELECT *` with an explicit projection (`build_export_projection`) that wraps JSON/JSON[] columns in a `json_valid`-guarded CASE, preserving a corrupt value as a JSON string instead of failing. Valid data is unchanged; csv/tsv/parquet paths are unaffected.

### DIFF
--- a/crates/lib/src/core/v_latest/workspaces/data_frames.rs
+++ b/crates/lib/src/core/v_latest/workspaces/data_frames.rs
@@ -339,8 +339,8 @@ pub fn extract_file_node_to_working_dir(
             let res = conn.execute(&delete.to_string(), [])?;
             log::debug!("delete query result is: {res:?}");
 
-            let excluded_cols = get_existing_excluded_columns(conn, TABLE_NAME)?;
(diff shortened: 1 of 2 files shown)


CleanCut (approved):
This _looks_ like it should work. This is my first time trying to wrap my mind around this DuckDB concept.

No tests seem to contain a JSON list. Is that a test case we should cover?

gschoeni (approved):
LGTM - let's test some migrations on dev

jcelliott:
> No tests seem to contain a JSON list. Is that a test case we should cover?

Good call. I'll add one in a follow-up.

PR #605, opened 2026-06-04
S3 (16c): Migrate DuckDB tabular read path off get_version_path

CleanCut:
Workspace data-frame indexing was the last tabular read still routed through `get_version_path`, whose S3 implementation buffers the whole object in memory before writing it to a file. This PR eliminates the intermediate buffering of the entire file in memory to avoid causing OOMs, allow dealing with larger files, and improve performance. We switch to using`copy_version_to_path` when using the S3 version store to stream the temp file.

Future note: If we find a solution other than DuckDB that doesn't require local files, and could deal with cloud locations, we could avoid creating the temporary file.

Other improvements/fixes along the way:
- (Bug fix) Create the S3 temp file on the oxen-server data volume, since `/tmp` may be tiny in production.
- (Performance fix) Run the DuckDB indexing inside `spawn_blocking` which moves it to a dedicated thread that doesn't block the async runtime that's servicing this request.
- (Cleanup) Drop the now-unused `bucket` field from `VersionLocation::S3`.
- (Bug fix) Flush the S3 `copy_version_to_path` `BufWriter` before returning. Its `Drop` does not flush.
- (Testing) Add `LocalRepository::new_for_testing` to swap in a caller-supplied `VersionStore`, plus an end-to-end test that indexes a workspace data frame whose version file lives only in S3.

ENG-1084

### DIFF
--- a/crates/lib/src/core/v_latest/workspaces/data_frames.rs
+++ b/crates/lib/src/core/v_latest/workspaces/data_frames.rs
@@ -1,3 +1,4 @@
+use async_tempfile::TempFile;
 use duckdb::Connection;
 
 use crate::constants::{DIFF_HASH_COL, DIFF_STATUS_COL, EXCLUDE_OXEN_COLS, TABLE_NAME};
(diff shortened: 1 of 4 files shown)


CleanCut:
@coderabbit review

subygan (approved):

PR #606, opened 2026-06-05
Fix refs manager early eviction

subygan:
When doing bulk "Sync from OxenServer" in the order of 300 repositories. the refs db gets corrupted, because the LRU evicts the open refs DB leading to a corrupted state.

This PR, splits the cache into a Weak-ref live map (uniqueness) and strong-ref warm LRU (performance), so evicted-but-still-in-use handles are reused instead of reopened as corrupting duplicates.

### DIFF
--- a/crates/lib/src/core/refs/ref_manager.rs
+++ b/crates/lib/src/core/refs/ref_manager.rs
@@ -4,10 +4,11 @@
 //! Per-operation `spawn_blocking` callers are fine because the guard lifetime is bounded
 //! by the closure.
 
+use std::collections::HashMap;
 use std::num::NonZeroUsize;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str;
-use std::sync::{Arc, LazyLock};
+use std::sync::{Arc, LazyLock, Weak};
 
 use lru::LruCache;
 use parking_lot::{Mutex, RwLock};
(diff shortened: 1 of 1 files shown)


CleanCut (commented):
> [snip] This matters because RocksDB's on-disk `LOCK` is a
> *per-process* fcntl lock: a second `DB::open` on the same path from within
> this same process is NOT rejected, so you end up with two independent
> handles whose background compaction threads race — one deletes SST files
> the other's MANIFEST still references, corrupting the DB

Nice digging into this! If you're seeing corruption, then there's definitely a bug here. I went down the RocksDB rabbit hole on this one though, andI think what is described in the comment above might not be quite what's happening.

It turns out RocksDB already knows fcntl locks are per-process, so it keeps an in-process hash map of exact file paths and [rejects a second `DB::open` on the same path](https://github.com/facebook/rocksdb/blob/2d68f304259d9271188033bf5faa7fd83e38552b/env/fs_posix.cc#L770-L791) before it even hits fcntl. So a same-path double open should fail cleanly rather than give us two live handles.

If that's right, the corruption probably comes from a different angle. It's a known problem that we've got some places passing relative paths and some passing absolute (or otherwise differently-spelled) paths for the same repo, so RocksDB's exact-path map would see different paths to the same file as two different files and let both opens through.

The thing is, our new `live` map keys on the exact path too, so it'd miss that same case. So I think the most robust fix is to back up and attack the spelling divergence at the root. Normalize the path everywhere before it reaches the cache (and RocksDB). Running it through `util::fs::canonicalize` seems like it'd do the trick. What do you think?

CleanCut:
@subygan and I chatted on Slack and it sounds like we may have been chasing a different problem than we thought. The errors Surya's seeing are the refs db trying to open a file under `/var/oxen/...`, a directory that doesn't actually exist, so it points to a separate root cause. Sounds like we can probably close this one out. Thanks for looking into this! If RocksDB was relying solely on an fcntl lock, this would have fixed it.

<details><summary>Sample error</summary>

`[2m2026-06-04T13:19:20.655065Z[0m [31mERROR[0m [1mHTTP request[0m[1m{[0m[3mhttp.method[0m[2m=[0mGET [3mhttp.route[0m[2m=[0m/api/repos/{namespace}/{repo_name}/branches [3mhttp.flavor[0m[2m=[0m1.1 [3mhttp.scheme[0m[2m=[0mhttp [3mhttp.host[0m[2m=[0mip-10-0-0-4.us-west-1.compute.internal [3mhttp.client_ip[0m[2m=[0m10.0.11.249 [3mhttp.user_agent[0m[2m=[0mreq/0.5.15 [3mhttp.target[0m[2m=[0m/api/repos/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/branches [3motel.name[0m[2m=[0mGET /api/repos/{namespace}/{repo_name}/branches [3motel.kind[0m[2m=[0m"server" [3mrequest_id[0m[2m=[0m393c848a-bd60-42bb-90d5-0d5045a27d4f [3mexception.message[0m[2m=[0mFailed to open refs database "/var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs": Corruption: Corruption: IO error: No such file or directory: While open a file for random read: /var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs/000120.sst: No such file or directory  The file /var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs/MANIFEST-000125 may be corrupted. [3mexception.details[0m[2m=[0mInternalOxenError(RefsDbOpenFailed { path: "/var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs", source: Error { message: "Corruption: Corruption: IO error: No such file or directory: While open a file for random read: /var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs/000120.sst: No such file or directory  The file /var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs/MANIFEST-000125 may be corrupted." } }) [3mhttp.status_code[0m[2m=[0m500 [3motel.status_code[0m[2m=[0m"ERROR"[1m}[0m[2m:[0m [2mtracing_actix_web::middleware[0m[2m:[0m Error encountered while processing the incoming HTTP request: InternalOxenError(RefsDbOpenFailed { path: "/var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs", source: Error { message: "Corruption: Corruption: IO error: No such file or directory: While open a file for random read: /var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs/000120.sst: No such file or directory  The file /var/oxen/data/8d3730d1-632b-4cba-8b46-86ea227203f6/76d6f681-f317-4f4b-acea-d3876da395a3/.oxen/refs/MANIFEST-000125 may be corrupted." } })`
</details>

subygan:
the problem was that there was a difference between the refsDB in memory and, the refsDB we have on disk after the swap. 
this was only meant as a one time exercise, so instead of adding complex tools to tackle this problem. I ended up restarting the dev server to flush the refsDB cache. and reload the refDB in memory

closing this in favor of that.

PR #613, opened 2026-06-07
  Streamline Tests (6): Delete redundant merge tests

CleanCut:
(Stacked on #611)  

This PR deletes some `merge` tests that are completely redundant.

Measurements
- 12 of 55 tests deleted (21.8%)
- Test count: 55 → 43
- nextest wall-clock: 5.870s → 1.775s (−70%)
- Cumulative CPU: ~89.8s → ~21s (~−77%)


Eric-Laurence (approved):
Looks good

Eric-Laurence:
I should do something like this for the hub

crates/lib/src/repositories/merge.rs
```diff
@@ -1497,111 +1292,6 @@ mod tests {
...
-            let content = util::fs::read_from_path(&hello_file)?;
-            assert_eq!(content, "Hello");
-
-            Ok(())
-        })
-        .await
-    }
-
-    #[tokio::test]
-    async fn test_merge_into_base_ff_with_deletion() -> Result<(), OxenError> {
```
subygan, line 1535: We have had bugs in these case before. although it seems logically be the same, with the other tests. it might be worth it to leave these.

subygan (approved):
merge is one of the pieces where bugs pop up quite often.
a few tests can be left as such, even if they seem overlapping.

crates/lib/src/repositories/merge.rs
CleanCut: Acknowledged! ✅ Restored.

PR #624, opened 2026-06-12
Prevent workspace commits from corrupting indexed data frames

gschoeni:
## Problem

Users on the Hub were hitting **"File node does not have metadata"** when opening `oxen/outputs.jsonl` — the file node was committed to main with `data_type: Tabular` but `metadata: None`, permanently breaking every read of the file.

## Root cause

Workspace commit (`core/v_latest/workspaces/commit.rs`) re-exports each *indexed* tabular file from its DuckDB table and re-parses it for metadata. Three gaps:

1. **Silent metadata loss** — `get_file_metadata` swallows tabular parse errors (`Ok(None)`), and `compute_staged_merkle_tree_node` had no guard, so an unparseable export (an empty jsonl, a torn read) was committed as a Tabular node with no metadata.
2. **Non-atomic export** — `extract_file_node_to_working_dir` ran DuckDB `COPY` directly onto the working path, truncating it in place while a concurrent commit may be hashing/parsing the same file.
3. **No commit serialization** — nothing locked per workspace, so two concurrent commits could interleave (one wipes the staged DB mid-flight of the other, or rewrites the export it is reading).

The common trigger: the Hub's media-history bootstrap left the indexed `outputs.jsonl` with only a staged-removed seed row; any workspace commit in that window (e.g. a context upload) exported **zero rows → 0-byte jsonl → no schema → metadata None → corrupted commit**. Hub-side companion fix: Oxen-AI/OxenHub#1299.

## Fix

- `compute_staged_merkle_tree_node` now **fails the commit** when a Tabular file's export can't be parsed, instead of committing a metadata-less node. The last good version stays readable.
- The DuckDB export writes to a temp file in the same directory and **atomically renames** into place.
- Commits are serialized through a **per-workspace async lock** (registry keyed by workspace repo path; entries dropped when no holder/waiter remains). Different workspaces still commit concurrently.
- **Restores the single-shot `POST /workspaces/{id}/files/{path}` upload route** removed in #608. It was the only workspace upload path that hashes content server-side; removing it broke every OxenHub staging flow (and any older CLI/SDK client), since the remaining two-step `/versions` protocol requires clients to compute oxen's xxh3-128 hash themselves. The Hub deliberately does not re-implement the versioning core's hash algorithm, so the route stays until the consolidated protocol grows a server-hashed upload path.

## Reproduce the old behavior

On `main`, this commits a corrupt node; on this branch the commit errors and the file stays readable:

```rust
// commit a 1-row data.jsonl, index it in a workspace, delete the only row, commit the workspace
cargo test -p liboxen --lib test_commit_all_rows_deleted_jsonl_fails_instead_of_corrupting
```

On `main` the workspace commit "succeeds" and subsequent reads fail with "File node does not have metadata". With this branch the commit returns an error and the previous commit's metadata is intact.

## Test

```bash
cargo test -p liboxen --lib repositories::workspaces::data_frames   # incl. new empty-export test
cargo test -p liboxen --lib core::v_latest::workspaces::commit      # lock registry + concurrent-commit tests
# server-dependent suites (start `oxen-server start --test` first):
cargo nextest run -p liboxen --lib --profile ci -E 'test(api::client::workspaces)'
```

Verified locally: 91/91 `api::client::workspaces` and 41/41 workspace repo/core tests pass against a server built from this branch — identical to a clean-main baseline server, including `test_fully_concurrent_workspace_operations` (20 parallel workspace commits over HTTP). The restored upload route is exercised end-to-end by OxenHub's integration suite (1064 tests) against this branch's server binary.

### DIFF
--- a/crates/liboxen/src/core/v_latest/workspaces/commit.rs
+++ b/crates/liboxen/src/core/v_latest/workspaces/commit.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
+use std::sync::{Arc, LazyLock, Mutex as StdMutex};
 use tokio::fs::File;
 use tokio::io::BufReader;
+use tokio::sync::Mutex as TokioMutex;
 
 use crate::constants::STAGED_DIR;
 use crate::core;
@@ -25,12 +27,59 @@ use crate::view::merge::{MergeConflictFile, Mergeable};
 use filetime::FileTime;
 use indicatif::ProgressBar;
 
+// Serializes commits per workspace. A workspace commit reads the staged db,
+// exports indexed data frames into the working dir, writes the commit, then
+// wipes the staged db — none of it under a lock. Two concurrent commits of
+// the same workspace can therefore tear each other's data-frame export
+// mid-read or wipe staged entries out from under the other commit. Keyed by
+// the workspace repo path (unique per workspace); entries are dropped once no
+// commit holds or waits on the lock.
+static COMMIT_LOCKS: LazyLock<StdMutex<HashMap<PathBuf, Arc<TokioMutex<()>>>>> =
+    LazyLock::new(|| StdMutex::new(HashMap::new()));
+
+fn commit_lock_for(key: &Path) -> Arc<TokioMutex<()>> {
+    let mut locks = COMMIT_LOCKS
+        .lock()
+        .expect("workspace commit lock registry poisoned");
+    locks.entry(key.to_path_buf()).or_default().clone()
+}
+
+fn cleanup_commit_lock(key: &Path) {
+    let mut locks = COMMIT_LOCKS
+        .lock()
+        .expect("workspace commit lock registry poisoned");
+    if let Some(lock) = locks.get(key) {
+        // The registry's reference is the only one left — no holder, no
+        // waiter — so the entry can be dropped. A late-arriving committer
+        // simply gets a fresh mutex, which is equivalent since nobody holds
+        // this one.
+        if Arc::strong_count(lock) == 1 {
+            locks.remove(key);
+        }
+    }
+}
+
 pub async fn commit(
     workspace: &Workspace,
     new_commit: &NewCommitBody,
     branch_name: impl AsRef<str>,
 ) -> Result<Commit, OxenError> {
-    let branch_name = branch_name.as_ref();
+    let lock_key = workspace.workspace_repo.path.clone();
+    let lock = commit_lock_for(&lock_key);
+    let result = {
+        let _guard = lock.lock().await;
+        commit_inner(workspace, new_commit, branch_name.as_ref()).await
+    };
+    drop(lock);
+    cleanup_commit_lock(&lock_key);
+    result
+}
+
+async fn commit_inner(
+    workspace: &Workspace,
+    new_commit: &NewCommitBody,
+    branch_name: &str,
+) -> Result<Commit, OxenError> {
     let repo = &workspace.base_repo;
     let commit = &workspace.commit;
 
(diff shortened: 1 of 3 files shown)


gschoeni:
Hub-side companion fix (removes the bootstrap window that triggered this in production): Oxen-AI/OxenHub#1299

gschoeni:
Three server-side findings from reproducing a Hub "Failed to create history dataframe" error (4 concurrent image generations on a fresh repo) against a server built from this branch:

1. **#608 removed `POST /workspaces/{id}/files/{path}`, which OxenHub used for every workspace upload.** The Hub is migrated to the two-step `/versions` protocol in Oxen-AI/OxenHub#1299, but any older external client calling the route (directly or via the hub proxy) breaks on server upgrade. Worth deciding whether a compat route should ship for one release cycle.

2. **Concurrent `GET /commits/history/{branch}` requests 500 on a RocksDB lock**: `IO error: lock hold by current process ... .oxen/cache/commit_count/LOCK: No locks available`. Four parallel requests to the same fresh repo reliably trigger it; clients only survive via retries. The commit_count cache DB open isn't safe under request concurrency — separate fix needed.

3. **`PUT /workspaces` (get_or_create) echoes the *requested* workspace id when it returns an already-existing named workspace** (`controllers/workspaces.rs` builds `WorkspaceResponse { id: workspace_id, .. }` from the request, not from the found workspace). A caller that lost a name race gets a 200 whose `id` doesn't exist on the server. Returning the actual workspace id would let clients detect and adopt the winner.

None of these are addressed in this PR — flagging for follow-up.

CleanCut:
This is obviously WIP, so I won't do a review at this point, but I had a couple thoughts to share on some of the items in the comments:

> 2\. **Non-atomic export** — `extract_file_node_to_working_dir` ran DuckDB `COPY` directly onto the working path, truncating it in place while a concurrent commit may be hashing/parsing the same file.

> 3\. **No commit serialization** — nothing locked per workspace, so two concurrent commits could interleave (one wipes the staged DB mid-flight of the other, or rewrites the export it is reading).

> 2\. **Concurrent `GET /commits/history/{branch}` requests 500 on a RocksDB lock**: `IO error: lock hold by current process ... .oxen/cache/commit_count/LOCK: No locks available`. Four parallel requests to the same fresh repo reliably trigger it; clients only survive via retries. The commit_count cache DB open isn't safe under request concurrency — separate fix needed.

It's worth noting that our design doesn't fully handle concurrent operations on the same repo just yet. In practice, concurrent operations sometimes work out fine, other times we'll error out safely, and in some cases they can cause problems. I've been rounding up these issues in the Repository Integrity project and working through them systematically.  We've made some good progress, but there's still a fair bit of road ahead. I've got it on my list to go through every server endpoint and make concurrent operations safe.

> 1. **[S3 (17a): Cleanup - Consolidate workspace upload paths #608](https://github.com/Oxen-AI/Oxen/pull/608) removed `POST /workspaces/{id}/files/{path}`, which OxenHub used for every workspace upload.** The Hub is migrated to the two-step `/versions` protocol in [Don't stage seed-row removal on v2 media-history bootstrap OxenHub#1299](https://github.com/Oxen-AI/OxenHub/pull/1299), but any older external client calling the route (directly or via the hub proxy) breaks on server upgrade. Worth deciding whether a compat route should ship for one release cycle.

Merging in at least [S3 (17a2): Restore the multipart workspace file-staging endpoint](https://github.com/Oxen-AI/Oxen/pull/619#top) into this PR should help with this.

> 3\. **`PUT /workspaces` (get_or_create) echoes the _requested_ workspace id when it returns an already-existing named workspace** (`controllers/workspaces.rs` builds `WorkspaceResponse { id: workspace_id, .. }` from the request, not from the found workspace). A caller that lost a name race gets a 200 whose `id` doesn't exist on the server. Returning the actual workspace id would let clients detect and adopt the winner.

That's definitely worth fixing now. Mixing up workspace IDs is a great way to get two things stomping on the same workspace when they think they're dealing with separate ones.

gschoeni:
Pushed `8187641` restoring the single-shot `POST /workspaces/{id}/files/{path}` upload route (clean reverse-apply of #608's server-side deletion; the client/python parts of #608 are untouched). This resolves finding 1 from the earlier comment — OxenHub and older external clients work against this branch without re-implementing oxen's content hashing client-side. Findings 2 (commit_count RocksDB lock 500s under concurrent requests) and 3 (`PUT /workspaces` echoing the requested id for existing named workspaces) remain open for follow-up.

jcelliott (approved):

CleanCut (approved):
This does the job! 😄  There are a couple additional improvements I'd like to make, but rather than hashing them out here, I captured them in [a follow-up PR](https://github.com/Oxen-AI/Oxen/pull/644).

PR #640, opened 2026-06-16
S3 (20): Migrate video thumbnail generation off get_version_path

CleanCut:
(Stacked on #639)

Stop using `get_version_path` on the server-side video thumbnail path. This time, we can use `materialize`, since the consumer needs a file on disk.

Part of ENG-1090.


subygan:
This is a good, I've been seeing a lot of these during playground export.

subygan (approved):

PR #646, opened 2026-06-17
LMDB (1): Reusable low-level LMDB layer

CleanCut:
Add a top-level `lmdb` module as a reusable substrate that any oxen store can build on. This should work well for our upcoming merkle tree refactor, as well as in places that we currently use RocksDB.

The next PR will add a trait that will make using this layer easier and more consistent.

Benefits:

- Stores get genuine multi-key ACID transactions and consistent MVCC read snapshots, instead of RocksDB's per-handle synchronization and ad-hoc atomicity workarounds.
- The only public read/write entry points are the `with_read_txn`/`with_write_txn` methods that take a closure and auto commit-on-Ok / abort-on-Err, so a bare transaction can't escape into an `.await` and a commit can't be forgotten.
- One shared layer can replace the multiple, divergent per-store handle caches we currently have
- The new registry enforces at most one live env per canonical path, deduping overlapping opens, and closing on the last  `Arc` drop, without the old tombstone/poison machinery.
- Opening LMDB is much faster than opening RocksDB since it is just an mmap plus a meta-page read (no WAL replay, SST enumeration, or index rebuild) and its hot pages survive the close in the OS page cache, so we no longer need a keep-warm cache.
- Every read is copied out into owned `bytes::Bytes`, so a value safely outlives the short transaction it came from and callers never hold a transaction open just to keep a result.
- Copy-on-write with no WAL, no compaction, and no background threads means no write-stalls or write amplification, stable `O(log n)` read latency. If we crash, that simply leaves the last committed state.

ENG-1127


crates/liboxen/src/lmdb/env_registry.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Path-keyed registry of shared `LmdbEnv` handles. `parking_lot::RwLock` internally (cannot
+/// poison), so there is no lock-poisoned error path. Holds no config — the per-store `map_size`
+/// etc. is passed to `get_or_open`, so one registry (the process-global [`open_shared_env`]) serves
+/// every store rather than needing one registry per store type.
+#[derive(Default)]
+pub struct LmdbEnvRegistry {
+    slots: RwLock<HashMap<PathBuf, Weak<LmdbEnv>>>,
+    /// Serializes opens so two first-opens of the same path cannot race.
+    open_lock: Mutex<()>,
```
subygan, line 42: I'm a little concerned with this global lock for the EnvRegistry.

in the server context, this lock will prevent every other DB open, if one gets stuck, in the loop in `get_or_open`. in this case its 200 ms.

crates/liboxen/src/lmdb/env_registry.rs
```diff
@@ -0,0 +1,274 @@
...
+/// How long `get_or_open` waits out a concurrent close before surfacing `EnvAlreadyOpened`.
+const REOPEN_RETRIES: u32 = 100;
+const REOPEN_RETRY_INTERVAL: Duration = Duration::from_millis(2);
+
+/// Path-keyed registry of shared `LmdbEnv` handles. `parking_lot::RwLock` internally (cannot
+/// poison), so there is no lock-poisoned error path. Holds no config — the per-store `map_size`
+/// etc. is passed to `get_or_open`, so one registry (the process-global [`open_shared_env`]) serves
+/// every store rather than needing one registry per store type.
+#[derive(Default)]
+pub struct LmdbEnvRegistry {
```
subygan, line 39: @CleanCut 

I ran a few benchmarks locally for cold opening a 1 GB lmdb, it was ~200 µs on my mac. warm was ~130 µs. comparing that to, rocksdb at 100ms.

We're already receiving a lot of benefits just by switching to lmdb.
The registry with its retry and caching might become a bottleneck

crates/liboxen/src/lmdb/env_registry.rs
```diff
@@ -0,0 +1,274 @@
+//! One process-wide, path-keyed cache of open `LmdbEnv` handles. It enforces ONE invariant: at
```
subygan, line 1: Can we just change this file name to env_registry.rs ? as this is more of a registry than a cache?

based on our discussion, this design makes sense for the LmDB access pattern.

subygan (approved):

subygan (approved):

crates/liboxen/src/lmdb/env_registry.rs
CleanCut: 500-770 times faster cold open in your measurements! 🎉

crates/liboxen/src/lmdb/env_registry.rs
CleanCut: Good call! 😄

PR #647, opened 2026-06-17
Streamline Tests (10): Speed up api::client::workspaces tests

CleanCut:
Batch 7. Three changes across the `api::client::workspaces` test family:

- Embeddings poll loop tuned from 1s to 10ms at 3 sites. The faster wake-up wastes less time sleeping. The 3 affected tests dropped from ~2.9s to ~1.2s each.
- 19 fixture downgrades to `run_readme_remote_repo_test` (tests that don't read the bounding-box CSV) + 1 to the bounding-box helper from the heavier training-fully-sync-remote.
- 2 duplicate tests deleted in `workspaces/files.rs`.

Results:
- Test count: 91 -> 89
- Slowest embeddings test: 2.993s -> 1.201s (-60%)

### DIFF
--- a/crates/liboxen/src/api/client/workspaces.rs
+++ b/crates/liboxen/src/api/client/workspaces.rs
@@ -457,7 +457,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_remote_commit_staging_behind_main() -> Result<(), OxenError> {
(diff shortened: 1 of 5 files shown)


Eric-Laurence (approved):
Looks good 🔥

PR #648, opened 2026-06-17
S3 (22): Probe the configured S3 bucket at server startup

CleanCut:
The server accepted any S3 bucket name in its storage policy and only discovered a missing, mis-regioned, unreachable bucket, or absent credentials when the first user request hit S3 and returned an error.

Probe the configured bucket at server startup with a live HeadBucket call, which validates the whole S3 configuration end-to-end: the bucket exists, the region and credentials are right, and the server has permission to reach it.

Part of ENG-1147.

### DIFF
--- a/crates/liboxen/src/storage.rs
+++ b/crates/liboxen/src/storage.rs
@@ -3,7 +3,7 @@ pub mod s3;
 pub mod version_store;
 
 pub use local::LocalVersionStore;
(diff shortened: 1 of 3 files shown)


Eric-Laurence (approved):
It looks good to me, but I'm not super comfortable approving these PRs, I feel like I don't fully understand the details. If you want, you could get Surya to take a look at these S3 ones.

PR #655, opened 2026-06-18
S3 (24): Overlap version materialization during workspace batch staging

CleanCut:
This PR updates the code to download up to 16 files from S3 concurrently to each other and to the staging.

Workspace batch staging materialized each file's version blob and then staged it in a serial loop, so a batch paid the sum of every materialization back-to-back. For an S3-backed store each materialization is a full object download, which made large batches bottleneck on serial network round-trips.

The speedup is S3-only. The local path is unaffected since the local store's materialize is a zero-IO path return, so there is no temp file and no behavior change for local-backed repos.

ENG-1136


Eric-Laurence (approved):
Looks good to me.

PR #656, opened 2026-06-18
Fix unindexed workspace data frame pagination

jcelliott:
The endpoint for getting unindexed data frames within a workspace was silently ignoring pagination parameters, so it would return the full data frame on every request. This makes it respect those parameters.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -118,7 +118,7 @@ oxen push origin main               # Push to remote
 
 # Making Changes
 
-- This repository is **public**. Do not mention Oxen's private/internal repositories — by name or description — in code comments, doc-comments, error messages, commit messages, PR titles or descriptions, or any other code or documentation committed here. Keep references to private repos out of public artifacts entirely; if internal context is genuinely needed, point to the relevant Linear issue rather than inlining private-repo details.
+- This repository is **public**. Do not mention Oxen's private/internal repositories — by name or description — in code comments, doc-comments, error messages, commit messages, PR titles or descriptions, or any other code or documentation committed here. Keep references to private repos out of public artifacts entirely; if internal context is genuinely needed, point to the relevant Linear issue rather than inlining private-repo details. This also bars describing a private/internal system's behavior as the justification for a change — even when the system is not named (e.g. "a downstream client loops forever without this"); describe what the code in this repo does and guarantees instead.
 - When changing something that is documented in nearby code, or appears in any markdown files in the repository, update the affected documentation.
 - When prompted to always do something a certain way in general, add an entry to this section of the CLAUDE.md file.
 - When calling `get_staged_db_manager`, follow the doc comment on that function: drop the returned `StagedDBManager` as soon as possible (via a block scope or explicit `drop()`) to avoid holding the shared database handle longer than necessary.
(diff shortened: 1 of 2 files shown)


.claude/CLAUDE.md
```diff
@@ -118,7 +118,7 @@ oxen push origin main               # Push to remote
 
 # Making Changes
 
-- This repository is **public**. Do not mention Oxen's private/internal repositories — by name or description — in code comments, doc-comments, error messages, commit messages, PR titles or descriptions, or any other code or documentation committed here. Keep references to private repos out of public artifacts entirely; if internal context is genuinely needed, point to the relevant Linear issue rather than inlining private-repo details.
+- This repository is **public**. Do not mention Oxen's private/internal repositories — by name or description — in code comments, doc-comments, error messages, commit messages, PR titles or descriptions, or any other code or documentation committed here. Keep references to private repos out of public artifacts entirely; if internal context is genuinely needed, point to the relevant Linear issue rather than inlining private-repo details. This also bars describing a private/internal system's behavior as the justification for a change — even when the system is not named (e.g. "a downstream client loops forever without this"); describe what the code in this repo does and guarantees instead.
```
Eric-Laurence, line 121: Hahaha, this is a very familiar frustration!

crates/oxen-server/src/controllers/workspaces/data_frames.rs
```diff
@@ -123,6 +123,26 @@ pub async fn get(
...
+        // page into a slice before reading (as controllers::data_frames::get does). Without it the
+        // read returns the whole frame regardless of the page. Skip if slice/row is set.
+        if opts.slice_indices().is_none() {
+            // Clamp to the valid 1-based domain: page 0 would derive an empty "0..0" slice (which
+            // panics downstream) and page_size 0 would skip slicing and return the whole frame.
+            let page = opts.page.unwrap_or(constants::DEFAULT_PAGE_NUM).max(1);
+            let page_size = opts
+                .page_size
+                .unwrap_or(constants::DEFAULT_PAGE_SIZE)
+                .max(1);
```
CleanCut: Nit: User input should be validated and corrected once when we retrieve it from the user (lines 110-111), so we don't have to do it every time we use it we use it later on.

crates/oxen-server/src/controllers/workspaces/data_frames.rs
```diff
@@ -123,6 +123,26 @@ pub async fn get(
...
+            // Keep the response's pagination metadata consistent with the values we slice by.
+            opts.page = Some(page);
+            opts.page_size = Some(page_size);
+
+            let start = page_size * (page - 1);
+            let end = page_size * page;
```
CleanCut: Nit: This could cause an overflow. We should use saturating_mul.

CleanCut (approved):
Good improvement. There are some super minor issues, but no need to slow down merging this.

The behavior is different when page size is 0 between the two paths...but that's kind of a nonsensical value any way.

crates/oxen-server/src/controllers/workspaces/data_frames.rs
jcelliott: Fixed

crates/oxen-server/src/controllers/workspaces/data_frames.rs
jcelliott: Fixed

PR #657, opened 2026-06-22
LMDB (2a) - Add MerkleNodeStore trait + file backend

subygan:
Introduce the engine-agnostic seam for Merkle tree node storage, as the first of a stack that will let the node backend switch between the file layout and an embedded KV store.


CleanCut (approved):
Interesting! You've approached this at a lower level (the actual interface between the merkle tree and writing bytes to storage) than I had been thinking of (the interface between the repository and the merkle tree). Your approach has the nice benefit of encapsulating a much smaller design surface, with the tradeoff of introducing an interface where no existing boundary exists today. I think it may well end up being the more pragmatic path for delivering LMDB speedups on a shorter timetable.

We can always revisit the broader merkle tree design at some point down the road when the time feels right.

PR #659, opened 2026-06-22
LMDB (2b) - Rewire MerkleNodeDB to read/write through MerkleNodeStore

subygan:
Replace MerkleNodeDB's direct file IO with delegation to a MerkleNodeStore (the trait + file backend added in the previous change). Writing now accumulates the `node` and `children` blobs in memory and persists them as one unit via `close()` (with `Drop` as a logged fallback); reading pulls the two blobs from the store and runs the existing lookup-table decode unchanged.

### DIFF
--- a/crates/liboxen/src/core/db/merkle_node/merkle_node_db.rs
+++ b/crates/liboxen/src/core/db/merkle_node/merkle_node_db.rs
@@ -46,22 +46,25 @@ For example, data for a vnode of hash 1234 with two children:
     {dir data node}
 */
 
-use std::fs::File;
 use std::io::Read;
 use std::io::Seek;
 use std::io::SeekFrom;
-use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use bytes::{Bytes, BytesMut};
 
 use crate::constants;
 use crate::error::OxenError;
 use crate::model::MerkleHash;
 use crate::model::merkle_tree::node_type::InvalidMerkleTreeNodeType;
-use crate::util;
 
 use crate::model::merkle_tree::node::{
     EMerkleTreeNode, MerkleTreeNode, MerkleTreeNodeType, TMerkleTreeNode,
 };
+
+use super::fs_merkle_node_store::FsMerkleNodeStore;
+use super::merkle_node_store::MerkleNodeStore;
 
 pub(crate) const NODE_FILE: &str = "node";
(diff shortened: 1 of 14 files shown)


crates/liboxen/src/core/db/merkle_node/merkle_node_db.rs
```diff
@@ -522,3 +486,19 @@ impl MerkleNodeDB {
...
+            return;
+        }
+        if let Err(err) = self.flush() {
+            log::error!(
+                "MerkleNodeDB: failed to flush node {} on drop: {err}",
+                self.node_id
+            );
+        }
+    }
+}
```
CleanCut, line 502: The `Drop` implementation might be better off discarding its buffers rather than flushing them. Flushing here would risk persisting a node holding only the children added before the error (a silently truncated node). Discarding is safer:

- In an error case, we'd want to propagate the error, which can only be done by calling `close()`.
- If we forget to call `close()`, we'd want to notice immediately (ideally a failed test) rather than silently persist a partial node.

The same reasoning applies to a future LMDB backend: discarding means we simply never issue the write, rather than committing a truncated node.

CleanCut (approved):
I think there's one change you should probably make. Double-check me! Otherwise, it looks good.

crates/liboxen/src/core/v_latest/commits.rs
```diff
@@ -290,10 +290,12 @@ pub fn create_empty_commit(
     )?;
 
     let parent_id = Some(existing_node.hash);
-    let mut commit_db = MerkleNodeDB::open_read_write(&repo.path, &commit_node, parent_id)?;
+    let mut commit_db =
+        MerkleNodeDB::open_read_write(repo.merkle_node_store(), &commit_node, parent_id)?;
     // There should always be one child, the root directory
     let dir_node = existing_node.children.first().unwrap().dir()?;
     commit_db.add_child(&dir_node)?;
```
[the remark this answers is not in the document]
CleanCut: @subygan It's true that we want to eliminate all `unwrap()` calls in production code, but you don't have to do that in lines you didn't touch in the PR, so these can be ignored for now.

crates/liboxen/src/core/db/merkle_node/merkle_node_db.rs
subygan: This is a good suggestion. I've added it.

PR #660, opened 2026-06-22
LMDB (2c) Make the Merkle node store repo-owned

subygan:
Have LocalRepository own an Arc<dyn MerkleNodeStore>, built once at construction via create_merkle_node_store (mirroring version_store / create_version_store), and hand it to MerkleNodeDB instead of resolving the file backend per-open. MerkleNodeDB::open_read_only/open_read_write now take the store; the static `exists` helper is gone — call sites use repo.merkle_node_store().exists(hash) directly.

No behavior change - the only backend create_merkle_node_store builds is the file store. This is the seam where config/env-driven backend selection (and an alternate engine) will plug in.

### DIFF
--- a/crates/liboxen/src/repositories/commits/commit_writer.rs
+++ b/crates/liboxen/src/repositories/commits/commit_writer.rs
@@ -269,7 +269,7 @@ pub(crate) fn commit_dir_entries_with_parents(
         }
     }
 
-    let mut commit_db = MerkleNodeDB::open_read_write(&repo.path, &node, parent_id)?;
+    let mut commit_db = MerkleNodeDB::open_read_write(repo.merkle_node_store(), &node, parent_id)?;
     write_commit_entries(
         repo,
         commit_id,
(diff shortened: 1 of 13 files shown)


crates/liboxen/src/repositories/commits/commit_writer.rs
```diff
@@ -778,17 +781,22 @@ fn write_commit_entries(
...
+        &mut maybe_dir_db,
         dir_hash_db,
         dir_hashes,
         entries,
         root_path,
         &mut total_written,
     )?;
+    if let Some(mut dir_db) = maybe_dir_db {
+        dir_db.close()?;
+    }
```
CleanCut, line 799: Nit: Nothing ever passes `None` to `r_create_dir_node` for the `dir_db` parameter. It would be cleaner to change `r_create_dir_node` to accept a `&mut MerkleNodeDB` instead of a `&mut Option<MerkleNodeDB>`.

CleanCut (approved):
Looking good!

crates/liboxen/src/repositories/commits/commit_writer.rs
subygan: I created a ticket here,
https://linear.app/oxenai/issue/ENG-1165/remove-optiondir-dband-use-andmut-dir-db
will get to it in a different PR.

PR #665, opened 2026-06-23
Attribute server-side merge commits to the initiating user

CleanCut:
This PR makes it so server merges thread the author through from the caller instead of trying to read the client's local user config, which doesn't exist on the server.

Server-side merges built the merge commit's author from oxen-server's local user config file, which is unset in production. The three-way merge path then failed with `UserConfigNotFound`, surfaced to clients as an opaque 500 . Merging a diverged branch through the Hub's "Merge" button hit this every time.

ENG-1163


Eric-Laurence (approved):
Looks good to me. @jcelliott can we push this through?

jcelliott (approved):

PR #666, opened 2026-06-23
Add some helpful instructions

CleanCut:
(Stacked on #664)

Add an instructions:
- Use `&str` instead of `impl AsRef<str>`
- Remember to update CLI docs
- Stream files' contents rather than buffer whole files in a `Vec`


Eric-Laurence (approved):
Looks good!

PR #668, opened 2026-06-24
Make `oxen info` honor the requested revision via `-r`/`--revision`

CleanCut:
(Stacked on #667)

`oxen info <path> <revision>` parsed a `revision` argument but never used it: metadata (hash, size, type) was computed from the working-tree file on disk and `last_updated` always came from HEAD, so the revision was silently ignored — and `info` returned wrong metadata, or failed outright, for a file whose committed bytes differ from the working tree or that isn't in the working tree at all.

Add `repositories::metadata::get_cli_at_revision`, which resolves the revision to a commit and reads the file's metadata from that commit's merkle-tree `FileNode` (hash, size, data type, mime type, extension, last-updated commit) instead of the on-disk file. `oxen info` routes through it when a revision is given and falls back to the working-tree path otherwise.

Change the revision from a silently-ignored positional argument to a `--revision`/`-r` flag, matching `oxen cat`, so `oxen info <path> -r <rev> --json` reports metadata for the same revision `oxen cat -r <rev>` streams. With no flag, `oxen info` still describes the working-tree file.

Closes ENG-1178

### DIFF
--- a/crates/oxen-cli/src/cmd/info.rs
+++ b/crates/oxen-cli/src/cmd/info.rs
@@ -7,6 +7,7 @@ use liboxen::opts::InfoOpts;
 use liboxen::repositories;
 
 use crate::cmd::RunCmd;
+use crate::helpers::path_relative_to_repo;
 pub const NAME: &str = "info";
 pub struct InfoCmd;
 
@@ -64,11 +65,13 @@ impl RunCmd for InfoCmd {
 
         // Look up from the current dir for .oxen directory
         let repository = LocalRepository::from_current_dir()?;
-        // With a revision, read metadata from that commit's merkle tree; without one, describe
-        // the working-tree file on disk.
+        // With a revision, read metadata from that commit's merkle tree (which is keyed by
+        // repo-relative paths); without one, describe the working-tree file on disk at the path
+        // as given.
         let metadata = match &opts.revision {
             Some(revision) => {
-                repositories::metadata::get_cli_at_revision(&repository, &path, revision)?
+                let repo_path = path_relative_to_repo(&repository, &path)?;
+                repositories::metadata::get_cli_at_revision(&repository, &repo_path, revision)?
             }
             None => repositories::metadata::get_cli(&repository, &path, &path)?,
         };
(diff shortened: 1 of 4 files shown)


crates/oxen-cli/src/cmd/info.rs
```diff
@@ -58,7 +64,14 @@ impl RunCmd for InfoCmd {
 
         // Look up from the current dir for .oxen directory
         let repository = LocalRepository::from_current_dir()?;
-        let metadata = repositories::metadata::get_cli(&repository, &path, &path)?;
+        // With a revision, read metadata from that commit's merkle tree; without one, describe
+        // the working-tree file on disk.
+        let metadata = match &opts.revision {
+            Some(revision) => {
+                repositories::metadata::get_cli_at_revision(&repository, &path, revision)?
```
Eric-Laurence: I think this is intended behavior?

```suggestion
                let current_dir = std::env::current_dir()?;
                let repo_path = util::fs::path_relative_to_dir(current_dir.join(&path), &repository.path)?;
                repositories::metadata::get_cli_at_revision(&repository, &repo_path, revision)?
```

Eric-Laurence (commented):
`oxen info <path> -r <rev>` doesn't resolve the path relative to the repo, so it errors from any subdirectory, but `oxen cat <path> -r <rev>` succeeds, breaking the parity between them

as I understand, you want to keep parity between these?

crates/oxen-cli/src/cmd/info.rs
CleanCut: Nice catch!

I made a little helper function and used it from both the `cat` and `info` commands so we stay consistent.

crates/oxen-cli/src/helpers.rs
```diff
@@ -106,3 +108,9 @@ pub fn check_repo_migration_needed(repo: &LocalRepository) -> Result<(), OxenErr
         "Error: Migration required".to_string().into(),
     ))
 }
+
+/// Resolves a user-supplied path (current-dir-relative or absolute) to a repository-relative path.
+pub fn path_relative_to_repo(repo: &LocalRepository, path: &Path) -> Result<PathBuf, OxenError> {
+    let current_dir = std::env::current_dir()?;
+    util::fs::path_relative_to_dir(current_dir.join(path), &repo.path)
```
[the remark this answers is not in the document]
CleanCut: ✅ Good catch! I've pushed up a fix along with some tests to make sure this works like we want it to.

Eric-Laurence (approved):
Looks good!

CleanCut:
Included in [v0.51.1](https://github.com/Oxen-AI/Oxen/releases/tag/v0.51.1)

PR #670, opened 2026-06-26
lmdb 3(a) Add list_hashes + node_byte_sizes to MerkleNodeStore

subygan:
- `list_hashes()` — enumerate every persisted node, so the whole-tree pack has a single backend-agnostic path to "what's in the store" instead of walking `tree/nodes` itself.
- `node_byte_sizes()` — the `(node, children)` blob lengths without reading the blobs, so the upload-size estimate can size the wire payload without pulling every node into memory.


crates/liboxen/src/core/db/merkle_node/fs_merkle_node_store.rs
```diff
@@ -55,6 +67,83 @@ impl MerkleNodeStore for FsMerkleNodeStore {
...
+            Ok(entries) => entries,
+            Err(e) if e.kind() == ErrorKind::NotFound => return Ok(Vec::new()),
+            Err(e) => return Err(MerkleDbError::Io(e)),
+        };
+
+        // `node_db_prefix` splits the hex hash into `{prefix}` (first 3 chars) and `{suffix}`
+        // (the rest), so a node normally lives at `{prefix}/{suffix}/{node,children}`. When the
+        // hex hash is 3 chars or shorter the suffix is empty and the `node`/`children` files sit
+        // directly under the prefix dir, so a prefix dir may hold both: subdirs (each a suffix)
+        // and the leaf files of the short-hash node sharing that prefix.
```
CleanCut, line 105: All hashes _ought_ to be the same length...so hopefully this is extra defensiveness that we don't need. Either way, the potential problem will all go away under LMDB, so 🤷🏻‍♂️

CleanCut (approved):
⚠️ The CodeRabbitAI finding is valid and should be fixed.

PR #671, opened 2026-06-26
lmdb 3(b) Route Merkle transport packing through MerkleNodeStore

subygan:
Pack node bytes for push/pull/clone transport through the store instead of reading the on-disk two-file layout directly, so the wire path no longer assumes a filesystem backend.


CleanCut (approved):
Looks good.

The coderabbit comment about the trailer is correct, but it's only used cosmetically for the progress bar, so whether you want to fix it now is up to you.

PR #674, opened 2026-06-26
lmdb 3(d) `prune` command Route node pruning through MerkleNodeStore

subygan:
This removes the last production caller that opened the nodes directory directly (outside the FS backend itself and the `#[cfg(test)]` byte-compat oracle), so node pruning is now backend-agnostic.


CleanCut (approved):
The coderabbit finding is correct, but pretty low risk. I'd be surprised if _any_ of our customers ever trigger a prune unless we tell them to, and the condition which would affect them is pretty rare. Fixing now is fine. Opening a linear issue to catch the fix later would also be fine.

PR #684, opened 2026-06-27
Branch Comparison (1): Exclude base history from compare commits

Eric-Laurence:
The compare commits endpoint walked from head until it reached the base commit and stopped. That's fine for a linear branch, but when head contains a merge commit the walk follows the merge's other parent back into commits that already live on the base branch, and since it only stops at that one base commit, all of those older commits get listed too. Greg ran into this after merging production back into dev: a request that should show a few commits showed everything back to the merge.

Now the range is a set difference, reachable from head and not from base, so anything already on base is excluded however head reaches it. Same set as `git log base..head`.


crates/liboxen/src/core/v_latest/commits.rs
```diff
@@ -830,6 +830,71 @@ pub fn list_between(
     Ok(results)
 }
 
+/// List commits reachable from `head` but not from `base` (equivalent to git's
+/// `base..head`), ordered so that a commit always appears before its parents.
```
CleanCut: Nit: `always appears before its parents` isn't actually guaranteed. When a parent commit has multiple children, timestamp skew can lead to popping the parent before some of its children (except the first one). That should be super rare given we use UTC timestamps and most people keep clocks synced nowadays, so it's probably not worth enforcing (it's quite a bit of work for a cosmetic problem). Let's just drop the stated promise above in favor of describing what we actually do:


```suggestion
/// List commits reachable from `head` but not from `base` (equivalent to git's
/// `base..head`), ordered newest-first by commit timestamp.
```

crates/liboxen/src/repositories/commits.rs
```diff
@@ -691,6 +694,228 @@ mod tests {
         .await
     }
 
+    // ----- list_between_exclusive: helpers -----
+    //
+    // Octopus merges (>2 parents) and disjoint histories aren't exercised because
+    // Oxen never produces them: `merge` is pairwise, so commits have at most two
+    // parents, and every repo shares a single root. The algorithm still handles
+    // them structurally (it loops over all parents and seeds from both tips).
```
CleanCut: Nit: Two of these claims are stale: repos don't all share one root (`pull` merges unrelated histories, which happens when you do a pull for the first time and the local and remote repos already have commits), and "seeds from both tips" describes the old single-heap walk, not this two-pass one.

```suggestion
    // Octopus merges (>2 parents) aren't exercised because Oxen never produces them:
    // `merge` is pairwise, so commits have at most two parents. Disjoint histories can
    // occur (`pull` merges unrelated histories) but are awkward to build with the test
    // helpers; the algorithm handles them structurally, as a pure set difference
    // between the two tips' reachable sets.
```

CleanCut (approved):
❤️  Nice!

- The rewrite in 7be4a37d1 resolves the big issue with the counter stuff.
- I went down a deep rabbit-hole about correctness under the assumption that our commits used naive, local timestamps and people frequently work in different timezones...only to discover that we always use UTC timestamps. 😂 So that whole worry basically went out the window. Unless someone has a box that doesn't sync time, and it's clock is way off, we can just assume time is monotonic. And in the extremely rare case that timestamps are wonky, it will just be a cosmetic problem.
- I found two extremely minor doc issues.

PR #685, opened 2026-06-27
LMDB (5c) Run both FS backend and LMDB backend in Github Actions

subygan:
enable running all tests with both LMDB and File system


CleanCut (commented):
This nearly doubles our CI time. 😭 I'm not sure this makes sense to do in CI while it's only an optional change. Let's brainstorm/discuss this with the team a bit. Maybe we do this right before we're going to require the upgrade? And then remove the FS tests shortly afterwards? Look at the CI times! 🙀

CleanCut:
Closing as stale. We need another way of enabling the LMDB tests.

PR #687, opened 2026-06-29
Surface pack errors and panics in the streamed tree download

CleanCut:
#673 misses failures with no trailing chunk. Also, panics in the packer are lost entirely with no log and no error. In both cases the client gets a truncated body under the already-sent 200.

This PR binds the worker's `JoinHandle` and has the closure return its `Result`. It also chains a terminal step onto the `ReaderStream` that awaits the handle once the body drains, turning a returned `Err` or a panic captured by the `JoinError` into a logged-and-returned error that terminates the stream abnormally.

Since a streamed response can't convert a mid-stream failure into a non-200 (the status is already sent at the start), we're using the abnormal termination of the body to make the client reject the gzip/chunked decode.

### DIFF
--- a/crates/oxen-server/src/controllers/tree.rs
+++ b/crates/oxen-server/src/controllers/tree.rs
@@ -1,6 +1,6 @@
 use actix_web::{HttpRequest, HttpResponse, web};
 use bytesize::ByteSize;
-use futures_util::stream::StreamExt as _;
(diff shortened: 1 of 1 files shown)


subygan:
Thanks for this PR! @CleanCut . this change makes sense. I'm merging this.

PR #688, opened 2026-06-30
Repo Integrity (1.10): Keep missing-files from hanging on a commit with no files

CleanCut:
`oxen push --missing-files` and `oxen push --revalidate` could hang indefinitely when run against a commit that contains no files. For example, one that removed the last tracked file. The same hang reaches the server's `list_missing_files` endpoint. This PR fixes them so that they return promptly with "nothing missing".


Eric-Laurence (approved):
Nice catch. Looks good!

PR #689, opened 2026-06-30
S3 (23): Require the S3 bucket region in server config

CleanCut:
`S3VersionStore::client()` detected the bucket's region at runtime with a `GetBucketLocation` round-trip. Each store owns its own client cell and the server builds a fresh store per repo open, so every repo's first S3 operation paid a region lookup -- for what is always the same server-wide bucket.

Make the region an explicit config field instead:

- [storage] `s3_region` is required when `"s3"` is in backends, mirroring `s3_bucket`
- `S3Opts` now has an `s3_region` field
- `version_location()` reads the region from the field instead of building a client.


Eric-Laurence (approved):
Looks good! Paired on this one 👍

PR #690, opened 2026-06-30
S3 (24): Overlap version materialization during workspace batch staging

CleanCut:
Optimize S3 downloads during workspace batch staging by doing multiple S3 downloads simultaneously whilst concurrently staging files. The speedup is S3-only, the local path is unaffected.

Also add an instruction to add, migrate, then delete when altering server endpoints.

ENG-1136


Eric-Laurence (approved):
Looks good! Paired on this one 👍

PR #691, opened 2026-06-30
Bump to rust 1.96.1

CleanCut:
A [hotfix](https://blog.rust-lang.org/2026/06/30/Rust-1.96.1/) release that fixes a few cargo issues and one compiler problem.

I also needed to update to the latest version of duckdb, since the latest GitHub Actions Windows Runner image broke something for duckdb C++ builds that duckdb fixed upstream.


Eric-Laurence (approved):
Looks good

PR #692, opened 2026-06-30
Streamline Tests (11): Speed up remote_mode tests

CleanCut:
Batch 8 — 22 fixture downgrades + 1 duplicate deletion across the `repositories::remote_mode` test family, plus deletion of one orphaned broken stub.

- Helper downgrades (16 plain swaps + 6 with body mods)
- Test deletion (1 test)
- Orphan deletion (1 file - `remote_mode/pull.rs`)

Results:
- Test count: 42 -> 41
- nextest wall-clock: 5.240s -> 5.048s (parallel-saturated by the kept training-fixture tests at ~2.4s)
- Per-test wins are uniform on swapped tests (~10-15% per test); the bigger cumulative-CPU savings will show on CI where parallelism is much lower than the local saturation point.


Eric-Laurence (approved):
Looks good! 🔥 🧹

PR #693, opened 2026-06-30
Make workspace data frame indexing atomic and self-healing

jcelliott:
Build the DuckDB table inside a transaction so a failed or interrupted index leaves no partial table behind, and treat a table missing any OXEN_COLS as not indexed so callers rebuild it instead of querying a table the read path cannot bind.

### DIFF
--- a/crates/liboxen/src/core/db/data_frames/df_db.rs
+++ b/crates/liboxen/src/core/db/data_frames/df_db.rs
@@ -312,6 +312,24 @@ pub fn table_exists(conn: &duckdb::Connection, table_name: &str) -> Result<bool,
     Ok(exists)
 }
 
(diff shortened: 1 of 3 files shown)


gschoeni (approved):
LGTM - built and tested locally, but it's hard to reproduce the original bug

Eric-Laurence (approved):
Looks good!

in `core/v_latest/workspaces/diff.rs` we have is_indexed, which should be replaced by the new table_is_fully_indexed.

also, `repositories::workspaces::data_frames::diff` should probably have a check for table_is_fully_indexed.

subygan (approved):

jcelliott:
@coderabbitai review

PR #695, opened 2026-07-01
Repo Integrity (1.11a): Push uploads files committed outside the fetched subtree

CleanCut:
A push from a subtree or shallow (depth-limited) clone computed its upload set by walking only the fetch-time subtree paths and depth. A file committed outside that scope — a new directory added by a subtree clone, or a file deeper than the clone's depth — was silently left out of the upload, so the branch ref advanced to a commit referencing content the remote never received. The gap only surfaced later, when something tried to read that content.

The push now walks the full local tree (root, unlimited depth) when computing what to upload. Subtrees not present locally load empty and are skipped, and objects already on the remote are still pruned, so this adds no work for full clones and only covers the genuinely-new local objects a partial clone created.

ENG-1195. Prerequisite for the server-side reachable-objects check (1.11b, ENG-1184).


Eric-Laurence (approved):
Looks good! Nice fix!

PR #697, opened 2026-07-01
Streamline Tests (12): `cargo nextest` -> `cargo test`

CleanCut:
This PR switches from using `cargo nextest` (an 3rd-party test runner) to `cargo test` for a bunch of reasons:
- speed up tests
- one less cargo tool to install
- one less config file to keep in sync
- less runner-specific quirks to reason about
 
Here's what we did in this PR:

1. Remove all the `nextest` bits, which initially resulted in 60 tests failing
2. Have test builds always build fresh `reqwest` clients instead of sharing one. This dropped failures from 60 to 2..
3. Serialize the two tests that touch the process-wide `DF_DB_INSTANCES`. This dropped failures from 2 to 1.
4. Use unique bucket names with the s3 test stub. This dropped failures from 1 to 0.
5. Cleaned up the output so that `oxen-server` nor test output reaches the terminal--just the test output.
6. Added a new testing entry in `.claude/CLAUDE.md` to help avoid the kinds of test-concurrency problems that we just fixed.

This resulted in some pretty significant speed improments. Cumulative `libtest` time drops 41% by amortizing the ~185ms per-test process startop across a whole crate in one process instead of paying it once for each of the ~1000 tests. Full-run wall-clock (including ramdisk mount, oxen-server startup, cleanup — constant overhead that both runners share) drops ~37% on median:

| Metric                                            | nextest | cargo test |
|---------------------------------------------------|--------:|-----------:|
| Cumulative libtest wall-clock, all crates         |  52.23s |     30.75s |
| Full `bin/test-rust` wall-clock, warm cache (med) |     63s |        40s |

Both wall-clock medians are over three back-to-back same-runner runs on the local 16-core box. **_The speedup on CI should be even larger!_** CI has fewer cores, so the per-test process floor is a bigger fraction of each test's cost, and cold loaders / IO contention amplify it further.


Eric-Laurence (approved):
Incredible speedup! 🎉

crates/liboxen/src/core/db/data_frames/df_db.rs
Eric-Laurence, line 1254: this asserts that the cache is empty but now that the tests share os processes can't this become flaky?

Eric-Laurence (commented):
Actually, have we tested this on Linux / Windows? We should make sure there aren't any regressions there. ~~CI durations actually look slower for Windows I think?~~ Nevermind, I spent some time looking into this, I think it's because the cache was cold. With a warm cache I believe all of these are sped up by cargo test. 🎉

CleanCut:
Hmm, I think you're actually on to something. I re-ran CI several times, and it looks like Linux/Windows are now consistently slower than they were (and macOS is consistently much faster). Very weird. I am testing to see if it was the client-caching changed that made the regression.

CleanCut:
Closing in favor of #724

PR #699, opened 2026-07-03
Cache release and wheel builds with sccache

Eric-Laurence:
CI rebuilds the rust graph and c/c++ from scratch on every release build, which is slow. This uses sccache for Linux and macOS release builsd and wheel builders.

We use the GH actions cache for CI, but it evicts after 7 days and if we used it for releases it would fight our CI for space. We use an s3 bucket here so we don't have to worry about that and so that the cache persists.

The bundled c/c++ stays cached across releases because it doesnt change between them.

~~I ran into some issues dealing with windows so i'll tackle it later.~~

### Windows

I got windows working, it was a real pain compared to linux and macos.

* sccache doesn't pick up the c/c++ compiler automatically on windows like it does on linux
* ffmpeg was getting rebuilt from source on every run
* Windows bulids proc macro dlls slightly differently every time which made sccache think everything using them had changed and then it would reocmpile half of the entire graph

That last one is fixed by saving the built dependencies to s3 and restoring them between runs.

I also had to turn off Defender during builds, since it was slowing things down, but this is already what GitHub does on its own runners and it shouldn't be an issue.

Also, we had a bug when building the windows wheels where maturin errors would be ignored, but I also fixed that.

### Final measurements

| Build | Before | After | Δ |
|---|---|---|---|
| Linux x86_64 release | 27m38s | 17m02s | -38% |
| Linux arm64 release | 27m35s | 19m05s | -31% |
| macOS release (both arches) | 52m52s | 31m31s | -40% |
| Windows release | 53m31s | 30m18s | -43% |
| Linux x86_64 wheels | 55m25s | 7m45s | -86% |
| Linux arm64 wheels | 64m53s | 9m43s | -85% |
| macOS wheels | 104m30s | 13m59s | -87% |

### DIFF
--- a/.github/workflows/build_and_publish_test_pypi.yml
+++ b/.github/workflows/build_and_publish_test_pypi.yml
@@ -109,6 +109,8 @@ jobs:
     needs: [set_env, compute_version]
     uses: ./.github/workflows/build_wheels_macos.yml
     with:
+      AWS_ROLE: ${{ needs.set_env.outputs.aws_role }}
+      AWS_REGION: ${{ needs.set_env.outputs.aws_region }}
       PYTHON_VERSIONS: ${{ needs.set_env.outputs.python_versions }}
       CHECKOUT_REF: ${{ inputs.commit }}
       VERSION_OVERRIDE: ${{ needs.compute_version.outputs.version }}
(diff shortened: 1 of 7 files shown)


CleanCut (approved):
Let's give it a try!

PR #702, opened 2026-07-06
Show Error message on Quota limit reached

subygan:


crates/liboxen/src/api/client/versions.rs
CleanCut, line 354: ⚠️ It looks like this loop converts the new `OxenError::QuotaExceeded` back into an `OxenError::Basic` before it reaches the CLI (also down on line 375).  If that's right, the chunked large-file path never sees the quota error, so the upgrade hint wouldn't show up there and the PR wouldn't really accomplish what it's setting out to do.

crates/liboxen/src/core/v_latest/push.rs
```diff
@@ -526,22 +526,20 @@ async fn push_commits(
...
-        let error_messages: Vec<String> = errors.iter().map(|e| e.to_string()).collect();
-        return Err(OxenError::basic_str(format!(
-            "Failed to push {} commit(s):\n{}",
-            errors.len(),
-            error_messages.join("\n")
-        )));
+    // Bubble one error up unchanged so callers can classify it (e.g. the CLI maps a quota
+    // error to a friendly hint). Every per-commit failure was logged above with its commit id.
+    if let Some(err) = errors.lock().await.pop() {
+        return Err(err);
```
CleanCut, line 542: `errors` contains any number of errors, and we're returning a random one here. If any of the errors are fatal, we should return one of the fatal ones.

crates/liboxen/src/api/client/versions.rs
CleanCut, line 352: ⚠️ This should be exiting immediately for non-retryable errors. Otherwise, we're paying for the full set of retries with exponential backoff and timeout for an error that will continue to be fatal. 

See `crates/liboxen/src/api/client/versions.rs:197` for an example of where we exit immediately for fatal errors.

CleanCut (commented):
This is getting really close! I think a couple of tweaks will get it to the finish line.

CleanCut:
I'm going to close this PR as stale. I've opened [ENG-1368](https://linear.app/oxenai/issue/ENG-1368/surface-a-clear-quota-exceeded-error-in-the-cli-on-pushupload-revive) to track taking another run at implementing this.

PR #703, opened 2026-07-07
Repo Integrity (1.11b): Compute which objects a commit adds over a base, and which are missing

CleanCut:
Add a `liboxen` primitive that walks a commit's tree relative to a base, descending only into subtrees whose hashes differ so cost scales with the changed set, and reports the merkle nodes and version blobs it adds that the repository is missing. The next PR will call this code.

Also add reachable-objects terminology convention to `CLAUDE.md`

ENG-1184.


crates/liboxen/src/repositories/tree.rs
```diff
@@ -1598,6 +1598,223 @@ pub fn print_tree_depth(
...
+        return Ok(MissingReachableObjects {
+            nodes: vec![head.hash()?.to_string()],
+            versions: vec![],
+        });
+    };
+
+    let store = repo.merkle_node_store();
+    let mut missing_nodes = Vec::new();
+    for hash in &added.nodes {
+        if !store.exists(hash)? {
```
Eric-Laurence: This is asking if the merkle node is on disk. However, our existing list_missing for commits requires both for the node to be on disk, _and_ history/id/dir_hashes. Should we be stricter here as well, or is this a design choice?

crates/liboxen/src/repositories/tree.rs
```diff
@@ -1598,6 +1598,223 @@ pub fn print_tree_depth(
...
+        .find_missing_versions(&added_versions)
+        .await?;
+
+    Ok(MissingReachableObjects {
+        nodes: missing_nodes,
+        versions: missing_versions,
+    })
+}
+
+/// The hash of a commit's root directory node, or `None` if the commit node isn't present.
```
Eric-Laurence, line 1705: Well, this is a minor nitpick, but this comment is slightly misleading. Could be slightly clarified:

```suggestion
/// The hash of a commit's root directory node.
/// None if the commit node isn't present, or if it exists but has no Directory child.
```

Eric-Laurence (approved):
This looks good to me!

crates/liboxen/src/repositories/tree.rs
```diff
@@ -1598,6 +1598,223 @@ pub fn print_tree_depth(
...
+    };
+
+    let mut added = AddedObjects::default();
+    collect_added_from_dir(
+        repo,
+        base_root,
+        head_root,
+        &mut added.nodes,
+        &mut added.versions,
+    )?;
```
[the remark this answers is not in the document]
CleanCut: ✅ Good catch! Fixed!

crates/liboxen/src/repositories/tree.rs
```diff
@@ -1598,6 +1598,223 @@ pub fn print_tree_depth(
...
+    for hash in &added.nodes {
+        if !store.exists(hash)? {
+            missing_nodes.push(hash.to_string());
+        }
+    }
+
+    let added_versions: Vec<String> = added.versions.into_iter().collect();
+    let missing_versions = repo
+        .version_store()
+        .find_missing_versions(&added_versions)
```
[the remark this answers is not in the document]
CleanCut: Another excellent find! ✅ Fixed!

crates/liboxen/src/repositories/tree.rs
CleanCut: At this level, we're dealing purely with merkle nodes, so the directory hash index isn't relevant (there's one per-commit, outside the merkle nodes). However, [in 1.11c](https://github.com/Oxen-AI/Oxen/pull/704) where we are dealing with the commit-level, we should definitely check it! I added a check there.

crates/liboxen/src/repositories/tree.rs
CleanCut: Great, thanks!

PR #705, opened 2026-07-07
Hardening (1.1): Centralize DuckDB connection creation

CleanCut:
Route all library-managed DuckDB connection creation through a single `open_duckdb_connection` helper, and forbid `duckdb::Connection::open` elsewhere via clippy so a stray open can't silently bypass the connection configuration that later steps will add. Pure refactor. No behavior change.

ENG-1219


Eric-Laurence (approved):
Looks good!

PR #706, opened 2026-07-07
Fix ENOTEMPTY on NFS when combining downloaded file chunks

CleanCut:
On an NFS-backed working directory (and probably on NTFS on Windows), `oxen clone` / `oxen pull` could fail with `Error: Directory not empty (os error 39)` while reassembling a large file from its downloaded chunks, aborting the whole operation. Retrying, `pull --missing-files`, `restore`, and `fsck` don't recover it: each attempt re-enters the same reassembly step and fails identically, leaving the repository effectively unclonable.

When combining chunks, the chunk files were still held open as chained readers at the moment their directory was removed. On NFS, unlinking a file that a process still has open triggers a rename to a hidden `.nfsXXXX` file rather than a real delete, so the follow-up `remove_dir_all` finds the directory non-empty and fails with `ENOTEMPTY`. This PR drops the readers to close the handles before removing the chunks directory. Local filesystems on linux/macOS unlink an open file cleanly, which is why this only surfaced on NFS.

I'm investigating adding some CI for NFS, specifically, to catch these sorts of issues in the future. I think the test I added covering chunk reassembly and cleanup would have caught this particular problem on NTFS on our existing Windows CI.

### DIFF
--- a/crates/liboxen/src/storage/local.rs
+++ b/crates/liboxen/src/storage/local.rs
@@ -368,6 +368,10 @@ impl VersionStore for LocalVersionStore {
                 .with_hash(expected_hash)
                 .stream(&mut combined)?;
 
+            // Close the chunk file handles before removing their directory. On NFS, unlinking a
+            // still-open file leaves a hidden .nfsXXXX entry that fails the rmdir with ENOTEMPTY.
+            drop(combined);
+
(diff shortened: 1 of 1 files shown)


Eric-Laurence (approved):
Wow, good catch. Looks good!

PR #708, opened 2026-07-07
Fix ENOTEMPTY on NFS in the `oxen download` path

CleanCut:
On an NFS working directory, `oxen download` of a large file could abort with `Error: Directory not empty (os error 39)`, blocking the download. This is the same NFS failure fixed for clone/pull in #706, just in a different subcommand.


Eric-Laurence (approved):
Same fix, looks good

PR #709, opened 2026-07-08
Revert "Cache release and wheel builds with sccache"

CleanCut:
Reverts Oxen-AI/Oxen#699

I'm having trouble with getting the S3 bucket for this created, and it's blocking a deploy. Reverting for now until Eric is back and we can look at it again.


CleanCut:
Never mind. It looks like the reason I couldn't get TF to show a change to add it is that the plan was already applied and the bucket was created.

PR #712, opened 2026-07-09
Bump ffmpeg version

CleanCut:
We're building off of ffmpeg nightly builds, which are only stored for a limited amount of time before being taken down.

This PR bumps to the latest build.

I will be downloading and storing the builds we use somewhere and then in a followup PR I'll point CI to download from our durable storage.


gschoeni (approved):
LGTM

Eric-Laurence (approved):
Looks good 👍

PR #714, opened 2026-07-09
Store version-batch files concurrently, off the event loop

CleanCut:
The `/versions` batch endpoint stored each uploaded file serially, so a batch's wall-clock time was the sum of every file's decompress-and-store. That's tens of seconds for a large batch of small files. It now decompresses and stores a batch's files in parallel in multiple threads so a batch completes much more quickly.

Two failure modes on the batch path were fixed along the way: each part's decompressed size is now capped, so a gzip bomb can't exhaust memory; and a task that panics now fails the request with a 500 instead of returning a success response.

The gzip-inflate-with-cap logic now lives in a shared helper and both places that we were doing that now call the helper.

### DIFF
--- a/crates/oxen-server/src/controllers/versions.rs
+++ b/crates/oxen-server/src/controllers/versions.rs
@@ -6,12 +6,12 @@ use crate::params::{app_data, parse_resource, path_param};
 
 use actix_multipart::Multipart;
 use actix_web::{Error, HttpRequest, HttpResponse, web};
-use async_compression::tokio::bufread::GzipDecoder;
 use async_compression::tokio::write::GzipEncoder;
 use async_zip::Compression;
 use async_zip::base::write::ZipFileWriter;
 use flate2::read::GzDecoder;
 use futures_util::{StreamExt, TryStreamExt as _};
+use liboxen::constants::stream_segment_size;
 use liboxen::error::OxenError;
 use liboxen::model::LocalRepository;
 use liboxen::model::metadata::metadata_image::ImgResize;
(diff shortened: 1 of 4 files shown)


crates/oxen-server/src/controllers/versions.rs
```diff
@@ -544,92 +543,126 @@ pub async fn save_multiparts(
...
-            while let Some(chunk) = field.try_next().await? {
-                field_bytes.extend_from_slice(&chunk);
+        // Bound the in-flight stores (and thus buffered bytes) before dispatching the next one.
+        while in_flight.len() >= max_in_flight {
+            if let Some(result) = in_flight.join_next().await
+                && let Some(err_file) = store_task_result(result)?
+            {
+                err_files.push(err_file);
             }
+        }
```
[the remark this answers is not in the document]
CleanCut: Well, don't set `OXEN_NUM_THREADS=0`. That's not a supported value!

crates/oxen-server/src/controllers/versions.rs
```diff
@@ -544,92 +543,126 @@ pub async fn save_multiparts(
...
+                }
             } else {
-                let cursor = std::io::Cursor::new(field_bytes);
-                Box::new(cursor)
+                field_bytes
             };
 
             match version_store
-                .store_version_from_reader(&upload_filehash, reader, field_size)
+                .store_version(&upload_filehash, data.into())
```
Eric-Laurence, line 623: You swapped from store_version_from_reader to store_version here.

I believe on S3, the former of the two hashes the bytes and rejects a mismatch with the filename hash before PUT, while the latter does not. So I think there is an s3 regression in here.

Eric-Laurence (approved):
Looks good, I have one concern which I left a comment about

crates/oxen-server/src/controllers/versions.rs
CleanCut: Good catch!

Checking the hash in any path to storing a version file is relatively new--we used to just trust the client. I thought I had gone back and added that to all paths, but it looks like I missed some.

I've added an item to my Repository Integrity workstream to circle back and verify that all version store upload methods verify the file hash along the way.

PR #715, opened 2026-07-09
Streamline Tests (12.2): weak-ref cache for changes_db

CleanCut:
Same as #713, but for `changes_db` instead of `workspace_name_index`

### DIFF
--- a/crates/liboxen/src/core/db/data_frames/changes_db.rs
+++ b/crates/liboxen/src/core/db/data_frames/changes_db.rs
@@ -15,68 +15,106 @@
 //! used to `DB::open` the change-tracking database fresh, and the second open
 //! collided with the lock the first still held.
 //!
-//! This module keeps one shared handle per path in a global LRU cache, mirroring
-//! [`crate::core::staged::staged_db_manager`]. Every caller that touches the
-//! same change-tracking database in this process shares that one handle, so they
-//! never race on the file lock.
+//! This module keeps one shared handle per path in a process-wide weak-ref registry.
+//! Every caller for the same path receives the same `Arc<RwLock<DB>>` for as long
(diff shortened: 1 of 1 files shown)


crates/liboxen/src/core/db/data_frames/changes_db.rs
```diff
@@ -15,140 +15,127 @@
...
 
-/// Global LRU cache of open change-tracking RocksDB handles, keyed by db path.
-/// `Mutex` rather than `RwLock` because `LruCache::get` requires `&mut self` to
-/// bump recency, so every access is exclusive anyway.
-static CHANGES_DB_INSTANCES: LazyLock<Mutex<LruCache<PathBuf, Arc<RwLock<DB>>>>> =
-    LazyLock::new(|| Mutex::new(LruCache::new(CHANGES_DB_CACHE_SIZE)));
+/// How long `get_changes_db` waits out a concurrent close before surfacing a LOCK error.
+/// See the retry loop in [`get_changes_db`] for the race this covers.
+const OPEN_RETRIES: u32 = 100;
+const OPEN_RETRY_INTERVAL: Duration = Duration::from_millis(2);
```
Eric-Laurence, line 50: Won't this block for like 200 ms on a lock collision

I know we're switching to LMDB so it probably doesn't matter

Eric-Laurence (approved):
Looks good 👍

crates/liboxen/src/core/db/data_frames/changes_db.rs
CleanCut: Yes, it will block _up-to_ 200ms on a lock collision. The collision should be an edge case, and if it does happen, it will likely be resolved in less than 200ms. And yes, we're switching to LMDB as well, which makes this all go away.

PR #717, opened 2026-07-10
Bump to Rust 1.97.0

CleanCut:
- Bump to Rust 1.97.0
- Fix harmless linker warning by adding config to `.cargo/config.toml` to ignore it
- Bump deep dependency `ethnum` from `1.5.2` to `1.5.3` to resolve undefined behavior which Rust 1.97.0 doesn't allow
- Everything else is just fixes for new clippy lints


Eric-Laurence (approved):
Looks good!

PR #718, opened 2026-07-10
Streamline Tests (12.3): weak-ref cache for ref_manager

CleanCut:
Same as #713, but for `ref_manager`

- Also makes a small tweak to #713 and #715 so that they don't attempt creating directories unless they have to.


Eric-Laurence (approved):
Looks good!

PR #721, opened 2026-07-10
Streamline Tests (12.5): weak-ref cache for dir_hashes_db

CleanCut:
Last of the five RocksDB caches with the same LRU-eviction pathology the earlier weak-ref refactors (#713, #715, #718, and #719). `dir_hashes_db` opens read-only for regular usage, so this one looks slightly different, though still pretty similar.

Callers of `with_dir_hash_db_manager` now always see the same shared handle for the life of any concurrent reader of the same commit. Idle handles close as soon as the last reader drops.

 The exclusive-access helper is renamed from `with_entry_evicted` to `with_exclusive_access`. 

RO opens don't take the per-directory `LOCK` file, because you can have multiple read-only opens of the same file, so the retry loop the earlier refactors carry isn't needed here.


Eric-Laurence (approved):
Looks good

PR #722, opened 2026-07-10
Streamline Tests (12.6): Fix FileNode's PartialEq implementation

CleanCut:
Fix a bug in `FileNode` where `PartialEq` wasn't comparing all the fields that `Hash` used to determine uniqueness.

### DIFF
--- a/crates/liboxen/src/model/merkle_tree/node/file_node.rs
+++ b/crates/liboxen/src/model/merkle_tree/node/file_node.rs
@@ -249,6 +249,8 @@ impl MerkleTreeNodeIdType for FileNode {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Eric-Laurence (approved):
How did you catch this?

CleanCut:
> How did you catch this?

I'm working on converting from `nextest` to `cargo test`. `nextest` hides flakes by retrying 3 times, so I started eliminating flakes. `test_list_tabular_files_in_repo` was one of the flaky tests. This turned out to be the reason it's flaky. It's an actual bug that only surfaced when things happened in a particular random order.

I think it's a good thing to not hide flakes. 😬

PR #723, opened 2026-07-10
Streamline Tests (12.7): Preload DuckDB JSON extension

CleanCut:
## Summary

Eliminates a Windows-only concurrent-startup race where DuckDB's first-use autoload of the JSON extension fails with `Could not move file: Access is denied`. 

`oxen-server` now runs a single `INSTALL json` against an in-memory connection during startup, so the extension is already resident on disk by the time real connections open it — no first-use race is possible.


Eric-Laurence (approved):
Nice fix

PR #726, opened 2026-07-11
Cache Windows release with dependency tree

Eric-Laurence:
Caching already works well on Linux and macOS, but the windows build has been resistant and so far my approaches have failed to make a difference.

Windows releases rebuild the whole dependency graph and the bundled c/c++ from scratch every run. I tried using sccache to fix that, but it gets 0 hits in release builds. The crates that change miss the cache anyway and the rest already comes back from the dep tree cache. So sccache was actually slowing us down.

Further, the dep tree cache wasn't properly working. It hashed the whole Cargo.lock, and every release bumps the workspace crate versions in there, so the cache always missed. Locally I was rerunning against the same commit, which is why it looked like it was working before. Now it only hashes the third party entries, so a version bump keeps the same key. When a dependency does change and the key misses, it falls back to the most recent tree, so cargo still only rebuilds what the bump touched.

Ffmpeg had a similar problem. We cloned vcpkg at HEAD so its binary cache key drifted, and ffmpeg rebuilt from source every run. Pinning vcpkg fixes it.

So to sum it up, this drops sccache on windows, along with all the setup it needed, and leans on the dep tree cache alone. I did some trial runs and it looks like windows release went from 59 to 35 min, but we will have to see how it works out under real usage.

### DIFF
--- a/.github/workflows/release_windows.yml
+++ b/.github/workflows/release_windows.yml
@@ -122,8 +122,8 @@ jobs:
           # from-scratch build recompiles the whole graph every release. Restore
           # the previous run's compiled dependencies (and the registry, mtimes
           # intact) so cargo marks them Fresh and only workspace crates rebuild.
-          # Keyed on the third-party lockfile entries plus toolchain, so a
-          # workspace version bump alone does not rotate the key.
+          # The key hashes the third-party lockfile entries plus toolchain, so a
+          # workspace version bump alone keeps the same key.
           $depText = ((Get-Content Cargo.lock -Raw) -split '\[\[package\]\]' | Where-Object { $_ -match 'source = ' }) -join ''
           $rv = (cmd /c "rustc -V 2>nul") -join ''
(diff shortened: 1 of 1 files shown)


.github/workflows/release_windows.yml
```diff
@@ -227,11 +178,23 @@ jobs:
               if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           }
 
-      - name: sccache stats
-        if: always()
+      - name: Save dependency build tree
+        # Persist even when a build step failed, so a workspace bug doesn't throw
+        # away the dependency compiles this cache exists to reuse.
+        if: ${{ !cancelled() }}
```
CleanCut: "cancelled" and "failed" are different statuses. So `!cancelled()` can cache a failed build that will then produce an exact cache hit the next time around. Then we restore a failed build and have to rebuild.

`if: success()` only ever saves a complete tree. 

```suggestion
        if: success()
```

CleanCut (approved):
I'm hoping this helps. One suggestion.

Update: Oh, that outside-range review comment above would be good to look into as well.

PR #727, opened 2026-07-11
Keep tree node unpacking from stalling the server

Eric-Laurence:
When you push, the server unpacks the commit's tree one node at a time, on the same threads it uses to handle network traffic. A big repo has tens of thousands of nodes, so a single unpack can take minutes and hold one of those threads the whole time, and while it does, that thread can't serve the other connections it's juggling.

That's synchronous filesystem and DB work running on the async workers, which should go in spawn_blocking. This moves it there, so a slow unpack can't tie up the threads that serve requests.

I also made it write nodes in bulk rather than one at a time: the file backend writes a batch in parallel, and the LMDB backend commits a batch in one transaction. That was about 30% faster in local testing, though I don't know how it translates to the server.

### DIFF
--- a/crates/liboxen/src/core/db/merkle_node/merkle_node_store.rs
+++ b/crates/liboxen/src/core/db/merkle_node/merkle_node_store.rs
@@ -112,24 +112,13 @@ pub(crate) trait MerkleNodeStore: Debug + Send + Sync {
     ///
     /// Unpacking a commit's tree writes one node for every directory and vnode. A large repo has
     /// tens of thousands of them, and writing them one at a time means tens of thousands of
-    /// separate trips to the store. Each backend has a faster way to write a whole batch, so this
-    /// hands it the batch and lets it use that way. The filesystem backend writes the files at once
-    /// across threads. The LMDB backend commits the batch in a single transaction. The default is
-    /// the plain serial loop that any backend can fall back on.
+    /// separate trips to the store. Each backend implements batching: the filesystem backend writes
+    /// files in parallel across threads; the LMDB backend commits the batch in a single transaction.
     fn write_nodes(
         &self,
         nodes: Vec<(MerkleHash, Bytes, Bytes)>,
         overwrite_existing: bool,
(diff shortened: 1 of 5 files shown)


crates/liboxen/src/core/db/merkle_node/merkle_node_store.rs
```diff
@@ -107,6 +107,30 @@ pub(crate) trait MerkleNodeStore: Debug + Send + Sync {
...
+        overwrite_existing: bool,
+    ) -> Result<Vec<MerkleHash>, MerkleDbError> {
+        let mut written = Vec::with_capacity(nodes.len());
+        for (hash, node, children) in nodes {
+            if overwrite_existing || !self.exists(&hash)? {
+                self.write_node(&hash, node, children)?;
+                written.push(hash);
+            }
+        }
+        Ok(written)
```
CleanCut: Nit: Given that both implementations of this trait override the default, we can remove the default here. The default implementation only serves as a footgun in the future if we add a third implementation and don't think to check for default methods we could override.

CleanCut (approved):
Thanks! 

Which implementation was 30% faster, LMDB or FS?

Any speedup in the FS implementation is useful at the moment. We'll see a slight slowdown in concurrent operations that also use the rayon pool, but there's only a couple other things that do, currently, so it shouldn't be an issue.

Fun fact: The FS implementation is more bottlenecked by 4 fsyncs per node than anything else (2 fsyncs per atomic file write x 2 files per node), which is probably why you didn't see any dramatic speedups. That adds up to 400k fsyncs per full batch! We need the fsyncs when we just write a single file, because after we exit the method we need to be certain it's all there.

But we _don't_ need the fsyncs in the middle of a batch that we can consider as having succeeded or failed as a whole. So if we were keeping the FS implementation long term, then we would want to dig into reimplementing the file writing in the batch to avoid fsyncs until the very end, and then use some wonky OS-dependent mechanisms for triggering a directory- or filesystem-wide fsync. But this is all peanuts compared to switching to the ~1000 faster LMDB implementation, so it's not worth doing right before migrating.

PR #728, opened 2026-07-12
Fix paths pushed by Windows CLI

Eric-Laurence:
Nested folders on repos pushed from a Windows client returned "Resource not found" in the web browser and the dir API.

The commit writer stores each directory path as a key in a `dir_hashes` lookup, and on Windows it wrote those keys with backslash separators. That db is taken at face value by the server, and a Linux server looking up `a/b/c` never matched the stored `a\b\c`.

Our pre existing tests only run the client and server on the same OS, so they always agree on separators. It would be useful to run more tests cross platform.

Writes now canonicalize keys to forward slashes. I also made the loader heal existing backslash keys on read, so existing repos and old clients with this problem will now work.


CleanCut (approved):
❤️ A clear step forward, thanks!

We'll need to push further in the future. I opened ENG-1249 to ensure we circle back and do a more comprehensive sweep/consolidation/lint-enforcement of all usage of paths-as-keys.

PR #729, opened 2026-07-13
Fix concurrent workspace creation race

Eric-Laurence:
When two requests create a workspace with the same name at once, exactly one should win the name and the rest should get WorkspaceAlreadyExists. Instead they can all fail and no workspace gets created, even though one of them should have succeeded.

The first create for a repo lazily builds the name index from disk, guarded only by a "does the index dir exist yet" check. Under concurrency every caller sees it missing and kicks off its own rebuild on the blocking pool. A rebuild clears the index and rescans the workspace configs on disk, so if one runs just after a caller has written its config but before it claims the name, it reads that config back and inserts the name itself. The caller then goes to claim the name, finds it already there pointing at itself, and rolls back. When that caller was the one that should have won, everyone loses.

This serializes that first build behind a per-repo lock, the same way commit.rs already does. Only the first caller rebuilds, and it runs before any config is on disk, so there's nothing for it to misread.

Ran the concurrent create test 10k times under parallel load: 1494 failures before, 0 after.

### DIFF
--- a/crates/liboxen/src/repositories/workspaces.rs
+++ b/crates/liboxen/src/repositories/workspaces.rs
@@ -359,8 +359,13 @@ async fn ensure_name_index(repo: &LocalRepository) -> Result<(), OxenError> {
                 Err(join_err) => {
                     // The blocking task panicked after `get_index` may have created
                     // the dir; invalidate so a retry rebuilds rather than trusting a
-                    // never-populated index.
-                    invalidate_name_index(repo);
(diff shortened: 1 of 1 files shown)


CleanCut (approved):
Good improvement!

Since my review comments were both rather complicated additions, I went ahead and stacked [a PR](https://github.com/Oxen-AI/Oxen/pull/732#top) instead of writing it out. That PR fixes the leak and adds a test that always triggers on the problem we fixed, since the existing test is so flaky.

PR #733, opened 2026-07-13
Min Version (4.1): split the client-protocol floor out of MIN_OXEN_VERSION

CleanCut:
The version the server gates incoming clients against is now a standalone `MIN_OXEN_CLIENT_VERSION` constant, independent of the repo-format version. The client gate and the /api/min_version endpoint read it directly, so removing the repo-format `MIN_OXEN_VERSION` in a later substep no longer threatens the gate.


Eric-Laurence (approved):
Looks good!

PR #734, opened 2026-07-13
Min Version (4.2): simplify the merkle node types

CleanCut:
Merkle node constructors no longer take a repository, and nodes no longer carry a version() accessor — both were vestiges of the multi-version dispatch that now only ever resolves to the current format. The single-variant node enums are left intact, so the on-disk node format is unchanged.

ENG-1252


Eric-Laurence (approved):
Looks good!

PR #736, opened 2026-07-14
Streamline Tests (12.7b): Fix flaky workspace embedding-index tests

CleanCut:
Three embedding tests waited for background index completion by polling the server in a loop that asserted `result.is_ok()` on every poll and read `columns[0]` directly. A single transient response while the server was busy, or an empty column list observed mid-index, failed or panicked the test even though the code under test was correct. This would fail more often when under heavy load.

Polling now tolerates a transient failure (retrying instead of asserting) and reads the status through columns.first(), so only a genuine failure to finish within the ~10s budget fails the test. The three duplicated loops collapse into one shared wait_for_embedding_index helper.


Eric-Laurence (approved):
Looks good!

PR #737, opened 2026-07-14
Fix NFS/DuckDB bug when deleting/clearing workspaces or deleting a repo

CleanCut:
Fix an `ENOTEMPTY` error that occurrs when deleting a workspace on NFS. We need to close a workspace's DuckDB connections before deleting the workspace.

We found this bug with our new NFS CI job.


Eric-Laurence (approved):
Looks good! Heads up, I think workspaces clear has the same gap.

CleanCut:
Good call! I went searching and found a _third_ place with the same problem as well. I'm pushing up a commit to fix them.

PR #745, opened 2026-07-15
DuckDB Tracking (4/4): Clean up client surfaces after the tracking-column removal

gschoeni:
**Stack:** #749 → #742 → #744 → **#745 (this)**

## Summary

Client-side polish now that the tracking columns are gone:

- **Python `DataFrame`**: removes the `filter_keys` machinery — the legacy tracking columns it hid (`_oxen_diff_hash`, `_oxen_diff_status`, `_oxen_row_id`) no longer exist server-side, and `_oxen_id` remains visible as the row handle (as it always was). Also drops the class docstring example calling a `diff()` method that never existed, clarifies the `restore()` docstring (discard staged edits by re-indexing), and removes a stray debug `print` from `insert_row`.
- **CLI**: fixes the stale `oxen df --delete-row` help text to name the `_oxen_id` column.

### Docs site
The workspace data frame pages on docs.oxen.ai need a follow-up in the docs repo: remove mentions of the staged data frame diff (`oxen workspace diff`) and row/column restore, and note that edits apply directly (whole-data-frame restore still discards all staged edits).

### Testing
- `bin/test-rust -p`: 80/80
- clippy clean

---

## How to review
- Small, client-only diff: the Python `DataFrame` cleanup (filter machinery removal — check `insert_row`/`get_row_by_id` still surface `_oxen_id`), the CLI help-text fix, and the `sort_by` clear in `JsonDataFrameView` (companion to sorting by DuckDB's virtual `rowid`; prevents the view-layer re-sort from panicking on SQL-only pseudo-columns).
- Follow-up owed after merge: the docs.oxen.ai pages for workspace data frames (separate docs repo) — remove staged-diff/row-restore mentions.

## How to merge
Top of the stack — merge last, after #744. Squash + delete branch. Once all four are in, the stack is fully landed and `main` carries: crash fix → reversibility removal → single-column schema → client polish.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #749, opened 2026-07-16
DuckDB Tracking (1/4): Remove the linker flag that turned DuckDB errors into macOS process aborts

gschoeni:
**Stack:** **#749 (this)** → #742 → #744 → #745

## Summary

One-line root-cause fix for every macOS-only `oxen-server` SIGABRT of the form `libc++abi: terminating due to uncaught exception of type duckdb::...`.

`.cargo/config.toml` passed `-Wl,-no_compact_unwind` on `aarch64-apple-darwin` (added to silence compact-unwind linker warnings, believed harmless). On arm64 Mach-O, the C++ objects in our bundled native deps (DuckDB, RocksDB) carry their unwind info as **compact unwind** — stripping it leaves C++ exceptions with no reachable handler, so `__cxa_throw` goes straight to `std::terminate`. Rust panics kept unwinding (rustc emits full eh_frame alongside), which is why only C++ error paths died and the breakage masqueraded as a DuckDB bug.

### This was the root cause of
- The production crash where a workspace data frame's WAL replay aborted the server (`duckdb::InternalException` during `duckdb_open`) — the WAL-recovery path in `get_connection` exists but the process died before it could run
- Server aborts on unbindable SQL (unknown column, unresolved parameter types) reaching `prepare`/`query_arrow`
- The three WAL-corruption recovery tests aborting permanently on macOS while CI (Linux) stayed green

### Second commit: the warning the flag was hiding
With the flag gone the linker reports `__eh_frame section too large (max 16MB) to encode dwarf unwind offsets in compact unwind table` — a performance note (slower dwarf unwind on cold error paths), not a correctness problem. It is deliberately allowed via `[workspace.lints.rust] linker_messages = "allow"`, with comments in both files recording that "fixing" this warning with the linker flag is exactly what caused the aborts.

### Verification
- Minimal harness (duckdb crate only): unknown-column query returns clean `Err` without the flag; SIGABRT exit 134 with it
- `bin/test-rust data_frames` on macOS: **121/121 passed, zero aborts** — first fully green run; previously 3 permanent aborts
- The corrupt-WAL recovery tests now exercise the actual recovery path (open fails cleanly → WAL removed → retry succeeds)
- `cargo build --workspace` emits zero warnings

Note: DuckDB 1.5.4 also has a genuine upstream WAL bug (WALs containing `ALTER TABLE ... ADD COLUMN ... DEFAULT uuid()` cannot be replayed — minimal repro in hand, to be filed upstream). With this fix that failure surfaces as a clean error and WAL recovery handles it, instead of killing the server.

---

## How to review
- This PR is the base of the stack; the diff against `main` is exactly two small commits: the linker-flag removal in `.cargo/config.toml` and the `linker_messages` lint allow in `Cargo.toml`. Read the comments in those two files — they encode the whole story.
- To verify the fix locally (Apple Silicon): check out `main`, run `bin/test-rust data_frames` and watch the three WAL-recovery tests abort; then check out this branch and watch them pass (121/121, zero `libc++abi` lines).

## How to merge (whole stack)
Merge order is bottom-up: **#749 → #742 → #744 → #745**. For each PR in order:
1. Squash-merge and delete the head branch — GitHub then automatically retargets the next PR in the stack onto `main`.
2. After each squash-merge, the next branch still carries the merged commits; rebase it onto `main` (`git rebase --onto main <old-base>`) and force-push so its PR shows only its own diff.
Each PR is independently green (CI + full local suites), so the stack can also pause at any point — nothing below depends on anything above it.


CleanCut (approved):
Shoot, it was my silencing of a linker warning that caused this? 🤦🏻‍♂️ 🤣 😭

PR #753, opened 2026-07-16
Async (7): Tree/entries read layer to async

CleanCut:
The read APIs behind directory listings, file and directory metadata, and repo detail are the highest-fan-out data path in oxen-server, and the handlers that serve those endpoints run their synchronous merkle/RocksDB/LMDB reads inline on the actix worker threads. Each read pins a worker for its full duration, so under load these paths starve new-connection accepts and health checks.

Give `entries::get_directory`, `entries::get_meta_entry`, `entries::list_directory_w_workspace_depth`, and `tree::get_file_by_path` async variants that run their reads off the worker thread, so the handlers can await them instead of blocking the async runtime. The variants land alongside the sync originals so the dir, meta, and repo-detail endpoints can adopt them independently, without a flag-day signature flip across every caller; the sync originals are removed in a later cleanup.

ENG-1283

### DIFF
--- a/crates/liboxen/src/repositories/entries.rs
+++ b/crates/liboxen/src/repositories/entries.rs
@@ -4,7 +4,7 @@
 use crate::core;
 use crate::core::versions::MinOxenVersion;
 use crate::error::OxenError;
-use crate::model::merkle_tree::node::FileNode;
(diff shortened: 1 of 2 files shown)


Eric-Laurence (approved):
👍

PR #754, opened 2026-07-16
Add back in Field.changes and subfields

gschoeni:
Add back in Field.changes, Changes, and PreviousField and updated current Field constructors to set changes: None because we were getting a crash serializing the object from the old format.

The crash was a little misleading:

Error: Err status [500 Internal Server Error] from url [Cannot decode a Merkle node: unknown variant `V0_25_0`, expected one of `Commit`, `File`, `Dir`, `VNode`, `FileChunk`
]

So we pass through a real error now.


CleanCut (approved):
I added a test, so we can take a run at removing this in the future and find a way to do it without breakage.

PR #755, opened 2026-07-16
Async (8): repo-detail (repositories::show) reads to async

CleanCut:
(Stacked on #753)

The repo-detail endpoint (`GET /api/repos/{namespace}/{repo_name}`) is the largest request-time consumer in oxen-server, and it ran all of its merkle/RocksDB/LMDB reads inline on the actix worker thread, which blocked that worker thread for the request's full duration. Under load, enough blocked workers stall other requests and even fresh-connection accepts and health checks. Move every read in the handler off the worker: repo lookup, revision resolution, the root directory node, and the branch list now run inside operation-granularity `spawn_blocking`, so the handler awaits them instead of blocking the async runtime.

I cut two redundant refs-DB reads along the way. The handler resolved the default branch with its own lookup even though the commit it had just resolved already carries that branch's commit id, and it tested emptiness with a separate scan on top of the branch-count scan. Build the branch from the resolved commit, and derive is_empty from the branch count — a repo is empty exactly when it has no branches.

ENG-1284

### DIFF
--- a/crates/liboxen/src/repositories/revisions.rs
+++ b/crates/liboxen/src/repositories/revisions.rs
@@ -24,6 +24,17 @@ pub fn get(repo: &LocalRepository, revision: impl AsRef<str>) -> Result<Option<C
     Ok(commit)
 }
 
+/// Get a commit object from a commit id or branch name, off the async worker.
+/// Returns Ok(None) if the revision does not exist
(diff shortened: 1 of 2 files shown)


Eric-Laurence (approved):
looks good

PR #759, opened 2026-07-16
Async (11): meta handler (metadata::file) to async

CleanCut:
(Stacked on #753)

The metadata endpoint (`GET /api/repos/{namespace}/{repo}/metadata/{resource}`) no longer blocks an actix worker thread on synchronous RocksDB/LMDB reads: repo resolution, resource parse, and the meta-entry read now run off the worker, so a slow metadata read can no longer delay health-check responses or new-connection accepts.

Also drops a redundant commit lookup on every request. `parse_resource` already resolves and returns the commit, and the handler builds its response from that; the extra `commits::get_by_id `only fed a debug line. Removing it saves one commit read and one offload round-trip per request on a hot path.

The handler now awaits `get_repo_async`, `parse_resource_async`, and `entries::get_meta_entry_async` (already present); no liboxen API changes. Response shape and behavior are unchanged.

### DIFF
--- a/crates/oxen-server/src/controllers/metadata.rs
+++ b/crates/oxen-server/src/controllers/metadata.rs
@@ -1,8 +1,6 @@
 use crate::errors::OxenHttpError;
-use crate::helpers::get_repo;
-use crate::params::{app_data, parse_resource, path_param};
-
-use liboxen::error::OxenError;
+use crate::helpers::{get_repo, get_repo_async};
+use crate::params::{app_data, parse_resource, parse_resource_async, path_param};
 
 use liboxen::view::StatusMessage;
 use liboxen::view::entries::EMetadataEntry;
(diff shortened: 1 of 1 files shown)


crates/oxen-server/src/controllers/metadata.rs
```diff
@@ -50,18 +48,15 @@ pub async fn file(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHttpE
         resource
     );
 
-    let latest_commit = repositories::commits::get_by_id(&repo, &commit.id)?
```
Eric-Laurence, line 53: I was worried about removing this but it looks like it was redundant 👍

Eric-Laurence (approved):
Looks good! let's get this release out 🎉

PR #760, opened 2026-07-16
Async (12): commits reads + handler (commits::history) to async

CleanCut:
The `commits::history` endpoint reads commit history synchronously, walking RocksDB and the merkle tree inline on the actix worker that serves the request. Each oxen-server worker is single-threaded, so a large-history request holds its worker for the read's full duration, blocking that worker from accepting or serving anything else.

Move the endpoint's synchronous work off the worker. `repositories::commits::` list_from_paginated and `list_by_path_from_paginated` become async, each wrapping its whole operation in one `spawn_blocking`. The history handler resolves the repo and resource via the `async get_repo/parse_resource` variants and awaits both reads, freeing the worker to serve other requests while the database is read.

### DIFF
--- a/crates/liboxen/src/repositories/commits.rs
+++ b/crates/liboxen/src/repositories/commits.rs
@@ -192,65 +192,76 @@ pub fn search_entries(
 }
 
 /// List paginated commits starting from the given revision
-pub fn list_from_paginated(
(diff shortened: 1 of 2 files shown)


Eric-Laurence (approved):
Looks good!

PR #761, opened 2026-07-17
Drop official Windows support for oxen-server

CleanCut:
oxen-server (just the server) is no longer officially supported or tested on Windows. It may still work, but it's best-effort. NOTE: The oxen CLI, the oxenai Python package, and liboxen code that backs them stay fully supported on Windows.

Windows CI no longer builds or runs oxen-server or its server-dependent tests, so the Windows Rust and Python jobs are substantially faster and no longer fail on an unsupported setup. Those tests are skipped on Windows only. They still run on Linux and macOS. The Windows release stops shipping `oxen-server.exe`, and the `README` now documents which components run on which platform.

Tracked in ENG-1233

### DIFF
--- a/crates/liboxen/src/test.rs
+++ b/crates/liboxen/src/test.rs
@@ -9,6 +9,7 @@ use crate::constants;
 use crate::constants::DEFAULT_REMOTE_NAME;
 use crate::core;
 use crate::core::db::merkle_node::MerkleNodeBackend;
+use crate::core::df::duckdb_setup;
 use crate::core::versions::MinOxenVersion;
 use crate::error::OxenError;
 use crate::model::Schema;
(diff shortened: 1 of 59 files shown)


crates/liboxen/src/test.rs
Eric-Laurence, line 142: Could the shared helpers themselves skip Windows, so we don’t have to tag every test?

Eric-Laurence (approved):
Looks good!

crates/liboxen/src/test.rs
CleanCut: No, unfortunately. I mean, we could make them skip the logic, but then the test still launches and reports success, which takes time and is misleading. Rust doesn't have any way to indicate "skip this test" from inside the logic of a test.

PR #762, opened 2026-07-19
Async (13): compare/commits to async

Eric-Laurence:
The compare-commits endpoint finds the commits that are on head but not on base. It does this by walking the commit graph and reading each commit from RocksDB. That walk ran on the worker thread that serves requests, so a long history kept the worker busy until it finished. When a lot of requests come in together, busy workers can't pick up new ones.

`commits::list_between_exclusive` is now async and runs the walk on a background thread, so the worker stays free. The walk itself works exactly as before.

Completes ENG-1289


CleanCut (approved):
Excellent!

PR #763, opened 2026-07-20
S3 (27.2): Fix workspace add reporting failure on a successful upload

CleanCut:
`oxen workspace add` printed `added 0 entries ... failed to upload N entries` and still exited 0 when the upload had actually succeeded. Every successful add with the CLI at the root directory showed this false failure. A genuinely failed upload was indistinguishable because the command exited 0 either way. Python's `workspace.add(file, ".")` raised a spurious "Failed to upload" error on success for the same reason.

This fixes the problem by deriving the staged path on both the server and the CLI through the same helper function (`util::fs::workspace_staged_path`). Now the CLI exits nonzero when a file genuinely fails to upload, and does not emit a failure message when it succeeds.

### DIFF
--- a/crates/liboxen/src/api/client/workspaces/files.rs
+++ b/crates/liboxen/src/api/client/workspaces/files.rs
@@ -679,15 +679,12 @@ fn staging_filename(staging_path: &Path) -> String {
 }
 
 /// The path the workspace `files` endpoint stages a file at (and echoes back in its response),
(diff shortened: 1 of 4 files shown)


Eric-Laurence (approved):
Normalizing joined vs joining normalized. That's a slightly funny mixup! Looks good 👍

PR #764, opened 2026-07-20
Bound file descriptors when reassembling chunked version files

CleanCut:
Reassembling an uploaded file's chunks opened every chunk at once and held all of them open for the whole copy. With 10 MB chunks, that means a 100 GB file opens ~10,240 descriptors in a single request. We were hitting the process file-descriptor limit and returning 500 errors.

Reassemble version files through a new `AtomicFile::stream_from_paths` that concatenates the chunks in offset order, opening one file at a time.

### DIFF
--- a/crates/liboxen/src/util/fs/atomic_file.rs
+++ b/crates/liboxen/src/util/fs/atomic_file.rs
@@ -561,18 +561,18 @@ mod tests {
         .await
     }
 
-    #[tokio::test]
-    async fn test_atomic_stream_from_paths_concatenates_and_verifies() -> Result<(), OxenError> {
+    #[test]
+    fn test_atomic_stream_from_paths_concatenates_and_verifies() -> Result<(), OxenError> {
         use xxhash_rust::xxh3::xxh3_128;
 
-        test::run_empty_dir_test_async(|dir| async move {
+        test::run_empty_dir_test(|dir| {
             let chunks_dir = dir.join("chunks");
-            tokio::fs::create_dir(&chunks_dir).await?;
+            std::fs::create_dir(&chunks_dir)?;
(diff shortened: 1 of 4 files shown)


crates/liboxen/src/util/fs/atomic_file.rs
```diff
@@ -517,6 +561,63 @@ mod tests {
...
+                .with_hash(bogus)
+                .stream_from_paths(&[chunk]);
+
+            assert!(matches!(result, Err(OxenError::HashMismatch { .. })));
+            assert!(!target.exists(), "target must not be published on mismatch");
+            Ok(())
+        })
+        .await
+    }
+
```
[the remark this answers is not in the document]
CleanCut: Good catch. ✅ Fixed

Eric-Laurence (approved):
Looks good!

crates/liboxen/src/util/fs/atomic_file.rs
```diff
@@ -272,6 +272,50 @@ impl AtomicFile {
         tmp.commit()
     }
 
+    /// Publish the in-order concatenation of `paths` to the target path, opening one source file
+    /// at a time. Bounds open file descriptors to O(1) regardless of how many paths there are —
+    /// the fit for reassembling a file from many chunk files. On success the concatenated bytes
+    /// are hash-verified — when a hash was set via [`with_hash`][Self::with_hash] — and atomically
+    /// renamed into place; on any error the scratch file drops without touching the target.
+    pub fn stream_from_paths(self, paths: &[PathBuf]) -> Result<(), OxenError> {
```
Eric-Laurence, line 280: might be worth adding this to the claude md

CleanCut: Good idea! ✅ Done.

PR #770, opened 2026-07-21
Merkle LMDB (0.1): Resolve the merkle backend per repo, drop the env override

CleanCut:
A repository's merkle node backend is now decided from the repository itself. Its persisted `config.toml` is authoritative, and a repo whose config predates determines which merkle backend is used from its on-disk data. An existing filesystem node tree means the filesystem backend, otherwise LMDB. This drops the `OXEN_MERKLE_NODE_BACKEND` environment variable, a process-wide override that could point every repo on a server at the wrong store and make them read as empty. The backend is no longer a global switch that can silently disagree with where a repo's nodes actually live.

Resolving an existing repo's merkle tree backend is now separate from choosing a new repo's backend. New repos record a single default in their config at init, so the on-disk inference only ever applies to legacy repos. The default is unchanged (new repos are still filesystem-backed) so this is behavior-neutral groundwork for making LMDB the production backend.

### DIFF
--- a/crates/liboxen/src/api/client/tree.rs
+++ b/crates/liboxen/src/api/client/tree.rs
@@ -513,11 +513,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_download_tree_from_path() -> Result<(), OxenError> {
-        // FS-pinned: this test counts node directories under the on-disk `tree/nodes` layout, which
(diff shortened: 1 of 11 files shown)


Eric-Laurence (approved):
Looks good! 👍 Did a basic smoke test and it worked as expected

PR #771, opened 2026-07-21
Keep large pushes alive by streaming heartbeats during tree-node unpack

CleanCut:
Stream a newline "heartbeat" byte every 15 seconds while `POST /tree/nodes` unpacks, then send the JSON result as the final bytes, so idle timers keep resetting and an arbitrarily long unpack stays alive.

Large `oxen push` operations could fail with a 502 when the server went silent unpacking an uploaded merkle tree: with no bytes on the connection, the idle timers along the path could kill the request mid-operation. 

Leading whitespace is skipped by every JSON parser, so the client reads the final body unchanged. This needs no client, proxy, or frontend change and no minimum-version bump. The request body is drained before the response starts, so heartbeats cover only the silent unpack and the request is never read while the response streams.

Because the 200 status commits with the first heartbeat byte, a failure once the unpack has started is reported as an error in the response body (which the client still reads as a failure) rather than as an error status code.

### DIFF
--- a/crates/oxen-server/src/controllers/tree.rs
+++ b/crates/oxen-server/src/controllers/tree.rs
@@ -21,7 +21,7 @@ use liboxen::view::tree::nodes::{
 };
 
 use crate::errors::OxenHttpError;
-use crate::helpers::get_repo;
+use crate::helpers::{get_repo, stream_with_heartbeat};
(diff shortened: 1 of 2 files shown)


Eric-Laurence (approved):
Looks good!

PR #773, opened 2026-07-21
Merkle LMDB (0.2): Select a repo's Merkle node backend at creation

CleanCut:
A new repository can now be created directly on the LMDB Merkle node backend instead of the filesystem backend, which lets us stand up and exercise LMDB-backed repos ahead of making LMDB the default. Without this, the only route onto LMDB was migrating an already-created filesystem-backend repo; now the choice can be made up front.

The selector is exposed only where we need it for our own testing:
- `oxen init --merkle-backend <filesystem|lmdb>` for local repos
- `oxen create-remote --merkle-backend <filesystem|lmdb>` for remote repos
- a `merkle_node_backend` field on the repo-create request that the server honors

It is deliberately not surfaced in the Python bindings or the public docs.

### DIFF
--- a/crates/liboxen/src/api/requests/repo_new.rs
+++ b/crates/liboxen/src/api/requests/repo_new.rs
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 use crate::constants::DEFAULT_HOST;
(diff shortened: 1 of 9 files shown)


Eric-Laurence (approved):
Looks good!

PR #775, opened 2026-07-22
Merkle LMDB (1.1.1): Batch the FS↔LMDB merkle-node transcode

CleanCut:
The filesystem <-> LMDB merkle node migration now copies nodes in batches instead of one per store transaction, improving performance by avoiding lock overhead (LMDB) and parallelizing writes with rayon (filesystem). Batches are capped by a byte budget so a large repo's tree is never buffered whole.

Behavior is otherwise unchanged.


Eric-Laurence (approved):
Looks good!

PR #780, opened 2026-07-22
 Merkle LMDB (1.1.3): Remove the CLI `oxen migrate --all` feature

CleanCut:
In production, we manage rolling repository migrations from the hub. Having an individual migration attempt to manage multiple migrations as well is not going to lead to a good experience for open source server users. For now, let users manage their own migration however they like. In the future, if we want to support multiple migrations from the server side, let's give it some real design thought and implement it in a nice way that will apply to any migration.


Eric-Laurence (approved):
Looks good

PR #781, opened 2026-07-22
Log requests at start, and fix the empty request id in the access log

CleanCut:
Every request now produces an `INFO` log line the moment it arrives, alongside the existing completion line. A request that is still in flight (or one that hangs and never finishes) previously left no trace in the logs at all. Now it is visible as soon as it is received, which is exactly what you need when the server stops responding and you want to see what it was working on.

The completion (access) log line also advertised a request id but always logged it empty. It read the `x-oxen-request-id` response header, which is only set after the access logger has already rendered its line. It now reads the id straight from the request, so every completion line has the real id.

Both lines are prefixed with `start` and `end` and end with `req=<id>`, so the two halves of one request are easy to find together and tell apart. Each carries the fields the access log already knew at request time.

ENG-1332

### DIFF
--- a/crates/oxen-server/src/main.rs
+++ b/crates/oxen-server/src/main.rs
@@ -34,7 +34,7 @@ use actix_web::{App, HttpServer, web};
 use actix_web_httpauth::middleware::HttpAuthentication;
 use thiserror::Error;
 
-use middleware::{MetricsMiddleware, RequestIdMiddleware};
+use middleware::{MetricsMiddleware, RequestIdMiddleware, RequestStartLogMiddleware, request_id};
 use tracing_actix_web::TracingLogger;
 
 // Note: These 'view' imports are all for the auto-generated docs with utoipa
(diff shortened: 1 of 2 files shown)


crates/oxen-server/src/middleware.rs
```diff
@@ -21,6 +22,17 @@ pub fn generate_request_id() -> String {
...
 
+/// The request id assigned by [`RequestIdMiddleware`], stored in the request's extensions.
+struct RequestId(String);
+
+/// Returns the request id [`RequestIdMiddleware`] stored on the request, or `"-"` if none.
+pub fn request_id(req: &HttpRequest) -> String {
+    req.extensions()
+        .get::<RequestId>()
+        .map(|id| id.0.clone())
+        .unwrap_or_else(|| "-".to_string())
```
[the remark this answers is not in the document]
CleanCut: I've logged this in an issue to follow up on later.

crates/oxen-server/src/main.rs
```diff
@@ -670,9 +670,15 @@ async fn start(
             .service(web::scope("/api/repos").configure(routes::config))
             .default_service(web::route().to(controllers::not_found::index))
             .wrap(DefaultHeaders::new().add(("oxen-version", OXEN_VERSION)))
-            .wrap(Logger::new(
-                "%a \"%r\" %s %b \"%{Referer}i\" \"%{User-Agent}i\" %T %{x-oxen-request-id}o",
-            ))
+            .wrap(
+                Logger::new(
+                    "end %a \"%r\" %s %b \"%{Referer}i\" \"%{User-Agent}i\" %T req=%{request_id}xo",
```
jcelliott, line 675: Would be nice to add all these as structured log fields, but that's a larger problem for another day.

jcelliott (approved):

crates/oxen-server/src/main.rs
CleanCut: I agree on both points! I opened [ENG-1366](https://linear.app/oxenai/issue/ENG-1366/emit-oxen-server-access-logs-as-structured-tracing-fields) to circle back to this when we get a moment.

It would help to identify which format we want, or equivalently, what service we'd like to parse the logs.

PR #782, opened 2026-07-23
Merkle LMDB (1.1.4): Open the LMDB merkle env lazily on first access

CleanCut:
Stop opening a repo's LMDB merkle-node env when the store is constructed. Open it (creating it if absent) on the first read or write instead. A `LocalRepository` that never touches its merkle store (that includes every workspace) now opens no env at all.


Eric-Laurence (approved):
Looks good

PR #783, opened 2026-07-23
Merkle LMDB (1.1.5): Bump the merkle node mapping to 16 GiB

CleanCut:
(stacked on #782)

Now that we've gotten past the bug where every workspace opened a new merkle tree database, we can restore the virtual address space mapping back to a generous reservation. This is still much smaller than we started with, allowing for 16,384 simultaneous repositories' merkle trees open vs the previous setting's 1,024.


Eric-Laurence (approved):
👍

PR #785, opened 2026-07-23
Pin ruff version in pre-commit hooks

CleanCut:
CI started failing due to a ruff update -- but we're supposed to have the ruff version pinned.


Eric-Laurence (approved):
looks good

Eric-Laurence (commented):
if we're going to always use the pinned version why do we have the unpinned version laying around

CleanCut:
> if we're going to always use the pinned version why do we have the unpinned version laying around

Exactly.

PR #786, opened 2026-07-23
Stream the merkle-node upload to disk to bound oxen-server push memory

CleanCut:
Bound oxen-server's memory during pushes so a large or concurrent push can't drive the process to an out-of-memory kill that 502s every in-flight request.

The /tree/nodes upload handler read the entire uploaded node tarball into memory before unpacking it, with no size cap. The archive of directory/vnode/commit nodes runs to multiple gigabytes on a large repo, and several pushes can unpack at once, so concurrent uploads grew resident memory without bound. It now spools the upload to a temp file and unpacks it through a streaming reader, keeping peak memory flat regardless of tree size or push concurrency. The request body is still fully received before heartbeats stream during the (silent) unpack, so the wire behavior is unchanged.


Eric-Laurence (approved):
Looks good!

PR #788, opened 2026-07-23
Repo Integrity (1.14.2): Hold the write guard on every server write endpoint

CleanCut:
Make the whole-repo exclusive lock actually exclude concurrent writes: every oxen-server write endpoint now holds the per-repo write guard for its duration, so a maintenance operation that takes the exclusive lock — a migration, prune, or fsck — drains the in-flight writes and rejects new ones with a 429 rather than racing them. The guard mechanism shipped earlier, but no writer held it, so the exclusive lock had nothing to drain and writes ran unblocked.

Two deliberate scoping calls: the few GET endpoints that write on a cache miss (image resize, repo size, data-frame schema) take the guard for the whole handler, so those reads also back off during a maintenance op — they're tracked for migration off GET (ENG-1373/1374/1375). Destructive ops (versions clean, fsck, workspace clear) and repo lifecycle (create/delete/transfer) are exclusive-side work handled in later steps, not here.

### DIFF
--- a/crates/oxen-server/src/controllers/commits.rs
+++ b/crates/oxen-server/src/controllers/commits.rs
@@ -5,6 +5,7 @@ use liboxen::constants::DIRS_DIR;
 use liboxen::constants::HISTORY_DIR;
 use liboxen::constants::VERSION_FILE_NAME;
 
+use liboxen::core::repo_locks;
 use liboxen::error::OxenError;
 use liboxen::model::{Commit, LocalRepository};
 use liboxen::opts::PaginateOpts;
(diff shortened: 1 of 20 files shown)


crates/oxen-server/src/controllers/commits.rs
Eric-Laurence, line 1114: shouldn't this have a write guard too?

Eric-Laurence (approved):
Looks good!

crates/oxen-server/src/controllers/commits.rs
CleanCut: It looks like it, doesn't it? The crazy thing is that `upload_tree` is actually unreachable, dead code. My earlier dead-code-cleanups didn't find it because it's referenced from utoipa documentation. I'm going to open up another PR to clean up the dead code.

PR #791, opened 2026-07-24
Streamline Tests (13.1): Speed up repositories::status tests

CleanCut:
Cut the `repositories::status` test suite from ~9.4s to ~0.6s by giving each test the lightest fixture it actually needs instead of a full training-data repo with six commits.

Twelve tests spun up `run_training_data_repo_test_fully_committed_async` (the whole training tree plus six commits) only to inspect a single file, a two-file subdirectory, or a search path. Each now uses `run_empty_local_repo_test_async` and commits inline exactly the files its assertions touch, so the fixture setup (which was the entire per-test cost) nearly disappears.

Also delete `test_status_rm_directory_file` since it was a duplicate of `test_status_rm_regular_file`.

### DIFF
--- a/crates/liboxen/src/repositories/status.rs
+++ b/crates/liboxen/src/repositories/status.rs
@@ -234,23 +234,26 @@ mod tests {
 
     #[tokio::test]
     async fn test_command_modified_files_status_with_search_paths() -> Result<(), OxenError> {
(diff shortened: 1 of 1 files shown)


Eric-Laurence (approved):
Over 15x faster! Nice!

PR #793, opened 2026-07-24
Streamline Tests (13.2): Speed up repositories::restore tests

CleanCut:
Give each `repositories::restore` test the lightest fixture it actually needs instead of a full training-data repo spun up only to inspect one file, so the suite does far less setup I/O per test.

All eleven tests that used `run_training_data_repo_test_fully_committed_async` now use a lighter fixture. 
20 restore tests, all passing. Isolated-subset wall-clock under the cargo-test runner (ramdisk) went 0.96s to 0.58s; the larger benefit is in the full suite, where every test shares one thread pool.


Eric-Laurence (approved):
Looks good! 👍

PR #794, opened 2026-07-24
Repo Integrity (1.16.1): Report oxen-server panics to Sentry

CleanCut:
Report oxen-server panics to Sentry so a crash raises an alert on its own instead of sitting unnoticed in a log. Each event carries the panic message, a backtrace, and the release version; a panic that unwinds while serving a request also carries that request's context, so a crash can be tied to the endpoint that triggered it. `SENTRY_ENVIRONMENT`, when set, tags events with the environment (dev, staging, production) that they came from.

Reporting is off unless `SENTRY_DSN` is set — without it Sentry never initializes and nothing leaves the process, so self-hosted and OSS deployments report nothing (unless they set it to their own sentry). Events never attach request or user PII, and only errors are reported; no performance tracing is enabled at this point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,6 +269,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Eric-Laurence (approved):
Oh, perfect! This is incredible! Thanks for wiring this up, it will make debugging so much easier

PR #795, opened 2026-07-24
Speed up file metadata and directory listings on workspace-heavy repos

CleanCut:
Make workspace file metadata and directory listings much cheaper on repositories that have accumulated workspaces. Also, don't let one broken workspace fail metadata for every tabular file in the repo.

Answering "does this data frame have a queryable index?" loaded and fully constructed every workspace in the repository, and did it once per tabular file in a directory listing. So listing a directory of CSVs cost a full workspace load per entry per workspace. Each load resolved that workspace's pinned commit out of the merkle tree, re-read and re-parsed the repository config, and built two LocalRepository values, only for the caller to read two fields that were already sitting in the workspace config it had just parsed. Query workspaces are created with random ids and nothing ever collects them, so that multiplier grows for the life of the repository.

The check now decides from each workspace's config alone and loads at most one workspace, or none at all for the boolean the metadata path asks for.

This also fixes a fault-isolation bug. Unrelated workspaces' commits are no longer resolved, so a workspace whose config names a commit the merkle tree no longer holds is skipped rather than failing the request. Previously a single such workspace made metadata fail for every tabular file in the repository.

This mitigates rather than fixes the underlying problem. A queryable index is a derived artifact determined by the source file's content, yet it lives inside a randomly-named mutable container, so it can still only be found by searching every container, and those containers still accumulate without bound. ENG-1415 makes it addressable.

### DIFF
--- a/crates/liboxen/src/core/v_latest/workspaces/data_frames.rs
+++ b/crates/liboxen/src/core/v_latest/workspaces/data_frames.rs
@@ -6,6 +6,10 @@ use crate::core::db::data_frames::df_db;
 use crate::core::db::data_frames::df_db::with_df_db_manager;
 use crate::core::staged::get_staged_db_manager;
 use crate::core::v_latest::workspaces::files::{add, track_modified_data_frame};
+use crate::repositories::workspaces::data_frames::duckdb_path_in_dir;
+use crate::repositories::workspaces::{
+    list_dirs as list_workspace_dirs, read_config as read_workspace_config,
+};
 use parking_lot::Mutex;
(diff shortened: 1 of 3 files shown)


Eric-Laurence (approved):
Looks good, I can't believe we were doing all that unnecessary work before!

CleanCut:
We're _still_ doing a lot of unnecessary work! Just a lot less of it. ENG-1415 is going to take 10 times more effort to clean up the last bits of unnecessary work that's going on in this endpoint.

PR #797, opened 2026-07-26
Pin uv version to prevent flaky tests

Eric-Laurence:
Every Install uv step left the version unpinned, so the action fetched the latest-version manifest from raw.githubusercontent.com on each run. That fetch is flaky and periodically fails CI.

Pin uv to 0.11.29 in tool-versions.env and have every install path read it: bin/install-prereqs, the eight setup-uv workflows, and the Windows choco installs. With the version fixed, setup-uv skips that lookup and uses its cache, so the flaky call is gone and local and CI stay on the same uv.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -130,7 +130,7 @@ jobs:
           workspaces: "."
 
       - name: Load tool versions
-        run: cat tool-versions.env >> "$GITHUB_ENV"
+        run: grep -E '^[A-Z0-9_]+=' tool-versions.env >> "$GITHUB_ENV"
 
       - name: Install uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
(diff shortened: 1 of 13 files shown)


CleanCut (commented):
Hard-coding the version in 8 places (and missing `bin/install-prereqs`) will lead to inconsistent version changes, which is also flaky.

Would setting the version in a `uv.toml` file work instead?

If not, we should probably define the version in `tool-versions.env` and then have `bin/install-prereqs` and the 8 workflow files all read the version from there.

Eric-Laurence:
> Hard-coding the version in 8 places (and missing `bin/install-prereqs`) will lead to inconsistent version changes, which is also flaky.
> 
> Would setting the version in a `uv.toml` file work instead?
> 
> If not, we should probably define the version in `tool-versions.env` and then have `bin/install-prereqs` and the 8 workflow files all read the version from there.

That's a really good point.

I decided to use tool-versions.env, which I think is the right call over uv.toml. Like you said, tool-versions.env lets us have bin/install-prereqs use the same pinned version, and it also lets us pin the versions installed by choco in the windows ci. It's also just the place where we pin other tool versions, so it makes sense to keep uv in there as well.

CleanCut (approved):

PR #799, opened 2026-07-26
Async (14): compare/entries offload to async

Eric-Laurence:
The compare-entries endpoints ran their tree work on the request worker: reading both commit trees, collecting which files and directories changed, and building the top-level directory summary. All of it now runs on a background thread.

Completes ENG-1290

### DIFF
--- a/crates/liboxen/src/core/v_latest/diff.rs
+++ b/crates/liboxen/src/core/v_latest/diff.rs
@@ -24,11 +24,22 @@ pub async fn list_diff_entries(
     page: usize,
     page_size: usize,
 ) -> Result<DiffEntriesCounts, OxenError> {
-    log::debug!(
-        "list_diff_entries base_dir: '{base_path:?}', head_dir: '{head_path:?}' base_commit: '{base_commit}', head_commit: '{head_commit}'"
-    );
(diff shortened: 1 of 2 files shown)


crates/liboxen/src/core/v_latest/diff.rs
```diff
@@ -24,11 +24,22 @@ pub async fn list_diff_entries(
...
+        let repo = repo.clone();
+        let base_commit = base_commit.clone();
+        let head_commit = head_commit.clone();
+        tokio::task::spawn_blocking(move || -> Result<_, OxenError> {
+            let repo = &repo;
+            let base_commit = &base_commit;
+            let head_commit = &head_commit;
+            log::debug!(
+                "list_diff_entries base_dir: '{base_path:?}', head_dir: '{head_path:?}' base_commit: '{base_commit}', head_commit: '{head_commit}'"
+            );
```
CleanCut: Hmm. Looks like you stumbled upon a pre-existing formatting bug. Please delete this debug log (I don't think it's useful) and run `cargo fmt` and the formatting should all fix itself. Then ship it.

CleanCut (approved):
A minor item to fix and then 🚢 it.

PR #800, opened 2026-07-26
Async (15): compare/file offload to async

Eric-Laurence:
The compare file endpoint builds a diff between two versions of a data frame. Several parts ran on the request worker: reading the two file versions, hashing both frames to find the changed rows, pulling those rows out, and, on a schema change, selecting the shared columns. All of it now runs on a background thread.

Completes ENG-1291

### DIFF
--- a/crates/liboxen/src/view/tabular_diff_view.rs
+++ b/crates/liboxen/src/view/tabular_diff_view.rs
@@ -131,8 +131,18 @@ impl TabularDiffView {
                 } else {
                     let cols = column_names.iter().map(col).collect::<Vec<Expr>>();
 
-                    let common_base_df = base_df.clone().lazy().select(&cols).collect().unwrap();
-                    let common_head_df = head_df.clone().lazy().select(&cols).collect().unwrap();
+                    // Selecting the common columns and collecting both frames is sync polars CPU.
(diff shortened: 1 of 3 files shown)


crates/liboxen/src/view/tabular_diff_view.rs
```diff
@@ -131,8 +131,18 @@ impl TabularDiffView {
...
+                    let (common_base_df, common_head_df) = {
+                        let base_df = base_df.clone();
+                        let head_df = head_df.clone();
+                        tokio::task::spawn_blocking(move || {
+                            let common_base_df = base_df.lazy().select(&cols).collect().unwrap();
+                            let common_head_df = head_df.lazy().select(&cols).collect().unwrap();
+                            (common_base_df, common_head_df)
+                        })
+                        .await
+                        .unwrap()
```
CleanCut, line 144: Ideally, we should propagate the error here. However, if that's going to be a lot of work, we can just change this to an `.expect(...)` with a message so we can track the panic back to this location.

Someday soon we need to do a full sweep of the codebase to convert unwrap/expect usage to error propagation anyway.

CleanCut (approved):

PR #803, opened 2026-07-27
Fix oxen-server crash by publishing data frame writes atomically

CleanCut:
Rewriting a cached data frame truncated the file in place, so any request still holding the old contents memory-mapped faulted on its next page touch and killed the whole server process with `SIGBUS`. A running server hit this repeatedly, dropping every in-flight request each time. Data frame writes now land in a sibling temp file and rename over the target, so a reader that already mapped the previous version keeps a valid file until it finishes.

We fix two other kinds of problems in this PR as well: Four of the five data frame writers turned a failed file create into a panic instead of a returned error. And the compare cache's commit-id markers were written non-atomically, so a crash mid-write could leave a torn marker that makes a stale cache look current.

We add the `AtomicFile::write_with` terminal method and the `HashingWriter` helper struct as part of this PR.

ENG-1420

### DIFF
--- a/crates/liboxen/src/core/df/tabular.rs
+++ b/crates/liboxen/src/core/df/tabular.rs
@@ -22,6 +22,7 @@ use crate::opts::{CountLinesOpts, DFOpts, PaginateOpts};
 use crate::repositories;
 use crate::storage::{VersionLocation, VersionStore};
 use crate::util::fs;
+use crate::util::fs::AtomicFile;
 use crate::util::hasher;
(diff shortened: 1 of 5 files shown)


Eric-Laurence (approved):
Looks good!

One small point, atomic_file.rs has extensive tests for write, stream, copy_from, etc., but not write_with (including its with_hash path). HashingWriter has no tests either, while HashingReader does

CleanCut:
Excellent point! I'll add some tests.

PR #804, opened 2026-07-27
Stop publishing two HTTP endpoints that don't exist

CleanCut:
docs.oxen.ai renders its HTTP API reference from oxen-server's OpenAPI spec, which listed POST /commits/{commit_id}/upload_tree and GET /commits as real endpoints. Neither has a route, so anyone following the reference got a 404. Both are gone from the spec now, along with the handlers behind them and two more handlers that no request could reach either.

Every deletion is unreachable code, so no request behaves differently:

- `commits::upload_tree` lost its route in November 2023 and its client caller in March 2025; only the OpenAPI entry kept it looking alive. Its `unpack_tree_tarball` helper goes with it.
- `commits::index` has had its route commented out in the commits service, but stayed documented.
- `oxen_version::resolve` was never routed, and was the only thing using the `repo_root` route name.
- `versions::chunks::download` was registered six lines after the `GET /versions/{resource:.*}` catch-all that swallows every path it could match. Its own comments recorded it as "not functional" and "not used anywhere". That URL already fell through to the catch-all, and still does.

The commit-upload endpoint's description pointed readers at the deleted upload_tree endpoint; it now points at the tree nodes endpoint, which is where merkle tree nodes actually go.

### DIFF
--- a/crates/oxen-server/src/controllers/commits.rs
+++ b/crates/oxen-server/src/controllers/commits.rs
@@ -70,31 +70,6 @@ pub struct ListMissingFilesQuery {
     pub head: String,
 }
 
-/// List commits
-#[utoipa::path(
-    get,
-    path = "/api/repos/{namespace}/{repo_name}/commits",
-    tag = "Commits",
-    description = "List all commits in the repository's history.",
-    params(
-        ("namespace" = String, Path, description = "Namespace of the repository", example = "ox"),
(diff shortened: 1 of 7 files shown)


Eric-Laurence (approved):
Looks good, nice cleanup! 🧹 

Because of this, I noticed we have a dead oxen server proxy route on the Hub too, for something that the oxen server dropped a while ago: https://github.com/Oxen-AI/OxenHub/pull/1414

PR #806, opened 2026-07-27
Recover stale workspace data frames on commit instead of erroring

jcelliott:
A staged DuckDB table that exists but fails the fully-indexed gate was either left by an older version of oxen (which tracked edits in extra columns and soft-deleted rows by tombstoning them as 'removed') or by an interrupted index. Committing such a workspace previously returned WorkspaceStaleStagedIndex: the base file could not be committed without silently dropping the staged edits, and the edits could not be recovered because re-indexing rebuilds from the committed base version file.

Add reindex_preserving_rows, which rebuilds a stale-but-present staged table from its own current rows rather than from the committed base: it drops every historical internal tracking column, omits rows the old flow tombstoned as 'removed' (honoring the pending deletion), preserves row order, and re-indexes through the current indexer so the table lands in the current format (fresh _oxen_id/_oxen_row_id and the index marker).

export_tabular_data_frames now recovers such a table and exports it like any indexed one, so appended and modified rows survive the commit while deleted rows stay deleted. WorkspaceStaleStagedIndex is retained only for the genuinely-unrecoverable case where the table has no user columns.

### DIFF
--- a/crates/liboxen/src/core/v_latest/workspaces/data_frames.rs
+++ b/crates/liboxen/src/core/v_latest/workspaces/data_frames.rs
@@ -1,8 +1,8 @@
 use duckdb::Connection;
 
 use crate::constants::{
-    DIFF_STATUS_COL, EVAL_DURATION_COL, EVAL_ERROR_COL, EVAL_STATUS_COL, EXCLUDE_OXEN_COLS,
-    OXEN_ID_COL, OXEN_ROW_ID_COL, TABLE_NAME,
+    EVAL_DURATION_COL, EVAL_ERROR_COL, EVAL_STATUS_COL, EXCLUDE_OXEN_COLS, OXEN_ID_COL,
+    OXEN_ROW_ID_COL, TABLE_NAME,
 };
 use crate::core::db::data_frames::DataFrameError;
 use crate::core::db::data_frames::df_db;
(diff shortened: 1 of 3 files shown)


crates/liboxen/src/core/v_latest/workspaces/data_frames.rs
```diff
@@ -246,6 +249,157 @@ pub async fn index(workspace: &Workspace, path: &Path) -> Result<(), OxenError>
...
+const LEGACY_AND_CURRENT_INTERNAL_COLS: [&str; 7] = [
+    OXEN_ID_COL,
+    OXEN_ROW_ID_COL,
+    DIFF_STATUS_COL,
+    // Legacy per-row change hash, dropped along with the other tracking columns;
+    // it no longer has a named constant, so it appears here as a literal.
+    "_oxen_diff_hash",
+    EVAL_STATUS_COL,
+    EVAL_ERROR_COL,
+    EVAL_DURATION_COL,
```
CleanCut, line 284: Nit: This table is supposed to list old column names as they once were, right. That might be safer as literal strings rather than constants that some other change could yank out from under them (or maybe already has). But if those values have never changed until now, maybe that's overkill.

CleanCut (approved):

PR #807, opened 2026-07-28
Prune stale deps from Windows release cache

Eric-Laurence:
The Windows dependency cache grows every time our deps change. Each change leaves the old versions' artifacts behind and nothing clears them, so the cached tree is up to 23GB now, most of it stale.

This clears the stale artifacts before saving, keeping only what the current build uses, which takes the tree from 23GB down to about 12GB. It leaves the bundled C/C++ builds (duckdb, rocksdb) alone, since clearing those would force a recompile, so the tree still creeps up slowly when a native dep changes.


Eric-Laurence:
Superseded by Nathan's PR

PR #812, opened 2026-07-30
Merkle LMDB (1.2.4): Clone into the same merkle backend the server uses

CleanCut:
Cloning an LMDB-backed repository produced a filesystem-backed local one, so there was no way to get an LMDB repository locally at all short of migrating it by hand afterward.

A clone now adopts the backend the remote reports, and `oxen clone --merkle-backend` overrides it — the same flag `oxen init` and `oxen create-remote` already take. A remote that reports no backend, which is any server predating the field, still resolves to the create-time default.


Eric-Laurence (approved):
Looks good, and passed my smoke test 👍

PR #825, opened 2026-08-01
Fix directory entry counts that miscounted subdirectories

CleanCut:
A directory's stored entry count over-reported, permanently and cumulatively. Each commit seeded the count from the previous one and then re-counted the directory's subdirectories on top of it, so any directory containing a subdirectory gained a phantom entry every time a commit rewrote it. Removing a subdirectory compounded this further, since only file removals ever decremented. `oxen tree` reported counts larger than the directory held.

The count is now taken from the entries the directory actually ends up holding rather than carried forward and adjusted. The vnodes available at that point already hold the complete post-commit child set with removals applied, and vnode entry counts were computed this way all along, which is why they stayed correct while directory counts drifted.

Directory listings were never affected. They count real entries rather than reading the stored field, so the visible reach was tree output and the value carried into the next commit. Existing repositories keep their drifted values in already-written nodes; recomputing them is left to the merkle node format migration, which reuses the true-count helper added here.

Also removes a latent underflow that the old subtraction could hit on directories whose stored count was zero, and corrects the `num_entries` documentation, which claimed the count included vnodes. It counts files and sub directories only.

Sentry: OXEN-SERVER-3

### DIFF
--- a/crates/liboxen/src/model/merkle_tree/node/dir_node.rs
+++ b/crates/liboxen/src/model/merkle_tree/node/dir_node.rs
@@ -15,7 +15,7 @@ pub trait TDirNode {
     fn name(&self) -> &str;
     fn set_name(&mut self, name: &str);
     fn num_files(&self) -> u64; // This us just the number of files
-    fn num_entries(&self) -> u64; // This is the number of files and directories and vnodes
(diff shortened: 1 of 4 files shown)


Eric-Laurence (approved):
Looks good, nice fix!

PR #829, opened 2026-08-04
Reject upload targets that resolve outside the repository

CleanCut:
A single-file upload used its target path as given, so a path that climbed out of the repository wrote its contents outside the repo before a later step failed the request with a generic 500. The response gave no sign that a file had landed outside the repository at all.

Single-file uploads now go through the same target-path validation the directory upload modes already used, so a target that escapes the repository is refused with 400 before anything is written. Valid relative targets are unchanged, and "./" segments collapse as they do elsewhere.

Both the normal and initial-commit upload paths share the one validation point, so neither can accept a target the other rejects.

### DIFF
--- a/crates/oxen-server/src/controllers/file.rs
+++ b/crates/oxen-server/src/controllers/file.rs
@@ -639,7 +639,7 @@ fn build_files_from_upload_parts(
 
             let temp_file = take_single_file_part(file_parts)?;
             Ok(vec![FileNew {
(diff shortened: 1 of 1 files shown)


Eric-Laurence (approved):
Ooh nice fix, how did you catch this 👀

CleanCut:
> Ooh nice fix, how did you catch this 👀

Coderabbit caught it as a too-late review on one of the other PRs that touched the code.

PR #833, opened 2026-08-04
Merkle LMDB (1.2.7): Delete repositories the server cannot open

CleanCut:
A repository whose config declares an on-disk format this build no longer reads could not be deleted: the handler reported it as "not found" while its directory stayed on disk, leaving no way for any caller to reclaim the space.

The handler now returns 404 only when the repository directory is genuinely absent. When the directory is present but the repository will not open, it removes the directory and logs the reason it could not be opened. Version blobs kept outside the directory are unreachable without the config, so they are left in place.

ENG-1542

### DIFF
--- a/crates/oxen-server/src/controllers/repositories.rs
+++ b/crates/oxen-server/src/controllers/repositories.rs
@@ -516,15 +516,31 @@ pub async fn delete(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHtt
     let namespace = path_param(&req, "namespace")?.to_string();
     let name = path_param(&req, "repo_name")?.to_string();
 
-    let Ok(repository) = get_repo(app_data, &namespace, &name) else {
-        return Ok(HttpResponse::NotFound().json(StatusMessage::resource_not_found()));
+    let repository = match get_repo(app_data, &namespace, &name) {
+        Ok(repository) => Some(repository),
+        Err(OxenHttpError::InternalOxenError(OxenError::RepoNotFound(_))) => {
+            return Ok(HttpResponse::NotFound().json(StatusMessage::resource_not_found()));
+        }
+        // A repository the server cannot open is still deleted. Reporting it as missing would
+        // strand the directory on disk with no way for a caller to reclaim it, and version blobs
+        // held outside the directory are unreachable without the repository config anyway.
+        Err(err) => {
+            log::warn!("Deleting unreadable repo {namespace}/{name}: {err}");
+            None
+        }
     };
+    let repo_dir = app_data.path.join(&namespace).join(&name);
 
     // Delete in a background task because it could take awhile; the blocking directory
     // removal runs inside delete's own spawn_blocking.
     tokio::spawn(async move {
-        match repositories::delete(&repository).await {
-            Ok(_) => log::info!("Deleted repo: {namespace}/{name}"),
+        let result = match &repository {
+            Some(repository) => repositories::delete(repository).await.map(|_| ()),
+            None => repositories::delete_dir(&repo_dir).await,
+        };
+
+        match result {
+            Ok(()) => log::info!("Deleted repo: {namespace}/{name}"),
             Err(err) => log::error!("Err deleting repo: {err}"),
         }
     });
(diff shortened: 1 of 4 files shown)


crates/oxen-server/src/controllers/repositories.rs
```diff
@@ -590,3 +606,73 @@ pub async fn transfer_namespace(
...
+        let req = test::repo_request(&sync_dir, "/", namespace, repo_name);
+        let resp = super::delete(req)
+            .await
+            .expect("delete handler should succeed");
+
+        assert_eq!(resp.status(), http::StatusCode::OK);
+        assert!(
+            wait_until_gone(&repo_dir).await,
+            "repo dir should be deleted: {repo_dir:?}"
+        );
```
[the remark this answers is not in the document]
CleanCut: Too trivial

crates/oxen-server/src/controllers/repositories.rs
```diff
@@ -516,15 +516,31 @@ pub async fn delete(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHtt
...
+        Err(OxenHttpError::InternalOxenError(OxenError::RepoNotFound(_))) => {
+            return Ok(HttpResponse::NotFound().json(StatusMessage::resource_not_found()));
+        }
+        // A repository the server cannot open is still deleted. Reporting it as missing would
+        // strand the directory on disk with no way for a caller to reclaim it, and version blobs
+        // held outside the directory are unreachable without the repository config anyway.
+        Err(err) => {
+            log::warn!("Deleting unreadable repo {namespace}/{name}: {err}");
+            None
+        }
```
Eric-Laurence, line 538: This catches every failure, meaning it also deletes healthy repos when the cause is transient.

(e.g. error reading `config.toml`, corrupt TOML, `S3BackendMissingServerOpts` if a server starts without its `--config`, etc.)

All of those used to 404 and let the caller retry. The S3 one wipes local metadata and leaves the blobs.

Returning the error also gets the status right, since `UnsupportedRepoVersion` already maps to 400 and the rest to 500.

```suggestion
        // A repository whose on disk format is unreadable is deleted.
        // Reporting it as missing would strand the directory on disk.
        // Version blobs held outside the directory are unreachable without the config regardless.
        Err(OxenHttpError::InternalOxenError(err @ OxenError::UnsupportedRepoVersion(_))) => {
            log::warn!("Deleting unreadable repo {namespace}/{name}: {err}");
            None
        }
        // Every other failure to open can be transient, the repository may be intact.
        // Return the error and let the caller retry.
        Err(err) => return Err(err),
```

crates/oxen-server/src/controllers/repositories.rs
```diff
@@ -516,15 +516,31 @@ pub async fn delete(req: HttpRequest) -> actix_web::Result<HttpResponse, OxenHtt
...
     };
+    let repo_dir = app_data.path.join(&namespace).join(&name);
```
Eric-Laurence: This deletes a path built from raw URL segments.

Because actix does not protect `.` from decoding, `%2e%2e` decodes to `..`. I tested this locally, and `/api/repos/%2e%2e/%2e%2e` resolves to `SYNC_DIR/../..`. In prod, that would be `/var`. Since `remove_dir_all` has no containment check, it can delete all repos if this path is reached. main just 404s here.

We can use a guard, or we could take the directory from the lookup that we already did instead of constructing it again from the params. I implemented the guard below since it's the simpler change.

```suggestion
    // namespace and name come from URL segments, where %2e%2e decodes to '..'
    // To prevent climbing out of the sync dir, accept only a plain segment for each.
    // This rejects things like ., .., /, and empty strings.
    let plain_segment = |segment: &str| {
        let mut components = Path::new(segment).components();
        matches!(components.next(), Some(Component::Normal(_))) && components.next().is_none()
    };
    if !plain_segment(&namespace) || !plain_segment(&name) {
        return Ok(HttpResponse::NotFound().json(StatusMessage::resource_not_found()));
    }
    let repo_dir = app_data.path.join(&namespace).join(&name);
```

crates/oxen-server/src/controllers/repositories.rs
Eric-Laurence, line 32: We'd need to add component and path for that second suggestion

```suggestion
use std::path::{Component, Path, PathBuf};
```

Eric-Laurence (approved):
Looks good, I have two primary concerns.

crates/oxen-server/src/controllers/repositories.rs
Eric-Laurence: This suggestion is rendering oddly, I'm not sure if it would apply correctly

crates/oxen-server/src/controllers/repositories.rs
CleanCut: 👍🏻 I agree that deletion needs to be improved further. I created issue [ENG-1547](https://linear.app/oxenai/issue/ENG-1547/repository-deletion-is-instant-and-irreversible-with-no-typed) yesterday to track further improvements.

This is a step in the right direction, though. This saves me hours of work now with no possible blast radius, yet. Yesterday I wasted a couple hours figuring out that repos I was deleting were getting deleted in the hub and then silently left on the server and I had to manually clean up the server. We don't actually have any S3-backed repositories yet, and they would only be left in a bad state if the repo itself were in an incorruptible, unopenable state after having successfully put versioned files in S3. So, I don't expect it to be a problem during the window between now and the followup in ENG-1547.

crates/oxen-server/src/controllers/repositories.rs
CleanCut: ✅ Excellent catch!

This problem also existed in a bunch of other places in the server, so I made a centralized solution for it and fixed it in all those places.

crates/oxen-server/src/controllers/repositories.rs
CleanCut: I implemented differently, so these specific imports aren't needed.

PR #835, opened 2026-08-05
Report each error condition as one issue instead of one per identifier

CleanCut:
Error reports are grouped by message text, so a hash, commit id, or path interpolated into an error message filed a separate issue for every distinct value. One missing version blob produced its own issue, a resolved one never reopened on recurrence because the repeat arrived under a new fingerprint, and a systemic condition looked like many unrelated one-off failures.

Eight reporting sites now log a constant message and carry the identifier as a structured field instead, so a condition groups as one issue while every occurrence keeps the value needed to act on it.

The unmapped-error fallback deliberately keeps its error text in the message. Every variant without its own arm reaches that single line, so a constant message there would collapse unrelated failures into one bucket. An unmapped error carrying an identifier still splits, and the remedy for that is giving it its own arm.

ENG-1540

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5711,6 +5711,7 @@ dependencies = [
  "tracing",
  "tracing-actix-web",
  "tracing-opentelemetry",
+ "tracing-subscriber",
  "url",
  "utoipa",
  "utoipa-swagger-ui",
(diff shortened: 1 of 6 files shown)


Eric-Laurence (approved):
Looks good! 🧹

PR #838, opened 2026-08-05
Report each failed request once instead of two or three times

CleanCut:
A single failure could raise three separate reports: an origin-site log, the error-response arm, and a second event the request middleware emits for every failed request. One condition then appeared as several unrelated issues, each needing its own triage, and event counts read several times busier than the real failure rate.

The middleware's duplicate event is now off. It was enabled only because it is a default feature of the crate, never a deliberate choice, and it added nothing the other reports lack. Request context comes from the per-request binding, so our own events already carry the method, URL, and transaction. Its message also embedded the error's debug output, so an identifier inside one split a single condition across many reports.

Reporting a server error is now the job of whichever code holds the detail. Two repository endpoints logged a failed lookup that the library function they call had already reported, so they now leave it to that function, which names the failing repository more precisely than the endpoint could. The generic internal-error response carried the error but logged nothing at all, leaving those 500s with no record anywhere, so it now reports through the shared path and returns the same body as before. The empty internal-error variant stays silent because it carries nothing worth saying, and every failure that reaches it is already reported.

The reports that remain group by condition rather than by value. Repository and namespace failures in the library interpolated a path and the error text into the message, splitting one condition into a separate issue per repository, and now log a constant message with those values as fields. Namespace size failures gained the repository path they never carried, and a failed size calculation reports at error level rather than warn, where it raised no alert at all.

The error handling doc records the obligation this creates, where the log belongs, and how to check that a report does not already exist.

ENG-1474

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -167,7 +167,9 @@ tokio-stream = "0.1.17"
 tokio-util = { version = "0.7.8", features = ["compat", "io", "io-util"] }
 toml = "0.8.19"
 tracing = "0.1"
-tracing-actix-web = "0.7"
+# `emit_event_on_error` (the crate's only default feature) is off; oxen-server reports every
+# failed request itself. See docs/error_handling.md.
+tracing-actix-web = { version = "0.7", default-features = false }
 tracing-appender = "0.2"
 tracing-log = "0.2"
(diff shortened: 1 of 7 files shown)


Eric-Laurence (approved):
👍

PR #841, opened 2026-08-06
Repo Integrity (1.17.1): Alert when a repo is missing data it references

CleanCut:
A repo whose server lost a file now raises a Sentry issue the first time anyone asks for that file, instead of staying invisible until a customer reports a clone that will not finish. Until now the only way we learned about a broken repo was someone telling us.

ENG-1567


Eric-Laurence (approved):
👍

PR #847, opened 2026-08-07
Hash a file without metadata the same way in every writer

CleanCut:
Three writers computed a file node's metadata and combined hashes, and the two that took an `Option` handled an absent value differently from `add`: they hashed the string "null" and stored it as a real metadata hash, giving identical content two different tree identities depending on which path committed it.

All three now leave the metadata hash unset and the combined hash equal to the content hash when there is no metadata, matching `add` and `FileNode::recompute_metadata_hashes`. Behavior is unchanged when metadata is present.

No behavior changes today: every path reaching these writers is currently gated so that metadata is always present. This removes the trap for whoever relaxes one of those guards, since a combined hash feeds the vnode and dir node hashes and a disagreement there is silent.

### DIFF
--- a/crates/liboxen/src/core/v_latest/data_frames/schemas.rs
+++ b/crates/liboxen/src/core/v_latest/data_frames/schemas.rs
@@ -267,14 +267,14 @@ pub fn add_schema_metadata(
     }
 
     let oxen_metadata = &file_node.metadata();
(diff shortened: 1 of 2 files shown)


Eric-Laurence (approved):
Looks good!

PR #852, opened 2026-08-10
Constrain inbound request ID format and length

jcelliott:
An `x-oxen-request-id` header was adopted verbatim, with no limit on length or content. The value is returned on the response, written to both access-log lines, and recorded on every span of the request as `oxen.request_id`, so an untrusted caller could inflate all three, and a value carrying tabs or spaces (which a header may) blurs the access-log format.

An id is now adopted only when it is 1 to 128 characters of `A-Za-z0-9`, `-`, or `_` — the shape of a UUID or a URL-safe base64 id, with room to spare. Anything else is replaced with a generated one, the same path a request arriving without the header already takes. Every consumer reads the id from this one function, so the response header, the two log lines, the task-local context and the span attribute all get the checked value.


Eric-Laurence (approved):
Looks good! We'll need a similar guard on the Hub as well in the places that it accepts a request id (from the frontend).

PR #854, opened 2026-08-10
Stop contending for the shared rayon pool when summing entry sizes

CleanCut:
We don't need to send entries to individual threads in a thread pool in order to read their `num_bytes` field. 🤣 

ENG-1250


Eric-Laurence (approved):
PARALLELIZE EVERYTHING!!! Or, actually, maybe don't. 😂

Looks good, nice catch

PR #855, opened 2026-08-10
Keep a data-frame transform's collect off the request worker

CleanCut:
Move sync work into a thread so it doesn't block the async request worker.

ENG-1432


Eric-Laurence (approved):
Looks good

PR #861, opened 2026-08-12
improve workspace api documentation and comments

gschoeni:
Realized there were some gaps in our workspace api docs when working on the text editor project


crates/oxen-server/src/controllers/workspaces/changes.rs
```diff
@@ -18,6 +18,25 @@ use actix_web::{HttpRequest, HttpResponse, web};
 
 use std::path::PathBuf;
 
+/// List staged changes in a workspace
+#[utoipa::path(
+    get,
+    path = "/api/repos/{namespace}/{repo_name}/workspaces/{workspace_id}/changes",
+    description = "List the staged changes (added, modified, and removed files) in a workspace, paginated.",
```
Eric-Laurence: I believe this endpoint paginates each of the three file lists independently, I think the description could be made a little more helpful

crates/oxen-server/src/controllers/workspaces/changes.rs
```diff
@@ -55,6 +79,26 @@ pub async fn list_root(
     Ok(HttpResponse::Ok().json(response))
 }
 
+/// List staged changes under a directory in a workspace
+#[utoipa::path(
+    get,
+    path = "/api/repos/{namespace}/{repo_name}/workspaces/{workspace_id}/changes/{path}",
+    description = "List the staged changes (added, modified, and removed files) under a directory in a workspace, paginated.",
```
Eric-Laurence: same here: I believe this endpoint paginates each one independently, I think the description could be made a little more helpful

Eric-Laurence (approved):
Looks good!

PR #863, opened 2026-08-13
Cache the dependency build in the registry with cargo-chef

jcelliott:
# Proposal - don't merge yet

Cuts the dev image build from about 8m50s to **6m07s** on an ordinary merge, a 31% saving.

`COPY . .` puts the whole tree in one layer, so any source change invalidates it and the build recompiles all 700 dependencies. cargo-chef splits that: a planner stage reduces the tree to `recipe.json`, the builder cooks dependencies from it, and source is copied afterwards. The runner is destroyed after each job, so the cache lives in the registry, which needs the docker-container buildx driver. Only `main` writes it.

## Measurements

arm64, `m7g.4xlarge`, `docker build` step only.

| | `main` today | cold | ordinary merge |
|---|---|---|---|
| `cargo chef cook` | n/a | 700 crates, 264.2s | 49.6s restored, 0 compiled |
| `cargo build` | n/a | 3 crates, 218.6s | 3 crates, 224.4s |
| cache export | n/a | 109.8s | 17.6s |
| build step | ~8m50s | 11m00s | **6m07s** |

Runs 31546916895, 31547808017 and 31728339790. The last one changed the build context without touching a manifest, which is the shape of a normal merge.

The cold column is what a build pays whenever `recipe.json` changes: the first merge after this lands, and every later dependency edit. A slow build there is expected rather than a regression. A release version bump does not invalidate it, because cargo-chef normalizes member versions in the skeleton.

## Why the toolchain moves into the base stage

The first attempt succeeded while being slower than what it replaced, at 17m11s. `cook` ran in a stage holding only `recipe.json`, so `rust-toolchain.toml` was absent and cook compiled against the image's 1.94 rather than the pinned 1.97.1. rustc's version is a global fingerprint input, so the build reused none of that work. `rust-toolchain.toml` and `.cargo/config.toml` now land in the `chef` stage ahead of anything that compiles.

That failure mode is the thing worth checking in review. Any global compilation input added outside the `chef` stage silently stops the cache working, and the only symptom is a slower build.

## For

- 2m43s off every merge to `main`, at roughly five merges a day.
- The dependency phase is gated by the `libduckdb-sys` and `librocksdb-sys` C/C++ build scripts, which no Rust-side tuning reaches.
- The apt and FFmpeg layers start being reused as well.
- Branches read the cache but cannot write it, so a branch cannot replace what `main` and the release builds read.

## Against

- Three build stages, a buildx driver switch, a registry cache, and a second GHCR package with its own pruning.
- The silent failure described above, guarded only by a comment.
- Roughly 100s of the theoretical 4m22s saving goes on restoring the cooked layer and copying source into two stages, so the payoff falls well short of the ceiling.
- `release_docker.yml` is untouched, so releases get none of this.

### DIFF
--- a/.github/workflows/build_branch.yml
+++ b/.github/workflows/build_branch.yml
@@ -40,6 +40,10 @@ env:
   AWS_ROLE: ${{ vars.EC2_GITHUB_RUNNER_ROLE }}
   # Lowercase: registry paths are case-sensitive and the owner name is not.
   REGISTRY_IMAGE: ghcr.io/oxen-ai/oxen-server
+  # Its own package: prune_server_images.yml keeps the newest 20 versions of
+  # oxen-server, and a cache manifest per build would spend that budget on entries no
+  # deployment can roll back to.
+  CACHE_IMAGE: ghcr.io/oxen-ai/oxen-server-buildcache
 
 # Each job below states what it needs; a job's own block replaces this one rather
 # than adding to it, so nothing inherits a write it has no use for.
@@ -113,14 +117,49 @@ jobs:
           echo "COMMIT=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
           echo "BUILD_DATE_UTC=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_OUTPUT
 
+      # Ahead of the build because the cache below reads and writes the registry.
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # The runner is destroyed after the job, so the cache has to live in the
+      # registry. The default docker driver refuses --cache-to type=registry.
+      - name: Set up Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker-container
+
       # Thin LTO in place of the release profile's fat LTO, a serial link-time pass
       # that reruns whenever this repository's own crates change, so no layer cache
       # can avoid it. `release_docker.yml` passes no override, so released binaries
       # keep fat LTO. `codegen-units` stays at 1 — raising it trades binary size for
       # less time than this saves.
       - name: Build oxen-server Docker Image (${{ inputs.ARCHITECTURE }})
+        env:
+          # Per architecture: a shared ref would have each build overwrite the other.
+          CACHE_REF: ${{ env.CACHE_IMAGE }}:${{ inputs.ARCHITECTURE }}
+          # Branches read the cache, only main writes it, so a branch cannot replace
+          # what main and the release builds read. `github.ref` alone would not do it:
+          # a manual run takes that from the workflow-selection dropdown, not from the
+          # `branch` input it actually builds.
+          WRITE_CACHE: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
         run: |
-          docker build \
+          cache_to=()
+          if [[ "$WRITE_CACHE" == "true" ]]; then
+            # mode=max: the dependency build is not in the final stage, and the
+            # default records only the final stage.
+            cache_to=(--cache-to "type=registry,ref=$CACHE_REF,mode=max")
+          fi
+
+          # --load: a docker-container builder does not populate the daemon, which the
+          # save, tag, and push steps below read from.
+          docker buildx build \
+            --load \
+            --cache-from "type=registry,ref=$CACHE_REF" \
+            "${cache_to[@]}" \
             --build-arg CARGO_PROFILE_RELEASE_LTO=thin \
             --label "org.opencontainers.image.revision=${{ steps.metadata.outputs.COMMIT }}" \
             --label "org.opencontainers.image.version=${{ steps.metadata.outputs.VERSION }}" \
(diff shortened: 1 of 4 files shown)


.github/workflows/build_branch.yml
```diff
@@ -113,14 +117,47 @@ jobs:
...
       # can avoid it. `release_docker.yml` passes no override, so released binaries
       # keep fat LTO. `codegen-units` stays at 1 — raising it trades binary size for
       # less time than this saves.
       - name: Build oxen-server Docker Image (${{ inputs.ARCHITECTURE }})
+        env:
+          # Per architecture: a shared ref would have each build overwrite the other.
+          CACHE_REF: ${{ env.CACHE_IMAGE }}:${{ inputs.ARCHITECTURE }}
+          # Branches read the cache, only main writes it, so a branch cannot replace
+          # what main and the release builds read.
+          WRITE_CACHE: ${{ github.ref == 'refs/heads/main' }}
```
CleanCut: The workflow dispatch dialog sets `github.ref` from your selection from the bulit-in `Use workflow from` dropdown:

Image

Since we're then asking for a _separate_ branch input and using it, we tend to leave the built-in one on `main`. Which means, `WRITE_CACHE` will tend to end up true most of the time even though people are manually entering other branches/tags/whatever in the user input.

So if we really only want to write the cache when we merge into main, we've got to check the event. When you hit the "merge" button, under the hood it creates a merge commit and does a push operation.


```suggestion
          WRITE_CACHE: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```

NOTE: Another option would be to just use the built-in dropdown to select a branch/tag and use that everywhere. You can't specify a commit that way, but you can always create a branch or tag on any commit and still select it.

Dockerfile
```diff
@@ -37,37 +29,62 @@ ENV PKG_CONFIG_PATH="/opt/ffmpeg/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
...
+# rustc, so cargo reuses none of that stage's work. Nothing fails, the build just
+# recompiles every dependency, so anything that decides how a crate compiles belongs
+# here ahead of every compile step. `rustup show` installs the pinned toolchain.
+COPY rust-toolchain.toml ./
+COPY .cargo/config.toml .cargo/config.toml
+RUN rustup show
+
+# cargo-chef keys the dependency build on the manifests instead of the source tree.
+# Version pinned in tool-versions.env, as with the FFmpeg install above.
+COPY tool-versions.env /tmp/chef-install/
```
CleanCut: This is redundant. We already copied in `tool-versions.env` on line 35.

Dockerfile
```diff
@@ -37,37 +29,62 @@ ENV PKG_CONFIG_PATH="/opt/ffmpeg/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
...
 WORKDIR /usr/src/oxen-server
+
+# rustup reads rust-toolchain.toml from the working directory, and the base image's
+# toolchain is not the pinned one. A stage that compiles without it uses a different
+# rustc, so cargo reuses none of that stage's work. Nothing fails, the build just
+# recompiles every dependency, so anything that decides how a crate compiles belongs
+# here ahead of every compile step. `rustup show` installs the pinned toolchain.
+COPY rust-toolchain.toml ./
+COPY .cargo/config.toml .cargo/config.toml
+RUN rustup show
```
CleanCut: Relying on `rustup show` to install the version of Rust we want as a side-effect is prone to breakage. We should explicitly install it.

```suggestion
# here ahead of every compile step.
COPY rust-toolchain.toml ./
COPY .cargo/config.toml .cargo/config.toml
RUN rustup toolchain install
```

tool-versions.env
CleanCut: We can remove this list now. We don't need to maintain a list of things that read this file any more.

```suggestion
```

CleanCut (approved):
I think this is fine, providing we fix the slowdown issue below.

⚠️ This makes release builds slower, since they use the same `Dockerfile` that builds and runs `cargo chef`, but they don't have the same caching setup, so they always do the complete build. If we're going to do this, we ought to do it for release builds as well.

.github/workflows/build_branch.yml
jcelliott: Good call. Updated.

Dockerfile
jcelliott: Updated. Also added `--no-self-update` so we don't waste time updating rustup itself.

jcelliott:
Rebased onto `main`. That picked up the debuginfo change (#864), which added a `CARGO_PROFILE_RELEASE_STRIP` override to the builder stage, so I moved its `ARG`/`ENV` pair above `cargo chef cook` alongside the LTO pair.

`strip` is a fingerprint input for dependencies the same way `lto` is: flipping `CARGO_PROFILE_RELEASE_STRIP` between `debuginfo` and `true` recompiles them, and flipping back is a no-op, so both artifact sets coexist in `target/`. Left below cook, the override would apply only to the final build and discard everything cook produced, taking the whole saving with it. `RUN strip target/release/oxen` stays after the build.

Both build paths leave the strip default in place, so unlike LTO it does not differ between this image build and a release build, and adds nothing to the cache key.

Dockerfile
jcelliott: Not redundant as it was: the earlier COPY landed in `/tmp/ffmpeg-install/`, and that same RUN ended with `rm -rf /tmp/ffmpeg-install`, so the file was gone before cargo-chef needed it. Fair signal that the shape was confusing, though, so it's now a single copy into /tmp/ that nothing deletes, with only install-ffmpeg cleaned up afterward.

PR #865, opened 2026-08-14
Document the visibility rule for new items in the agent instructions

CleanCut:
New code had no stated default for visibility, so items landed `pub` whenever that was the first thing that compiled.

The instructions now say to start at the narrowest visibility that compiles and widen only for a real caller, and record what `pub` means in liboxen specifically: intended for oxen-server, oxen-cli, or the Python bindings.


Eric-Laurence (approved):
Good addition!

PR #866, opened 2026-08-14
Replace instances of "oracle" with "reference implementation"

CleanCut:
In some comments, our AI agents had started using the term "oracle" which was confusing. This replaces the term with "reference implementation", which is what the agents meant.

ENG-1236


Eric-Laurence (approved):
😂

PR #867, opened 2026-08-14
Fail loudly on a truncated merkle archive instead of installing an incomplete node set

CleanCut:
`extract_tar_under` logged and skipped the per-entry error from the tar iterator. A stream cut at an entry boundary therefore ended the unpack loop quietly and the function returned `Ok` with every node past the cut missing, so callers treated a partial install as a success. An ordinary mid-pull network truncation was enough to trigger it — no crafted archive needed — and it was reachable from both the client download/install path and the server upload consumer.

Only the entry-boundary error was swallowed; a cut landing inside a blob already propagated through `read_to_end`. That asymmetry is what let the failure stay silent in exactly the case the post-loop completeness check cannot catch: when the cut falls after a complete node pair, the pending slot is empty, so `IncompleteNode` never fires.

The new test cuts an archive at an entry boundary — trailing zero blocks removed, a partial header appended — and asserts the unpack reports an error. It fails against the previous behavior.

ENG-1232


Eric-Laurence (approved):
Looks good

PR #869, opened 2026-08-17
Fix NFS CI by closing LMDB envs before cleaning up repository directories

CleanCut:
Fix the nightly NFS Test Suite by closing each repo's LMDB env before the directory holding it is removed.

`repositories::delete` also consumes its repository now, so the LMDB env closes before the directory gets deleted.


Eric-Laurence (approved):
Looks good!

PR #872, opened 2026-08-19
Streamline Tests (14): Speed up rm, add, and clean tests

CleanCut:
Use the lightest fixtures needed for `rm`, `add`, and `clean` command tests. Instead of resizing images to change file content, just change file content. Delete a couple redundant tests.

Before: 29.88s (58 tests)
After: 20.50s (56 tests) -- 9.38s (31%) faster


Eric-Laurence (approved):
🏎️💨

PR #873, opened 2026-08-19
Dedup (0.1): Fix df --revision and drop dead chunker

Eric-Laurence:
`oxen df <path> --revision <rev>` should print the committed version of a data frame. Instead it panics on every file, in every format:

```
thread 'main' panicked at crates/liboxen/src/io/chunk_reader.rs:97:62:
called `Result::unwrap()` on an `Err` value: Basic(Chunk not found)
```

The revision path read through `ChunkReader`, which reassembles a file from fixed-size shards written by `FileChunker`. Nothing has ever called that writer (it came from a short lived 2024 attempt at block level dedup).

The shard database is always empty. `commit_writer` fills `chunk_hashes` with a single entry holding the hash for the whole file. The reader looks that hash up, gets nothing, and unwraps the error. Every attempt also left an empty RocksDB behind at `.oxen/tree/chunks`.

`show_node` now reads through `read_version_df`, the same version-store path that diffs, checkout, and workspaces already use, so it works against local and S3 stores and picks the reader from the committed file's extension. Extension handling fixes a second bug: the old code sent anything other than parquet or arrow to the JSON lines reader, so a CSV would have been mis-parsed even if shards had existed.

With the reader gone the whole fixed-size chunking layer became unreachable, so this deletes it: `chunk_reader.rs`, `file_chunker.rs`, the now-empty `io` module, `util::fs::chunk_path`, and the `CHUNKS_DIR` constant. Around 660 lines. Neither file had any tests to remove.

There are still some gaps that this PR doesn't fix: `--schema` is silently dropped when `--revision` is passed, `--item` and paging still behave differently from the working-tree path, and `--write` is ignored on the revision path. I ticketed those as ENG-1672 and ENG-1673.

Completes ENG-1671.

Completes ENG-1671


CleanCut (approved):
Excellent clean up!

PR #874, opened 2026-08-19
Streamline Tests (15): Report per-test durations with `--timings`

CleanCut:
Add `bin/test-rust --timings`, which prints how long each test took next to its result, so it is possible again to see which tests are slow and whether a change made any individual test faster.

When we switched from `cargo nextest` to `cargo test` we lost the individual test times and this gets them back for us. Super helpful when you're looking for slow tests to speed up.

ENG-1658


Eric-Laurence (approved):
👍

PR #880, opened 2026-08-19
Remove the deprecated --workspace flag from workspace status

CleanCut:
Remove the deprecated `--workspace` flag from workspace status. Use `--workspace-id` or `--workspace-name` instead.


Eric-Laurence (approved):
Looks good

PR #881, opened 2026-08-20
Remove the deprecated JSON workspace-staging endpoint

CleanCut:
Clients stage workspace files through the multipart endpoint, which lets the server compute file metadata itself. The JSON endpoint, where the client pre-hashed content and the server staged metadata without reading it, has returned 426 to Oxen clients since 0.51.0 and is now gone.

The version gate that produced the 426 goes with it, along with the error variant it raised, which had no other caller.

Requests that the gate deliberately let through, those with no User-Agent or a non-Oxen one, now receive a 404 instead of a working endpoint; the deprecation registry records that for breakage triage.


Eric-Laurence (approved):
Looks good! 🧹

PR #882, opened 2026-08-20
Stop calling the no-op synced-marker endpoint on push

CleanCut:
Push made one extra round trip per commit to an endpoint that accepts commit hashes and echoes them back without writing anything. Dropping the call removes that round trip and the client wrapper behind it.

The server endpoint stays until 0.59.0. Clients already in the field still send the POST and propagate the failure, so a 404 would fail their pushes over a call that does nothing.

ENG-994


Eric-Laurence (approved):
Why did we have this?

CleanCut:
The client would call it on the server to get the server to write a marker file that indicated a commit was present on the server. The problem is that it would be called before a push succeeded, so it could be wrong, and there was also no reason for it in the first place since we can just look and see if the commit is there or not.

PR #883, opened 2026-08-20
Return an error instead of panicking when a commit's tree root is absent

CleanCut:
Listing or counting a commit's entries asserted that the commit's tree root was present, so a repository missing that node took down the request handler rather than reporting the problem. Both now return an error naming the commit.


Eric-Laurence (approved):
👍

PR #887, opened 2026-08-21
Dedup (1.1): Freeze FastCDC chunk boundaries

Eric-Laurence:
First piece of block level dedup.

This does two important things
1. Chunker splits a stream into content-defined chunks using FastCDC 2020 with normalized chunking at min 8 KiB, target 64 KiB, max 128 KiB.
2. Freezes that exact boundary calculation behind an append only ID. The ID of this chunker is 1.

`Chunker::chunk` takes a reader and yields `RawChunk { offset, data }` in file order. Concatenating the chunks reproduces the input exactly. Chunking is sync and streams with bounded memory (at most one maximum chunk plus a window).

Same bytes with the same chunker id must always produce the same boundaries or deduplication would fail without any errors. `golden_fastcdc_boundaries` tests exact offset and length pairs to ensure boundaries never shift unintentionally. If boundaries ever move due to an intentional chunker update or a fastcdc upgrade or any other reason, we must set a new chunker ID. Also, ID 0 is permanently reserved, so that zeroed or corrupt metadata cannot point to an actual chunker.

`chunker()` which maps chunker id to implementation has no callers yet. Added here to ensure that unknown chunkers fail loudly.

Tests ensure (a) exact tiling, where offsets are contiguous, (b) every chunk except the last sits between the min and max size, (c) every chunk's bytes match the input at its offset, for these inputs:
* Input of all zeros, which doesn't match any mask and cuts at exactly the maximum size every time
* Empty and sub minimum input
* A mid file insertion, which must leave at least 80 percent of chunks sharing hashes with the original

This also introduces `ChunkedError` on the theory that dedup will have its own category of errors. If we want to handle errors differenty, I'm open to ideas.

Completes ENG-1693

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ dotenvy = "0.15.7"
 duckdb = { package = "duckdb", version = "=1.10504.0", default-features = false, features = ["serde_json"] }
 dunce = "1.0.4"
 env_logger = "0.11.3"
+fastcdc = "3.2.1"
 filetime = "0.2.22"
 flate2 = { version = "1.1.0", default-features = false, features = ["zlib-ng"] }
 fs_extra = "1.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,7 @@ dotenvy = "0.15.7"
 duckdb = { package = "duckdb", version = "=1.10504.0", default-features = false, features = ["serde_json"] }
 dunce = "1.0.4"
 env_logger = "0.11.3"
-fastcdc = "3.2.1"
+fastcdc = "=4.0.1"
 filetime = "0.2.22"
 flate2 = { version = "1.1.0", default-features = false, features = ["zlib-ng"] }
 fs_extra = "1.3.0"
--- a/crates/liboxen/src/storage/chunked/chunker.rs
+++ b/crates/liboxen/src/storage/chunked/chunker.rs
@@ -103,12 +103,12 @@ mod tests {
             assert_eq!(chunk.offset, expected_offset, "chunk {i} offset");
             assert!(!chunk.data.is_empty(), "chunk {i} empty");
             assert!(
-                chunk.data.len() <= MAX_CHUNK_SIZE as usize,
+                chunk.data.len() <= MAX_CHUNK_SIZE,
                 "chunk {i} exceeds max size"
             );
             if i + 1 < chunks.len() {
                 assert!(
-                    chunk.data.len() >= MIN_CHUNK_SIZE as usize,
+                    chunk.data.len() >= MIN_CHUNK_SIZE,
                     "non-final chunk {i} under min size"
                 );
             }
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -67,6 +67,7 @@ dotenvy = "0.15.7"
 duckdb = { package = "duckdb", version = "=1.10504.0", default-features = false, features = ["serde_json"] }
 dunce = "1.0.4"
 env_logger = "0.11.3"
+fastcdc = "=4.0.1"
```
CleanCut: We should start on the latest major version (`5.0.0`).

crates/liboxen/src/storage/chunked/error.rs
```diff
@@ -0,0 +1,16 @@
+//! Errors from the dedup layer.
+//! crate::error absorbs these into OxenError through a #[from] declared there.
+
+use thiserror::Error;
+
+/// Failures from chunking.
+#[derive(Debug, Error)]
+pub enum ChunkedError {
```
CleanCut: We should add the variant to `OxenError` directly and delete this entire error module. There's no reason for a separate `ChunkedError` enum that is never matched on before conversion to `OxenError`.

crates/liboxen/src/storage/chunked/registry.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// The fastcdc crate is pinned to an exact version.
+    /// A release that moved a cut by one byte would keep the same API,
+    /// and it would break the format.
+    /// Consumers of liboxen resolve their own dependencies, where no test of ours runs.
+    /// The golden boundary test covers the rest:
+    /// a change to the sizes above, to the chunking code, or to the pin.
+    pub const GENERIC_FASTCDC_V1: ChunkerId = ChunkerId(1);
+
+    // 0 stays permanently unassigned.
+    // Zeroed or corrupt metadata reads as 0, and it has to fail the lookup below.
```
CleanCut: I don't think this concern is valid. We don't expect metadata to be zeroed, and corrupted metadata is no more likely to be zero than any other value.

Let's drop the special handling of zero and start the IDs at zero instead. We can assume that if the metadata is incorrect, then it will be detected at some point upstream via a malfunction or bad data of some sort.

crates/liboxen/src/storage/chunked/chunker.rs
```diff
@@ -0,0 +1,237 @@
...
+pub trait Chunker: Send + Sync {
+    /// The registered ID recorded in manifests produced with this chunker.
+    fn id(&self) -> ChunkerId;
+
+    /// Yield the chunks in file order.
+    /// Concatenating the yielded bytes reconstructs the input exactly.
+    fn chunk<'r>(
+        &self,
+        reader: Box<dyn Read + Send + 'r>,
+    ) -> Box<dyn Iterator<Item = Result<RawChunk, ChunkedError>> + Send + 'r>;
```
CleanCut: Chunks need to be fundamentally retrievable in parallel via some means, so returning a sequential iterator including the actual bytes isn't going to meet our performance needs.

PR #912, opened 2026-09-01
Add `iter_keys` method to produce references to LMDB keys without allocating/copying the values

CleanCut:
`list_hashes` walks the whole node store to collect the keys (hashes) but discards every value, which is relatively expensive because each value is copied into a newly-allocated `Bytes`.

This PR adds `LmdbDb::iter_keys` which borrows each key and avoids allocating anything, and changes `list_hashes` to use it.

ENG-1234

### DIFF
--- a/crates/liboxen/src/lmdb/lmdb_db.rs
+++ b/crates/liboxen/src/lmdb/lmdb_db.rs
@@ -8,10 +8,10 @@
 //! multi-byte integer scanned in numeric order — would have to be encoded big-endian so byte order
 //! matches numeric order, so this layer does not model it (and exposes no scan API).
 //!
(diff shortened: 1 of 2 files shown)


crates/liboxen/src/lmdb/lmdb_db.rs
Eric-Laurence: This READ SAFETY block is no longer strictly true, no?

Eric-Laurence (approved):
Looks good!

crates/liboxen/src/lmdb/lmdb_db.rs
CleanCut: True! ✅ Fixed.

PR #921, opened 2026-09-02
Keep a held DuckDB connection in the cache instead of locking around callers

jcelliott:
Folds the per-data-frame write lock from #896 into the DuckDB connection cache. The hazard is two DuckDB instances over one file, an identity property the cache already owns; it only broke because eviction could drop an entry a caller still held. The cache now never evicts a held entry, `with_db_closed` covers `rename`'s filesystem work, and `core::data_frame_locks` is deleted along with the wrappers around the row writes. Readers never took the lock, so this also closes that gap.

`test_concurrent_row_appends_all_land` now uses LRU pressure with back-to-back writers. Reverting the holder check fails it with a DuckDB checkpoint error. Rust and Python suites green.

#897 and #898 will be rebased onto this and drop their wrappers.


CleanCut (approved):
You can try this in any case. It might be better than what's there currently.

PR #927, opened 2026-09-08
S3 (28.5.2): Record the repository UUID when a remote is attached from a URL

CleanCut:
- Read the remote repository when an Oxen client connects to it and record the UUID it reports.
- Require a reachable server and an existing repository to attach a remote.
- Add tests

ENG-1807


Eric-Laurence (approved):
Looks good

PR #935, opened 2026-09-11
S3 (29.1): Total the version-file bytes a repository references

CleanCut:
Report the bytes a repository's version files take, so the figure comes out the same whether those files sit in the repo directory, on a mount `versions_path` points at, or in an S3 bucket. Each version counts once, however many names reach it.

ENG-1584


Eric-Laurence (approved):
Looks good

PR #937, opened 2026-09-12
Read JSONL data frames from bytes with the trailing newline trimmed

gschoeni:
## Problem

We are getting sentry errors from LLM usage:

https://oxen-ai.sentry.io/issues/7707549043/?alert_rule_id=16453399&alert_type=issue&notification_uuid=0b939a3f-0ef5-40a2-91af-11b163869842&project=4504679959035904&referrer=slack

`GET /repos/{ns}/{repo}/data_frames/{branch}/<path>.jsonl` intermittently returns a 500 of the form:

```
task ... panicked with message "called `Result::unwrap()` on an `Err` value: JoinError::Panic(..., \"range start index 5476 out of range for slice of length 5475\")"
```

It is a polars 0.49 bug in the NDJSON reader (`polars_io::ndjson::core::get_line_stats_json`, still the same on polars `main`). When a scan carries a positive slice, which every paginated read does, the reader samples line lengths from the 75% byte mark. `next_line_position_naive_json` returns the index *after* the newline and the sampler then skips one more byte, so when the first `}\n` at or past that mark is the file's last byte it indexes one past the end. The file needs six or more rows (the first sampling pass consumes up to five and bails at EOF) and a last row that spans the 75% mark. A chat history whose newest conversation is longer than the rest combined is exactly that, and OxenHub's history writer was recreating the file on the error.

Reproduced in a standalone polars 0.49.1 project: path scan of that shape + `.slice(0, 100)` panics; the same bytes as a `ScanSources::Buffers` with the trailing newline trimmed, or a file with no trailing newline, read fine.

## Fix

JSONL reads now hand polars an in-memory scan source with trailing newlines trimmed (`ndjson_scan_source`), which makes the bad position unreachable; NDJSON without a final newline is valid. Local files are memory-mapped and sliced without copying. The S3 path reads the object through the version store instead of a cloud scan, which the NDJSON reader was downloading in full anyway. Adds a direct `polars-utils` dependency (already in the lockfile via polars) for `MemSlice`.

## Tests

- New `test_read_jsonl_slice_when_last_row_spans_three_quarter_mark` builds the panic-shape file and reads it with a page slice and with a row limit.
- `core::df::tabular::tests::test_read_jsonl*` and all 15 `storage::s3::tests::cloud_reads` tests pass; clippy is clean.

https://claude.ai/code/session_01XetgBiYyiqP3yFsbF5ooN6

### DIFF
--- a/crates/liboxen/src/core/df/tabular.rs
+++ b/crates/liboxen/src/core/df/tabular.rs
@@ -1,6 +1,7 @@
 use duckdb::ToSql;
 use polars::io::cloud::CloudOptions;
 use polars::prelude::*;
+use polars_utils::mmap::MemSlice;
 use serde_json::json;
 use std::collections::HashSet;
 use std::fs::File;
@@ -70,13 +71,36 @@ fn read_df_csv(
         .map_err(|_| OxenError::basic_str(format!("{}: {:?}", READ_ERROR, path.as_ref())))
 }
 
-fn read_df_jsonl(path: impl AsRef<Path>) -> Result<LazyFrame, OxenError> {
-    let path = path
-        .as_ref()
-        .to_str()
-        .ok_or_else(|| OxenError::basic_str("Could not convert path to string"))?;
-    LazyJsonLineReader::new(path)
+/// An NDJSON file's bytes as an in-memory polars scan source, with trailing newlines trimmed.
+///
+/// polars' NDJSON reader (`polars_io::ndjson::core::get_line_stats_json`, still the same on main
+/// as of 0.49) samples line lengths from the 75% byte mark whenever the scan carries a positive
+/// slice, which every paginated data frame read does. The sampler takes the position *after* the
+/// first `}\n` in that tail and then skips one more byte, so when that newline is the last byte of
+/// the file it indexes one past the end and panics with "range start index N+1 out of range for
+/// slice of length N". That is any file of six or more rows whose last row spans the 75% mark: a
+/// chat history whose newest conversation is longer than the rest combined, for one. With no
+/// trailing newline that position cannot be produced, and NDJSON without a final newline is valid.
+///
+/// Memory-mapped and in-memory sources are sliced without copying, so this costs nothing over
+/// handing polars the path, and the streaming reader loads the whole file either way.
+fn ndjson_scan_source(bytes: MemSlice) -> ScanSources {
+    let end = bytes
+        .iter()
+        .rposition(|&b| b != b'\n' && b != b'\r')
+        .map_or(0, |i| i + 1);
+    ScanSources::Buffers(Arc::from([bytes.slice(0..end)]))
+}
+
+fn lazy_jsonl_reader(bytes: MemSlice) -> LazyJsonLineReader {
+    LazyJsonLineReader::new_with_sources(ndjson_scan_source(bytes))
         .with_infer_schema_length(Some(NonZeroUsize::new(10000).unwrap()))
+}
+
+fn read_df_jsonl(path: impl AsRef<Path>) -> Result<LazyFrame, OxenError> {
+    let path = path.as_ref();
+    let bytes = MemSlice::from_file(&File::open(path)?)?;
+    lazy_jsonl_reader(bytes)
         .finish()
         .map_err(|_| OxenError::basic_str(format!("{READ_ERROR}: {path:?}")))
 }
@@ -148,12 +172,9 @@ pub fn scan_df_csv(
 }
 
 pub fn scan_df_jsonl(path: impl AsRef<Path>, total_rows: usize) -> Result<LazyFrame, OxenError> {
-    let path = path
-        .as_ref()
-        .to_str()
-        .ok_or_else(|| OxenError::basic_str("Could not convert path to string"))?;
-    LazyJsonLineReader::new(path)
-        .with_infer_schema_length(Some(NonZeroUsize::new(10000).unwrap()))
+    let path = path.as_ref();
+    let bytes = MemSlice::from_file(&File::open(path)?)?;
+    lazy_jsonl_reader(bytes)
         .with_n_rows(Some(total_rows))
         .finish()
         .map_err(|_| OxenError::basic_str(format!("{READ_ERROR}: {path:?}")))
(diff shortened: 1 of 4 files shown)


crates/liboxen/src/core/df/tabular.rs
```diff
@@ -70,13 +71,36 @@ fn read_df_csv(
...
-fn read_df_jsonl(path: impl AsRef<Path>) -> Result<LazyFrame, OxenError> {
-    let path = path
-        .as_ref()
-        .to_str()
-        .ok_or_else(|| OxenError::basic_str("Could not convert path to string"))?;
-    LazyJsonLineReader::new(path)
+/// An NDJSON file's bytes as an in-memory polars scan source, with trailing newlines trimmed.
+///
+/// polars' NDJSON reader (`polars_io::ndjson::core::get_line_stats_json`, still the same on main
+/// as of 0.49) samples line lengths from the 75% byte mark whenever the scan carries a positive
```
Eric-Laurence, line 77: I would like to make the fix upstream actually. It's a small fix, but their first time contributor policy and ai policy will make it require a little more effort. I'm not in a rush to do this, but there should be something that lets us know to check when the issue is fixed so we don't keep the workaround and polars-utils dependency forever.

Concretely, we should probably add a test that asserts that it fails without the trimming. When it fails it can let us know to 1. delete ndjson_scan_source, 2. get rid of polars-utils, 3. keep the regression test.

crates/liboxen/src/core/df/tabular.rs
```diff
@@ -70,13 +71,36 @@ fn read_df_csv(
...
+/// slice of length N". That is any file of six or more rows whose last row spans the 75% mark: a
+/// chat history whose newest conversation is longer than the rest combined, for one. With no
+/// trailing newline that position cannot be produced, and NDJSON without a final newline is valid.
+///
+/// Memory-mapped and in-memory sources are sliced without copying, so this costs nothing over
+/// handing polars the path, and the streaming reader loads the whole file either way.
+fn ndjson_scan_source(bytes: MemSlice) -> ScanSources {
+    let end = bytes
+        .iter()
+        .rposition(|&b| b != b'\n' && b != b'\r')
```
Eric-Laurence, line 90: I don't think we need the \r part here.

crates/liboxen/src/core/df/tabular.rs
```diff
@@ -70,13 +71,36 @@ fn read_df_csv(
...
+}
+
+fn lazy_jsonl_reader(bytes: MemSlice) -> LazyJsonLineReader {
+    LazyJsonLineReader::new_with_sources(ndjson_scan_source(bytes))
         .with_infer_schema_length(Some(NonZeroUsize::new(10000).unwrap()))
+}
+
+fn read_df_jsonl(path: impl AsRef<Path>) -> Result<LazyFrame, OxenError> {
+    let path = path.as_ref();
+    let bytes = MemSlice::from_file(&File::open(path)?)?;
```
Eric-Laurence, line 102: If the file is missing or unreadable this returns an io::Error. We could wrap it so the error includes the path.

crates/liboxen/src/core/df/tabular.rs
```diff
@@ -148,12 +172,9 @@ pub fn scan_df_csv(
...
 
 pub fn scan_df_jsonl(path: impl AsRef<Path>, total_rows: usize) -> Result<LazyFrame, OxenError> {
-    let path = path
-        .as_ref()
-        .to_str()
-        .ok_or_else(|| OxenError::basic_str("Could not convert path to string"))?;
-    LazyJsonLineReader::new(path)
-        .with_infer_schema_length(Some(NonZeroUsize::new(10000).unwrap()))
+    let path = path.as_ref();
+    let bytes = MemSlice::from_file(&File::open(path)?)?;
```
Eric-Laurence, line 176: Same here as the above

crates/liboxen/src/core/df/tabular.rs
```diff
@@ -1922,6 +1945,43 @@ mod tests {
...
+            let path = dir.join("chat_history.jsonl");
+            let mut contents = String::new();
+            for id in 0..6 {
+                contents.push_str(&format!("{{\"id\":{id},\"text\":\"short\"}}\n"));
+            }
+            contents.push_str(&format!(
+                "{{\"id\":6,\"text\":\"{}\"}}\n",
+                "x".repeat(20_000)
+            ));
+            std::fs::write(&path, contents)?;
```
Eric-Laurence, line 1964: Shouldn't this be `util::fs::write_to_path`?

Eric-Laurence (dismissed):
I was able to verify the issue, everything looks good save for a few minor things.

CleanCut (approved):
@Eric-Laurence Thanks for providing the space for me to do a little research! I'm restoring the approval you gave.

The polars bug is a single line fix upstream, but we're using an old version of polars (0.49) and a less-old version (0.53) already has the fix, so contributing the fix to 0.49 and consuming it is not going to happen like I hoped.

Upgrading to polars 0.53 (or the latest) is a significant, breaking change that doesn't need to be done right this second. Let's merge this PR in, but let's also revert it when we upgrade to a newer polars in the (hopefully) near future.

crates/liboxen/src/core/df/tabular.rs
gschoeni: Added in 21169ba71: `test_polars_path_scan_still_fails_when_last_row_spans_three_quarter_mark` does a plain path scan of the same fixture with a slice and asserts it does not succeed (on 0.49 it panics at `ndjson/core.rs:498` with the same nested `JoinError::Panic` as the Sentry report). Its doc comment and assertion message spell out the cleanup: delete `ndjson_scan_source`, drop `polars-utils`, keep the regression test. The fixture is now shared between the two tests.

crates/liboxen/src/core/df/tabular.rs
gschoeni: Agreed, dropped in 42b370a9e. `next_line_position_naive_json` only takes a boundary at a `}` immediately before `\n`, so a file ending in `}\r\n` never reaches the bad index and stripping `\r` bought nothing.

crates/liboxen/src/core/df/tabular.rs
gschoeni: Fixed in 29e61f9a1: both readers now go through `mmap_jsonl`, which opens via `util::fs::open_file` (already names the path) and wraps the mmap error with the path too.

crates/liboxen/src/core/df/tabular.rs
gschoeni: Same fix, 29e61f9a1.

crates/liboxen/src/core/df/tabular.rs
gschoeni: Yes, fixed in 512c1657a.

PR #939, opened 2026-09-14
Accept absolute and parent-relative clone destinations

jcelliott:
`oxen clone <url> <dst>` rejected destinations that were absolute or contained `..`, so a scripted caller had to `cd` to the parent and clone with a relative name. Git accepts both forms.

The guard was framed as a path-traversal defense, but the destination is an argument the user types in their own shell. `repositories::clone` already accepts any `dst`, so the library and the Python bindings could always pass an absolute path; only the CLI could not. `clone_repo` still refuses a destination that exists, and `create_dir_all` makes missing parents. The `starts_with(&current_dir)` check below the guard goes with it: it was already dead code.

Adjacent fix: the default directory name now comes from the last non-empty URL segment, so a trailing slash no longer resolves the destination to the current directory.

`resolve_destination` and `repo_name_from_url` are split out for unit tests, which this file had none of.


Eric-Laurence (approved):
Looks good!

PR #940, opened 2026-09-14
S3 (29.2): Size a repository by the version files its tree references

CleanCut:
Calculate the size of a repository from the file sizes stored in the merkle tree, so the size is accurate when the version files are stored in S3 or outside the `.oxen` directory at some other disk location. This also fixes counting repository metadata as stored version file sizes.

- start a size recalculation when a push or a workspace commit finishes
- Add a specific `POST /action/completed/push` handler to trigger the size recalculation (the endpoint was already called under a wildcard `POST /action/completed` handler).
- Skip a recalculating size while repository maintenance like a a migration, a prune, or a workspace clear holds the repository 
- Change the file extension of the cached file size JSON object file from `.toml` to `.json` and delete the legacy toml file if it exists
-  Fix the documented return value of `GET /size` for the OpenAPI docs

ENG-1821


crates/liboxen/src/repositories/size.rs
```diff
@@ -64,19 +80,21 @@ pub fn update_size(repo: &LocalRepository) -> Result<(), OxenError> {
...
+
+        let size_result = repo.version_bytes();
         match size_result {
             Ok(calculated_size) => {
                 let size = RepoSizeFile {
                     status: SizeStatus::Done,
                     size: calculated_size,
                 };
-                if let Err(e) = AtomicFile::new(&path_clone).write(size.to_string().as_bytes()) {
+                if let Err(e) = AtomicFile::new(&path).write(size.to_string().as_bytes()) {
```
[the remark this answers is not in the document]
CleanCut: That's deliberate. We're addressing it in step 29.4

crates/liboxen/src/repositories/size.rs
```diff
@@ -96,22 +114,56 @@ pub fn update_size(repo: &LocalRepository) -> Result<(), OxenError> {
...
-            Ok(size)
-        }
-        Err(e) => {
-            log::info!("Size file not found, creating it: {e}");
-            update_size(repo)?;
-            get_size(repo)
-        }
+    if let Ok(recorded) = util::fs::read_from_path(&path) {
+        return Ok(serde_json::from_str(&recorded)?);
+    }
```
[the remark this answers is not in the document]
CleanCut: This isn't a concern, because of the refactors in the next steps.

crates/liboxen/src/repositories/size.rs
```diff
@@ -96,22 +114,56 @@ pub fn update_size(repo: &LocalRepository) -> Result<(), OxenError> {
...
+
+/// Poll the figure recorded for `repo` until a recalculation has landed, panicking past 30s.
+#[cfg(test)]
+pub(crate) fn wait_for_recorded_size(repo: &LocalRepository) -> Result<u64, OxenError> {
+    use std::time::{Duration, Instant};
+
+    let deadline = Instant::now() + Duration::from_secs(30);
+    loop {
+        let recorded = get_size(repo)?;
+        if recorded.status == SizeStatus::Done {
```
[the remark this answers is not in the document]
CleanCut: That is an excellent find! ✅ Fixed.

crates/liboxen/src/repositories/size.rs
```diff
@@ -38,7 +40,20 @@ impl fmt::Display for RepoSizeFile {
...
+/// Each call starts its own pass, and the last one to finish is the figure that sticks. No pass
+/// starts while a maintenance operation holds `repo`, and the figure already recorded stays as it
+/// is.
 pub fn update_size(repo: &LocalRepository) -> Result<(), OxenError> {
+    // An exclusive maintenance operation drains this reservation before it runs, so the walk
+    // never reads a store that is being deleted or migrated.
+    let Ok(write) = repo_locks::acquire_write(repo) else {
+        log::info!("Skipping a size recalculation while the repository is held for maintenance");
+        return Ok(());
+    };
```
Eric-Laurence, line 55: Should we be holding the write lock while the background thread for size calculation is running? Seems like we could run into timeouts.

crates/liboxen/src/repositories/size.rs
```diff
@@ -86,7 +104,7 @@ pub fn update_size(repo: &LocalRepository) -> Result<(), OxenError> {
                     status: SizeStatus::Error,
                     size: 0,
```
Eric-Laurence, line 105: This is pre existing, but won't this make failed recalculations wipe the cached size value? Also, since we propagate the error in get_storage_for_repo, this means one repo failing the size calculation will cause the whole namespace's calculation to fail. Probably a good idea to fix this while we're here

Eric-Laurence (approved):
Looks good

crates/liboxen/src/repositories/size.rs
CleanCut: Yes, but this is worth explaining. This is a trickier-than-normal lock. `repo_locks::acquire_write` doesn't block itself or other calls of `repo_locks::acquire_write`. It increments a counter and returns immediately, and then decrements the counter when it's dropped. The big maintenance operations like `migrate` or `prune` call `with_repo_exclusive` which _does_ block until that counter goes to zero. And `acquire_write` blocks while a maintenance operation is running. In both cases, we'll timeout with a 429 (retry-after). So, if lots of activity is going on, the maintenance request will timeout and try again, and if maintenance is going on then regular operations will time out and try again.

crates/liboxen/src/repositories/size.rs
CleanCut: Good catch! Fixed in #954

crates/liboxen/src/repositories/size.rs
CleanCut: ✅ Renamed over in #957

PR #943, opened 2026-09-14
Speed up every RocksDB open by not spawning a thread per core to do it

CleanCut:
Open these stores with a single file-opening thread, since none of them hold more than a single database file.

This benefits OxenHub's upstream test suite as well, taking it from 47.5s to 42.8s, and its slowest repository commit from 214ms to 156ms.

ENG-1243


Eric-Laurence (approved):
Makes sense!

PR #946, opened 2026-09-15
Open the workspace staged database once per file move

jcelliott:
`workspaces::files::mv` opened the workspace's staged RocksDB twice per rename and resolved the source path in the base commit's merkle tree twice. It now does each once.

The first `get_staged_db_manager` was a temporary: its `StagedDBManager` dropped at the end of the statement that read the source entry, closing RocksDB and releasing the per-directory `LOCK`, and a second call reopened it to write the destination. The two merkle lookups were the fallback node for a source that is not already staged, and the check for whether the source has to be staged for removal. Both resolve the same path against the same commit.

## Measurement

2000 sequential renames against a 2000-file workspace:

| | total | per 100 renames |
| --- | --- | --- |
| before | 61.8s | 2.9s rising to 3.3s |
| after | 38.3s | flat at ~1.9s |

The per-rename cost no longer climbs as renames accumulate in the staged database.

## Test

`test_mv_twice_leaves_one_removed_marker` pins the branch the shared source lookup feeds. In a chained rename, the second move's source is staged but absent from the base commit, so it is unstaged outright rather than staged for removal, and only the originally committed path keeps the `Removed` marker.


crates/liboxen/src/core/v_latest/workspaces/files.rs
```diff
@@ -1096,16 +1096,25 @@ pub fn mv(
 
     let workspace_repo = &workspace.workspace_repo;
 
+    // One handle for the whole rename: the staged DB opens and closes with the manager, so a
+    // second `get_staged_db_manager` here would reopen the same RocksDB mid-rename.
```
Eric-Laurence: This comment is a little confusing, with the first db open the second db would use a shared handle. The issue was the first db getting dropped, then the second one would have to do an open from scratch, no?

Eric-Laurence (approved):
Looks good

crates/liboxen/src/core/v_latest/workspaces/files.rs
jcelliott: Yep, good call. I just dropped the comment entirely, it wasn't really adding any value.

PR #948, opened 2026-09-15
Add `oxen reset` to move the current branch to another commit

jcelliott:
Rewriting a commit meant reaching for `oxen branch -f <current-branch> <commit>`, which moves the ref without touching the working tree. That is what `git reset` spells directly, but nothing in Oxen named it, so the recovery path for a bad commit was undiscoverable.

`oxen reset <revision>` takes a commit id or a branch name and moves the checked out branch to it.

- **Default (`--mixed`)** leaves the working tree alone, so files from the dropped commits stay on disk and can be staged and committed again.
- **`--hard`** also restores the working tree to the target commit.
- **`--soft`** is rejected with a message pointing at the default rather than silently doing something else.

`--soft` needs something that does not exist yet: `add_file_node_to_staged_db` can write staged entries without rehashing, but nothing exposes the set of file nodes that differ between two commits outside `list_diff_entries`, which is paginated and shaped for display. Worth doing separately, since `--soft` would let a large repo skip rehashing its whole working tree to recreate a commit.

Two follow-ups this does not cover: a docs page for the rewrite workflow in the `docs` repo, and a Python binding (`PyRepo` exposes a curated subset of operations and does not have one for this).


CleanCut:
Closing for now. We'll revisit when we work on ENG-1938

PR #956, opened 2026-09-16
Crate Deps (2): Update rocksdb to 0.25

CleanCut:
Update from RocksDB 8.10 to 11.8

- Hold the block-based table format at version 6, which RocksDB 11.8 would otherwise raise to 7. A version 7 table reads as corruption to 8.10, so taking the new default would make a local repository written by a newer `oxen` or `oxenai` unreadable to a released one still in the field. Version 6 is the highest that every release back to v0.48.0 can read, and it adds footer checksum protection that 5 lacks. This does not affect oxen-server, which only rolls forward.
- Enable `bindgen-runtime`, a rocksdb default feature that this dependency's `default-features = false` turns off. Without it clang-sys links libclang at build-script link time rather than loading it dynamically, and the build script fails to load.
  - A plain `cargo check --workspace` passes either way, because `oxen-py` build-depends on bindgen with the `runtime` feature and Cargo unifies build-dependency features across the graph. The selection that exercises this is `--workspace --exclude oxen-py`, which is what the Docker image and all three release workflows build.
- Add a test that reads the format version out of an SST footer, so a future engine bump that moves the default fails here rather than in the field.
- No API changes.

ENG-1773


Eric-Laurence (approved):
Tested locally, everything works as expected 👍

PR #957, opened 2026-09-16
Stop the per-repo write guard from promising mutual exclusion it does not provide

CleanCut:
_A rename with no behavior change to help us more intuitively understand what's going on. Prompted by the misunderstanding in [this review comment](https://github.com/Oxen-AI/Oxen/pull/940#discussion_r4023257811)._

Rename `acquire_write` to `begin_write` and `RepoWriteGuard` to `WriteInFlight`, so nobody mistakes the server's write path for a serialized one. Any number of writes hold the guard at once, and it never waits for another writer, only failing fast while a maintenance operation holds the repo.


Eric-Laurence (approved):
Looks good! I had claude scan for any relevant places that weren't updated, looks like you might have missed a comment here and there:

Image

CleanCut:
Nice catch! ✅ Fixed them all.

PR #958, opened 2026-09-17
Crate Deps (3): Bump the eight dependencies that need no porting

CleanCut:
Bring every direct dependency that can move without a code change or a build tooling change up to its latest stable, leaving behind only the ones that need real work.

- Move `s3s` and `s3s-fs` 0.14 to 0.16, picking up the crates s3s split its signing and encoding into.
- Move `duckdb` and `libduckdb-sys` 1.10504.0 to 1.10505.0 together, since both carry exact pins.
- Move `async_zip` to 0.0.19, `dirs` to 7.0, `sqlparser` to 0.63.0, and oxen-py's `bindgen` build dependency to 0.73.2.
- No code change.

ENG-1877


Eric-Laurence (approved):
Looks good, that failing windows test is worth digging into of course, doesn't seem like it's obviously a flake on inspection.

PR #964, opened 2026-09-21
Tag Sentry events with the request id to link them to logs and OxenHub

CleanCut:
Every `oxen-server` error reported to Sentry now carries a `request_id` tag, matching an issue there to the server's own access-log lines, which key on `req=`, and to the OxenHub event for the same request, which already tags the id under `request_id`. The tag goes on the per-request hub `sentry-actix` installs, so it also reaches events reported from the blocking-pool tasks that inherit that hub. Self-hosted deployments configure no DSN and are unaffected.


Eric-Laurence (approved):
Looks good

PR #967, opened 2026-09-21
Streamline Tests (19): Build each pull test's scenario once

CleanCut:
_This is a test-only change._

Cut repositories::pull from 26.7s to 25.5s, and 28 tests to 27.

- Merge the two separate-branch tests, identical for the 43 body lines before their final clone and differing only in which branch they clone, into one that builds the scenario once and clones both ways.
- Take the repository the fixture already committed, attached, and pushed as one of the two actors in eight tests that were discarding it and cloning a second one.
- Add a testing rule to AGENTS.md: a helper handing back a local and remote repository pair has already attached them, so a test needing two actors clones only the other one.

ENG-1707


Eric-Laurence (approved):
👍

